Issue #38, opened 2014-10-01
Add Websockets support


saidimu:
The current nginx template does not allow upgrading websocket connections.

I added a few lines and an environment variable to allow this. Details in pull request #37.

PR #46, opened 2014-10-26
Send "Connection: upgrade" when "Upgrade" header is received

md5:
Closes #37 #38

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -5,6 +5,13 @@ map $http_x_forwarded_proto $proxy_x_forwarded_proto {
   ''      $scheme;
 }
 
+# If we receive Upgrade, set Connection to "upgrade"; otherwise, delete any
+# Connection header that may have been passed to this server
+map $http_upgrade $proxy_connection {
+  default upgrade;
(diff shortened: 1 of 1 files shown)


mko-x:
@md5 or @jwilder 
Could you please tell me, why the 'Connection' header value has to be deleted if 'Upgrade' is unset? Usual suspects would set it to 'close' like to see in [nginx docs sample](http://nginx.org/en/docs/http/websocket.html), don't they?

Is it related to [w3 rfc2616 14.10](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html)?

Thanks in advance!

md5:
@mko-x I think you're right that it should send `close` as in the Nginx docs example. The comments on #49 seem too indicate that I had included that change in that PR originally, but ended up dropping the change. It looks like the change never made it into a new PR after that.

If you want to open a PR to make the change, I think it would be a good idea.

mko-x:
@md5 
I was wondering as it made sense to me as stated in #49. 
Ok, I'll provide a PR.

Thank you

PR #72, opened 2014-12-24
Ability to set a default host for nginx

### LINKED ISSUE #25: Allow setting catch all virtual hosts
It would be really useful if I didn't have to set a virtual host. On EC2 I have to set the virtual host to a whole load of values because:

1. On ELB health checks the Host header is the instance private IP address
2. On normal requests the Host header is the domain

http://serverfault.com/questions/527156/setting-nginx-to-catch-all-unhandled-vhosts

BenHall:
When launching nginx-proxy you can use an env var to set which should be the default host header. If an env var is not defined then the default 503 will be returned. 

Closes #25

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN apt-get update \
 RUN echo "daemon off;" >> /etc/nginx/nginx.conf \
  && sed -i 's/^http {/&\n    server_names_hash_bucket_size 64;/g' /etc/nginx/nginx.conf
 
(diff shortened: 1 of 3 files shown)


wrow:
:+1:

jmillerdesign:
Yes, please! :thumbsup:

ReedD:
I need this :+1:

jwilder:
Thanks @BenHall!

PR #78, opened 2015-01-03
Display a list of discovered virtual hosts

mathroc:
When trying to access the reverse proxy without a valid virtual host, instead of throwing a 503, shows a list of available virtual hosts links. if there is no virtual hosts yet, briefly explains how to start one

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,3 +95,12 @@ Note that in the latter case, a browser may get an connection error as no certif
 to establish a connection.  A self-signed or generic cert named `default.crt` and `default.key`
 will allow a client browser to make a SSL connection (likely w/ a warning) and subsequently receive
 a 503.
+
+### Basic Authentication Support
+
+In order to be able to securize your virtual host, you have to create a file named as its equivalent VIRTUAL_HOST variable on directory
+/etc/nginx/htpasswd/$VIRTUAL_HOST
+
+    $ docker run -d -p 80:80 -p 443:443 -v /path/to/htpasswd:/etc/nginx/htpasswd -v /path/to/certs:/etc/nginx/certs -v /var/run/docker.sock:/tmp/docker.sock jwilder/nginx-proxy
+
+You'll need apache2-utils on the machine you plan to create de htpasswd file. Follow these [instructions](http://httpd.apache.org/docs/2.2/programs/htpasswd.html)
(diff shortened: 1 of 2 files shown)


thaJeztah:
I think this makes up a nice example on how to possibly extend your setup, but I'm not comfortable having this in the default image; especially not when using nginx-proxy in a public/production environment and you possibly don't want to expose all virtual hosts (think of `staging.some-domain.com`, `sneak-preview.example.com`).

For that reason, I'm -1 on this.

md5:
I also don't think this is good default behavior, but I could see it being useful.

It would be interesting to make an image using [docker-gen](https://github.com/jwilder/docker-gen) that could maintain a current list of servers in its sibling containers based on their `VIRTUAL_HOST`. If that were combined with #72, you could make that server the default host for `nginx-proxy` to show a list of known host names.

md5:
In fact, now that I look closer at the diff, this PR could very easily be repurposed to create such an image by using the same template and `docker-gen` command.

thaJeztah:
Yes, I do see some interesting bits in this proposal. My comment was primarily to prevent this becoming the default template :)

mathroc:
@md5 yes making another image could be a good solution, I'll look into that

mathroc:
and here it is : https://registry.hub.docker.com/u/texthtml/docker-vhosts/ !

just try `docker run --restart always -it -e VIRTUAL_HOST=dev.127.0.0.1.xip.io -v /var/run/docker.sock:/tmp/docker.sock --name virtual_hosts texthtml/docker-vhosts` if your nginx-proxy is on 127.0.0.1:80 it should be working on http://dev.127.0.0.1.xip.io/ :)

thaJeztah:
Nice!  I think you can even simplify the Dockerfile to:

    FROM jwilder/nginx-proxy:latest

    COPY default.conf /etc/nginx/conf.d/default.conf
    COPY app/ /app/

That way you don't have to maintain all the things needed to install/setup docker-gen

mathroc:
you're absolutely right, and it's done

thaJeztah:
Nice!

jdoose:
@mathroc Nice work! Exactly what I was looking for. I found it after I developed nearly the same ;)
What do you think about putting a link to the docker-vhosts project into the nginx-proxy readme file? Maybe in a separate section?

thaJeztah:
Perhaps @jwilder could consider adding a "contrib" directory to the repository, where people can add custom templates that may be useful

PR #80, opened 2015-01-04
avoid installing wget

### LINKED ISSUE #79: use 'ADD ...' instead of 'RUN wget ...' to avoid installing wget
To simplify the Dockerfile and make the a image a bit lighter. ADD could be used instead of wget to download forego & docker-gen

I'll make a PR soon if you think it would be an improvement

mathroc:
download forego & docker-gen locally before build to avoid having to
download it on build
copy Procfile, nginx.tmpl & forego in a single command

fix #79 

it makes the image lighter but the repository become bigger, I'll let you take a look at this option and merge it if you think it's worth it


jwilder:
I'd rather not add the binaries to this repo.

PR #81, opened 2015-01-05
Gzip application/javascript

asmundg:
As per RFC4329, nginx uses application/javascript as the default MIME
type for .js files. Nginx-proxy will now gzip these files if the client
requests it.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -12,7 +12,7 @@ map $http_upgrade $proxy_connection {
   ''      '';
 }
 
-gzip_types text/plain text/css application/json application/x-javascript text/xml application/xml application/xml+rss text/javascript;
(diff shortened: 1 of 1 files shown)


jwilder:
Thanks!

md5:
Has anyone confirmed that this `gzip_types` setting is even having any effect? I know that I touched that line when I moved it to the top-level of the config, but seeing this PR merged made me wonder whether it's also necessary to set [`gzip_proxied`](http://nginx.org/en/docs/http/ngx_http_gzip_module.html#gzip_proxied) for this to have any effect.

md5:
I just realized that `gzip_proxied` controls whether proxied _requests_ get gzipped, not proxied _responses_. Sorry!

Issue #83, opened 2015-01-08
license missing


wenlock:
@jwilder Hi!  just ran across this, and it looks like a very elegant and simple solution to proxy ports.   I'm liking it.   Do you have a license picked yet.  Would be great to see apache 2 or mit license on this work.

jwilder:
I usually make stuff MIT. docker-gen which this uses is MIT.

wenlock:
Hoping you don't mind letting me help you make that quick change.

jwilder:
Nope. Thanks for the help!

PR #84, opened 2015-01-08
Adding license for usage

wenlock:
Help close: https://github.com/jwilder/nginx-proxy/issues/83

Adding a license so we can use it with other projects.


jwilder:
Thanks!

Issue #74, opened 2014-12-28
[SSL]VIRTUAL_HOST + (+.crt/.key) does not work for me


tobiasbaehr:
* my VIRTUAL_HOST is mysite.dev.
-> So my files are mysite.dev.crt and mysite.dev.key.

But the nginx response in the log is: 

no "ssl_certificate" is defined in server listening on SSL port while SSL handshaking

What means he found something. \o/ ;-)

When I rename the files to default(.crt/.key) then it works.

What is here wrong?

engine07:
Hi 

I had same issue and altered nginx.tpl file by replacing:
listen 443  ssl;
with: 
listen 443 default_server ssl;

in this part:
server {
        error_log /app/file.log debug;

        server_name {{ $host }};
        listen 443 default_server ssl;

and it works now.

Did this by making new image, Dockerfile:
FROM jwilder/nginx-proxy
COPY ./nginx.tmpl /app/nginx.tmpl

log0ymxm:
This also worked for me. I've only got one ssl service running and a few non-ssl, will this cause an issue if there are multiple ssl services? Otherwise this is worth fixing.

Nitesedge:
Also worked for me.

Took a couple attempts to get it right so here's a pastebin for clarification: http://pastebin.com/0psiBGyN

I am running multiple ssl containers without issue.

phal0r:
I have some problems with your suggested fix. Is anyone of you running more than one docker image with ssl? 

If i start one everything is working, but if i start a second container that uses ssl i get the following error in nginx:

[emerg] 35#0: a duplicate default server for 0.0.0.0:443 in /etc/nginx/conf.d/default.conf:243

and it does not connect the second container to nginx. Any ideas?

Nitesedge:
How are you testing this fix? I tried to update the template after pulling and running the trusted image and ran into the same issue. When I built from the repo (git clone https://github.com/jwilder/nginx-proxy.git) it worked with no changes (since this fix was already in the repo).

log0ymxm:
@phal0r I figured that might be an issue. I've only got the one ssl container.

phal0r:
@log0ymxm The behaviour is really strange. Now i am using the nginx-proxy without the fix suggested here and the following does work:

[...]

PR #85, opened 2015-01-09
Fix ssl server listen statement

log0ymxm:
This is a possible fix for #74


jwilder:
Can you elaborate on what this is fixing?

log0ymxm:
This implements the change recommended by @engine07. I can't comment on it too much do to my unfamiliarity with nginx configuration, but it seemed to fix the issue I was facing locally. This may be addressed better with #91, but I'm not sure.

PR #90, opened 2015-01-22
added wildcard hosts, fixes #89

### LINKED ISSUE #89: support for wildcard virtual-hosts
We're using [xip.io](http://xip.io) to name and access our vHosts, which works very nice so far.

The only downside is, that we have to add several possible options to our `VIRTUAL_HOST` env variable, if we'd like to access it on different environments or from other clients on the network.

So our config looks like this:

```
VIRTUAL_HOST: kitchen-sink.127.0.0.1.xip.io,kitchen-sink.192.168.59.103.xip.io,kitchen-sink.192.168.77.10.xip.io
```

but it would be much nice, if we could use:

```
VIRTUAL_HOST: kitchen-sink.(.*).xip.io
```

Is this even possible with nginx?

PS: Thanks for this wonderful code, it saves so much time and makes things really easy to setup.

schmunk42:
tested with PHP Web Application https://github.com/phundament/app

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,11 @@ If your container exposes multiple ports, nginx-proxy will default to the servic
 
 ### Multiple Hosts
 
-If you need to support multipe virtual hosts for a container, you can separate each entry with commas.  For example, `foo.bar.com,baz.bar.com,bar.com` and each host will be setup the same.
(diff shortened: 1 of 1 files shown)


jwilder:
Thanks!

md5:
:+1:

thaJeztah:
Nice!

PR #91, opened 2015-01-22
fixes SSL support while mixing HTTPS and non-HTTPS hosts

pirelenito:
nginx was throwing the following error: `no "ssl_certificate" is defined in server listening on SSL port while SSL handshaking`

fixes: https://github.com/jwilder/nginx-proxy/issues/74

the idea is to only create the 443 server if there is a default certificate available.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -126,16 +126,16 @@ server {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


devinrsmith:
This fix worked for me, thanks.

md5:
LGTM

jwilder:
Thanks!

Issue #92, opened 2015-01-24
Feature request: 301 redirect to www or non-www


jmillerdesign:
If you pass in multiple domains, it would be great to be able to 301 redirect the www subdomain to the naked domain. ```VIRTUAL_HOST: example.com,www.example.com```

johnjelinek:
I think you might want to specify a different syntax in that case. Perhaps `example.com->www.example.com` to set a 301 redirect.

md5:
You guys may want to comment on the implementation in #98.

jmillerdesign:
Thanks to #106 this is now possible. Here's how to do it:

* Create a volume for /etc/nginx/vhost.d
* In /path/to/vhost.d/www.example.com, add the contents: ```rewrite ^(.*) https://example.com$1 permanent;```

johnjelinek:
works!

skwasha:
I'm having an issue with this... My VIRTUAL_HOST is set to domain.com, www.domain.com. I have a blank file domain.com and one for www.domain.com with the line: 
> return 301 https://domain.com$request_uri;
in it. My run line is as follows:
> docker run -d -p 80:80 -p 443:443 --restart=always --volumes-from #{deploy[:application]} --name nginx-proxy -v /etc/nginx/vhost.d:/etc/nginx/vhost.d:ro -v /var/run/docker.sock:/tmp/docker.sock jwilder/nginx-proxy
I thought this should redir any www requests to non-www. but all I get are 502s. Any suggestions?

jmillerdesign:
Sounds like you're doing everything correctly. Perhaps there's an unrelated issue with the container you're running. What's the nginx.conf like? Do you see any errors there?

skwasha:
Are you referring to the generated one? I'm just using the default template with the addition of the vhost specific one-liner.

skwasha:
Seems to be some sort of issue with the volumes not mounting properly from the app container. I can see both /etc/nginx/certs and /etc/nginx/vhost.d being mounted in the app container and the nginx proxy container. But only the files from the certs dir are visible on the proxy container. 

The 502's were something else entirely...

ryneeverett:
I don't see how @jmillerdesign's strategy could work without setting `HTTPS_METHOD=noredirect` (which I wouldn't want to do because that would disable HSTS). Otherwise you get redirected to `https://www.example.com` before it gets to the vhost configuration file. Relevant parts of the template:

```
[...]

PR #98, opened 2015-02-11
added rewrite for 'www'-prefixed requests

mko-x:
After several users claiming not serving proxy via 'www'-prefixed URLs, added rewrite rule to template.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,6 @@
 FROM nginx:1.7.8
 MAINTAINER https://m-ko-x.de Markus Kosmal <code@m-ko-x.de>
 
-# set max size within a body
-ENV GLOB_MAX_BODY_SIZE "10m"
-# set default msg set within basic auth msg
-ENV GLOB_AUTH_MSG "Restricted :"
-# set default session timeout
-ENV GLOB_SSL_SESSION_TIMEOUT "5m"
-# default return code for errors
-ENV GLOB_HTTP_NO_SERVICE "503"
-
-# enable some kind of prefix redirection
-ENV AUTO_REDIRECT_WITH_PREFIX_ENABLED false
-# set prefix to be used for auto redirect
-ENV AUTO_REDIRECT_PREFIX "www"
-# set direction
-# - 0: redirect from prefix to non-prefix
-# - 1: redirect from non-prefix to prefix
-ENV AUTO_REDIRECT_DIRECTION 0
-
 # install packages
 RUN apt-get update \
  && apt-get install -y -q --no-install-recommends \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,24 @@
 FROM nginx:1.7.8
 MAINTAINER https://m-ko-x.de Markus Kosmal <code@m-ko-x.de>
 
+# set max size within a body
+ENV GLOB_MAX_BODY_SIZE "10m"
+# set default msg set within basic auth msg
+ENV GLOB_AUTH_MSG "Restricted :"
+# set default session timeout
+ENV GLOB_SSL_SESSION_TIMEOUT "5m"
+# default return code for errors
+ENV GLOB_HTTP_NO_SERVICE "503"
+
+# enable some kind of prefix redirection
+ENV AUTO_REDIRECT_WITH_PREFIX_ENABLED false
+# set prefix to be used for auto redirect
+ENV AUTO_REDIRECT_PREFIX "www"
+# set direction
+# - 0: redirect from prefix to non-prefix
+# - 1: redirect from non-prefix to prefix
+ENV AUTO_REDIRECT_DIRECTION 0
+
 # install packages
 RUN apt-get update \
  && apt-get install -y -q --no-install-recommends \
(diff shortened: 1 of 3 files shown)


md5:
I'm :-1: on this behavior being a default as you've implemented it.

I'm neutral on it existing as a feature that's triggered in some way by the value of `VIRTUAL_HOST`.

johnjelinek:
Ya, I'd prefer it to be configurable as an env var instead of hardcoded like this commit.

johnjelinek:
Also a 301 instead of rewrite

md5:
I suppose this and #92 are different things now that you point it out... So that's going to be even more confusing...

thaJeztah:
Why not handle the redirects in the "back-end" server/container? In my experience redirects `www->no-www` or `no-www->www` really differ per-application, I'm not sure this is something to handle by the proxy (which should IMO just be "transparent")

mko-x:
I don't know how to handle 'www.' prefixed requests in the backend if
they don't reach it due to nginx only listening for non-'www' ones. With
a 'www.' request the 503 at the top of the template will take place. 

Could you tell me how I could achieve that?

>From my point of view the most users of the provided services expect the
'www.' prefixed URLs to work just like the non-'www'  ones. I received a lot of claims about that. Additionally there are several browsers adding the 'www.' automatically.

Sebastiaan van Stijn <notifications@github.com> wrote:
> Why not handle the redirects in the "back-end" server/container? In my
> experience redirects `www->no-www` or `no-www->www` really differ
> per-application, I'm not sure this is something to handle by the proxy
> (which should IMO just be "transparent")
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/jwilder/nginx-proxy/pull/98#issuecomment-74628460

-- 
Sent using Mailpile, Free Software from www.mailpile.is

thaJeztah:
> I don't know how to handle 'www.' prefixed requests in the backend if they don't reach it due to nginx only listening for non-'www' ones.

You should be able to add both `www.` and `non-www` versions to your `VIRTUAL_HOST` environment var, for example;

    docker run --name my-website -d -e VIRTUAL_HOST=mysite.example.com,www.mysite.example.com nginx

The "my-website" container will now be accessible at both http://mysite.example.com and http://www.mysite.example.com

*Inside* the my-website container, you can handle what should happen for each (redirect one way or the other)

jmillerdesign:
Yeah, this commit forces non-www, which is bad.

What about this? Does this make too many assumptions, or would it be helpful enough in a large majority of projects?
If VIRTUAL_HOST=mysite.example.com, then automatically also include www.mysite.example.com and redirect it to mysite.example.com.
If VIRTUAL_HOST=www.mysite.example.com, then automatically also include mysite.example.com and redirect it to www.mysite.example.com.
If VIRTUAL_HOST=mysite.example.com,www.mysite.example.com then don't do any redirects.

thaJeztah:
> Does this make too many assumptions, or would it be helpful enough in a large majority of projects

It certainly wouldn't help *me*; thinking of `api.domain.com`, `static.domain.com` and other subdomains, sometimes pointing to a different container/back-end.

In general, I don't think a proxy should make *any* assumptions, unless I explicitly specify it in the config-file or there's a config-parameter to specify this.

just my 0.02

mko-x:
Ah, thank you. 

Somehow I didn't think of configuring it within the VIRTUAL_HOST env var. I will give it a try and if it works like expected this pull request might be superfluous.

Anyway I agree with johnjelinek in the mentioned points here and #92 . 
- Solution should work with 'Moved Permanently' 301 
- For my IMHO not that uncommon usecase - provide an env var flag or similar to configure global redirect handling at will, reducing redundancy and improving readability at least a bit
- From #92 : To set a redirect policy - global flag may be some kind of enum 
´´´AUTO_REDIRECT_TARGET{
   www,
   non-www,
   custom,
   off
}´´´

Defaulting to 'off' for backwards compatibility. With following effects:
- www: redirect non-www to www
- non-www: redirect www to non-www
- custom: redirect with custom prefix, as an idea e.g. for APIs, CDNs or what I see once a while 'www8' at amazon urls if I remember correctly + new env var like 'AUTO_REDIRECT_CUSTOM_TARGET' to set the value, non-www will be redirected as well if not prefixed with the custom value
- off: as default the behaviour like it is now, no auto redirecting

If I find time, I would change that in my fork and create a new pull request. 

Maybe someone has another better solution, will wait a while for feedback.

jmillerdesign:
I still think it could be fine to automatically redirect ```www.api.example.com``` to ```api.example.com```, if your ```VIRTUAL_HOST=api.example.com```. I can understand the points though that perhaps it shouldn't do anything automatically. I'd be OK with using some sort of ```AUTO_REDIRECT_TARGET``` flag if you'd prefer that.

jwilder:
I think #106 should resolve this.  You can add `client_max_body_size` and the rewrite in the custom include now and have your container's `VIRTUAL_HOST` var be the list all the domains it should work with.

johnjelinek:
LGTM!

Issue #87, opened 2015-01-15
Per VIRTUAL_HOST custom includes


Nitesedge:
How hard would it be to add support for a custom include file per host? Much like the htpasswd where if the file exists for a virthost it gets included in the config?

 I'm thinking along the lines of custom rewrite rules per container as well as other custom includes.

I'm playing with the configs now, but am not savvy with nginx or the template language. Any pointers would be appreciated!

Nitesedge:
I'm thinking something like this: http://pastebin.com/51SD6Ph5

Thoughts?

md5:
@Nitesedge I think your implementation of this looks fine. I would probably call the directory `/etc/nginx/proxy.d` instead of `/etc/nginx/includes`, but this is exactly what I was going to implement when I thought I should look through the existing issues and PRs one more time.

thaJeztah:
+1 would be nice to have a PR for this!

md5:
I ended up opening my own PR for this at #106 since I had some time and wanted to include some docs.

md5:
This can be closed since #106 was merged. See the README for instructions: https://github.com/jwilder/nginx-proxy/blob/master/README.md#custom-nginx-configuration

PR #76, opened 2015-01-02
Set nginx client_max_body_size for containers exporting MAX_UPLOAD_SIZE.

jperville:
This PR is my minimal attempt at solving #29 and #39 .

It extends the nginx template to add exactly one `client_max_body_size` directive for each virtual host that export the `MAX_UPLOAD_SIZE` variable on at least one upstream container.

Issue #71, opened 2014-12-22
Custom Templates


nVitius:
Would it be possible to add support for using custom templates for a subdomain?

captncraig:
This is actually possible now. 

docker run .....  -v /path/to/my.tmpl:/app/nginx.tmpl jwilder/nginx-proxy

PR #70, opened 2014-12-17
Adds additional configuration options for NGINX

wjdix:
I needed to add SSL client authentication and a few other options for an application I'm working on. This adds the ability to specify client auth, as well as any generic directive into the location declaration.

PR #106, opened 2015-02-22
Per VIRTUAL_HOST configuration files

### LINKED ISSUE #29: Allow to set the value for client_max_body_size
I'm trying to make ownCloud working and everything work great until I try to upload a file of more than 3 Mb.

It seems that the instruction `client_max_body_size` is missing.

It would be nice to add another variable to manipulate the `client_max_body_size` value.

md5:
This PR adds the ability to configuration Nginx settings on a per `VIRTUAL_HOST` basis by adding a configuration file under `/etc/nginx/vhost.d`. As in the case of SSL configuration and basic authentication, the files in this directory are expected to be named after the `VIRTUAL_HOST` value verbatim (without `.conf`).

I also added some documentation around providing custom Nginx configuration on a proxy-wide basis. The whole README.md is starting to feel pretty messy to me, so it may make sense at some point to do an editing pass on it as a whole or to move large parts of it into wiki pages.

Fixes #29, #36, #39, #70, #76, #87 
Related #71

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,3 +114,36 @@ In order to be able to securize your virtual host, you have to create a file nam
     $ docker run -d -p 80:80 -p 443:443 -v /path/to/htpasswd:/etc/nginx/htpasswd -v /path/to/certs:/etc/nginx/certs -v /var/run/docker.sock:/tmp/docker.sock jwilder/nginx-proxy
 
 You'll need apache2-utils on the machine you plan to create de htpasswd file. Follow these [instructions](http://httpd.apache.org/docs/2.2/programs/htpasswd.html)
+
+### Custom Nginx Configuration
+
+If you need to configure Nginx beyond what is possible using environment variables, you can provide custom configuration files on either a proxy-wide or per-`VIRTUAL_HOST` basis.
+
+#### Proxy-wide
+
+To add settings on a proxy-wide basis, add your configuration file under `/etc/nginx/conf.d` using a name ending in `.conf`.
+
+This can be done in a derived image by creating the file in a `RUN` command or by `COPY`ing the file into `conf.d`:
+
+```Dockerfile
+FROM jwilder/nginx-proxy
+RUN { \
+      echo 'server_tokens off;'; \
+      echo 'client_max_body_size 100m;'; \
+    } > /etc/nginx/conf.d/my_custom_proxy.conf
+```
+
+Or it can be done by mounting in your custom configuration in your `docker run` command:
+
+    $ docker run -d -p 80:80 -p 443:443 -v /path/to/my_custom_proxy.conf:/etc/nginx/conf.d/my_custom_proxy.conf:ro -v /var/run/docker.sock:/tmp/docker.sock jwilder/nginx-proxy
+
+#### Per-VIRTUAL_HOST
+
+To add settings on a per-`VIRTUAL_HOST` basis, add your configuration file under `/etc/nginx/vhost.d`. Unlike in the proxy-wide case, which allows mutliple config files with any name ending in `.conf`, the per-`VIRTUAL_HOST` file must be named exactly after the `VIRTUAL_HOST`.
+
+In order to allow virtual hosts to be dynamically configured as backends are added and removed, it makes the most sense to mount an external directory as `/etc/nginx/vhost.d` as oppposed to using derived images or mounting individual configuration files.
+
+For example, if you have a virtual host named `app.example.com`, you could provide a custom configuration for that host as follows:
+
+    $ docker run -d -p 80:80 -p 443:443 -v /path/to/vhost.d:/etc/nginx/vhost.d:ro -v /var/run/docker.sock:/tmp/docker.sock jwilder/nginx-proxy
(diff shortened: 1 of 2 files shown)


thaJeztah:
Wondering, if I defined multiple domains (e.g. `VIRTUAL_HOST=www.foo.com,foo.com`), will I have to define a configuration file for each separately?

md5:
Yes. I thought about clarifying that in the docs. It's the same situation
as with certs and htpasswd.

md5:
Docs updated.

md5:
I'm not sure what to recommend for people using regex-based or other more complex `VIRTUAL_HOST` configurations, but at least this can handle the common cases.

thaJeztah:
Ah, I *thought* it would follow the same rules, wasn't entirely sure.

Thanks a lot for this contribution, I think it will be very useful feature and I like this a lot more than having special "flags" or "environment-vars" for each feature!

md5:
While I was working on this PR, I started having ideas about somehow using `docker cp` to grab the custom config from `/etc/nginx-proxy` or some such inside the backing containers themselves. This would run into issues if multiple containers with the same `VIRTUAL_HOST` had different conf files, but it would make it really easy to plug in backends with custom configuration without having to have a separate step to put a config file into `/etc/nginx/proxy.d` for the `nginx-proxy` container.

jwilder:
This looks good.  Thanks for updating the docs too.  I agree, the README is getting a little messy.

For the custom config files, two questions:

1. What do you think about having the custom config file actually be a template that is rendered instead of a static file that is included?  We'd probably need a custom template func to compile and render the template and pass along the current context.
2. Instead of just including/rendering a config file within the server block, we could let you generate the whole server block for that virtual host via the template.  The template that is used would have to have to add the location and proxy_pass directives and we'd have to document what the bare minimum template needs to include.  I think that would let you configure just about anything (e.g. custom paths, SSL criteria, etc..)

Thoughts?

md5:
I actually started looking into implementing the necessary function for `docker-gen` a few months ago, but never progressed much with it: https://github.com/md5/docker-gen/compare/include-tmpl

Stepping back a bit, I think that both of your proposed changes sound like the sort of things that could lead to an increased support burden without substantial benefits.

In the case of the first change, users of `nginx-proxy` will be exposed directly to writing Go templates themselves, something they aren't expected to do currently. The benefit to be gained is that users will be able to do things like `client_max_body_size {{ Env.MAX_UPLOAD_SIZE }};` in their own config files instead of having to generate or otherwise maintain the config files directly.

In the case of the second change, users will gain total control over the `server` block, but I think with the per-`VIRTUAL_HOST` config files it will be possible to override anything being done in `nginx.tmpl` anyways. I think at a certain point, users need to roll their own Nginx proxy image if they want to have complete control of everything.

jwilder:
Good points.  

Not everything is overideable with this approach I think.  I don't think you could override the `location / {}` block with the include.  I believe nginx will complain about about duplicate entries.  I know there have been some requests for virtual path support so that still needs some other hooks to resolve.

The include is for both SSL and non-SSL.  I'm not sure if it matters but adding SSL options would get included in the non-SSL block.  Maybe that can be handled w/ having a `.ssl` suffix to the vhost file and `nginx.tmpl` would look for that and fallback to the main one if it doesn't exist?   Or perhaps add two separate includes?

There's been other requests to not redirect http->https.  Not sure if an include could resolve that issue.  

I still think being able to generate the server block is useful.  It could actually be a separate feature independent of this.  The majority of the cases could be handled w/ this PR and anything not covered could be handled w/ a custom vhost template.  Maybe a naming conventions like: `example.com.tmpl`.  If that exists, use that for the server block.  This would be the advanced/you need to understand go templates option but should negate the need to fork and maintain a custom image.

md5:
I should have been clearer when I was talking about overriding. I was only referring to the other things that are currently set in the `server {}` block other than the `location / {}` part.

Implementing virtual path support requires grouping the backends that serve a particular path for a particular host into their own upstream and connecting each upstream to the appropriate `location /path {}` block. I started on some code for that at one point, but I haven't kept it up-to-date with other changes `nginx.tmpl`: https://github.com/md5/nginx-proxy/compare/multiple-paths

Regarding the SSL thing, I did consider the fact that users might want to use a different config file for the HTTP and HTTPS blocks, but I wasn't sure what the right approach would be to support that. I'd probably lean toward supporting a `/etc/nginx/vhost.ssl.d` directory instead of a suffix approach, but I'm happy to update the patch either way. Perhaps @thaJeztah or someone else might have an opinion on the issue.

I like the idea of having the custom Go template thing be able to replace the entire `server {}` block. That way, users that take that approach will be 100% responsible for setting it up the way they want (with some reasonable docs as a starting point) if they decide to go the "power user" route.

jwilder:
@md5 I think we can punt on separate SSL configs for now and address it later if there is an issue.

md5:
:+1: to dealing with the SSL thing if anyone complains about needing separate configs

curtiszimmerman:
Awesome!

PR #108, opened 2015-02-25
Fix HTTP->HTTPS redirect for wildcard hosts

md5:
Uses Nginx's $host instead of interpolating `{{ $host }}` in the template

Refs #92


jwilder:
Thanks!

md5:
:+1:

PR #111, opened 2015-02-28
Remove includeSubdomains from HSTS header

### LINKED ISSUE #109: remove includeSubdomains
https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl#L108

I think includeSubdomains should be removed here.  This bit me pretty hard without realizing that dev.app.com and stage.app.com would be automatically redirected to ssl too and I don't have a cert for those urls.

jwilder:
includeSubdomains can lead to issues where not all subdomains are
able to use HTTPS.  This options might be too strict for the general
case: https://www.owasp.org/index.php/HTTP_Strict_Transport_Security.
It can be re-enabled w/ a custom template if needed.

Fixes #109

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -105,7 +105,7 @@ server {
 	ssl_certificate /etc/nginx/certs/{{ (printf "%s.crt" $cert) }};
 	ssl_certificate_key /etc/nginx/certs/{{ (printf "%s.key" $cert) }};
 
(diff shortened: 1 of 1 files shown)


md5:
I'd possible remove STS entirely from the default configuration and document how to add it back on a per-`VIRTUAL_HOST` basis.

Otherwise, this change LGTM

PR #124, opened 2015-03-14
Support regex VIRTUAL_HOST values

### LINKED ISSUE #117: Regular expression virtual host
I'm try to use regex for domain matching (*-staging.my-domain.com)

```
-e VIRTUAL_HOST="~^.*\-staging\.my\-domain\.com$"
```

Any idea why it does not work?

md5:
This will be a breaking change for anyone who was using the SSL, Basic Auth, or custom template functionality with a hostname of "`*.domain.tld`". Their extension files will now need to be named using "`%2A.domain.tld`" instead of using a literal asterisk. It shouldn't break any other cases that were previously working, but that may be a mistaken assumption on my part.

Fixes #117

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -70,8 +70,9 @@ server {
 {{ end }}
 
 {{ range $host, $containers := groupByMulti $ "Env.VIRTUAL_HOST" "," }}
+{{ $upstream := queryEscape $host }}
 
-upstream {{ $host }} {
+upstream {{ $upstream }} {
 {{ range $container := $containers }}
 	{{ $addrLen := len $container.Addresses }}
 	{{/* If only 1 port exposed, use that */}}
@@ -97,7 +98,7 @@ upstream {{ $host }} {
 {{ $certName := (first (groupByKeys $containers "Env.CERT_NAME")) }}
 
 {{/* Get the best matching cert  by name for the vhost. */}}
-{{ $vhostCert := (closest (dir "/etc/nginx/certs") (printf "%s.crt" $host))}}
+{{ $vhostCert := (closest (dir "/etc/nginx/certs") (printf "%s.crt" $upstream))}}
 
 {{/* vhostCert is actually a filename so remove any suffixes since they are added later */}}
 {{ $vhostCert := replace $vhostCert ".crt" "" -1 }}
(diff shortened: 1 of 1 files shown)


md5:
@abdollahpour @danhealy If you've got some time, check out these changes to see if they address the problems you had with regex-based `server_name` values in `VIRTUAL_HOST`.

danhealy:
I don't think this is working for me.

    ~^example.*

works, but these don't:

    ~^example\.com(\..*\.xip\.io)?$


    ~^example.*$


    ~^example.com.192.168.59.103.xip.io$


    $ docker exec nginx-proxy sed -n '80p' /etc/nginx/conf.d/default.conf
        proxy_pass http://~^example.com.192.168.59.103.xip.io$;

danhealy:
:+1: Awesome, works perfectly for me.  Sorry about the confusion.

md5:
> :+1: Awesome, works perfectly for me. Sorry about the confusion.

:metal:

md5:
@jwilder any thoughts on this change? I can't imagine there are a ton of people out there with custom configs, certs, etc, named "`*.mydomain.com`" for whom this will be a breaking change, but perhaps that's more common than I think.

md5:
If we really want to avoid the breaking change, we could do something like `{{ $upstream := replace (queryEscape $host) "%2A" "*" -1 }}`.

jwilder:
@md5 Have not had a chance to try this yet.  Does this mean that files will need to be named `%2a.domain.tld`?  That might trip people up.  Any way to avoid the escaping in the filenames on the filesystem?

md5:
I think that “*"/“%2A" is probably the only escape sequence that could
affect existing users, but I could be forgetting some other common
server_name chars that would have worked previously instead of being a
syntax error. Assuming that's the case, the “replace" workaround I
mentioned should be fine.

md5:
In terms of not escaping on the fs at all, I suppose the only character
that would matter is “/", which makes no sense in “server_name", even in a
regex. I suppose “/", could be a problem in the abstract, but we aren't
running on Windows.

md5:
I should have said `\` in the last comment in regards to Windows. That being said, backslashes will probably be problematic regardless of Windows-ness since they're pretty common in regexes on their own.

md5:
@jwilder Are there any changes you want made to this PR? I think this functionality is still needed to support regex virtual hosts.

jwilder:
@md5 I agree that the regex functionality would be useful but I'm not keen on the backward incompatibility and the query escaping in the file names.  Unfortunately, I don't have any other ideas right now though.

md5:
@jwilder What about if we deal with `*/%2A` as a special case as I mentioned in this comment: https://github.com/jwilder/nginx-proxy/pull/124#issuecomment-87013087

md5:
TBH, I thought I had already incorporated that change into the patch, but I guess we just talked about it :expressionless:

md5:
I just pushed a commit to avoid breaking changes for the `*.example.com` case. This probably needs some more testing though. Perhaps it can wait until the test suite that @thomasleveil is working on makes its way into a PR.

devvyn:
FWIW, this patch made my day. Works nicely, so thanks! 

I threw this crazy virtual host argument into ``docker-compose.override.yml`` and nginx didn't barf this time:

```YAML
VIRTUAL_HOST: "~^((m|www?)\\.)?otv(te\\.ch|\\.technology|tech\\.(ca|com|dev|app))$$"
```

md5:
@devvyn Glad to hear it! :metal: 

I'm happy to update this to fix the conflicts if @jwilder is interested in merging it. It looks like our last discussion was around the potential for breaking changes. Since we now have a test suite, perhaps I can add some tests if there's agreement on merging this functionality.

Brah0um:
Hi everyone,
When will this patch be merge?

sakoht:
Nice.  This would have kept me from having to fork and hack up the template recently.

rjcf18:
Hello everyone, is this PR going to be merged at all? 
I think this functionality is much needed imo

buchdag:
superseded by #671

PR #132, opened 2015-03-25
Add missing volumes

brikou:
Allow the use `--volumes-from` parameter.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,6 @@ WORKDIR /app/
 
 ENV DOCKER_HOST unix:///tmp/docker.sock
 
-VOLUME ["/etc/nginx/certs"]
+VOLUME ["/etc/nginx/certs", "/etc/nginx/htpasswd", "/etc/nginx/vhost.d"]
 
 CMD ["forego", "start", "-r"]


RobinBressan:
:+1:

md5:
I'm :-1: on this change. It's not possible to "undo" a `VOLUME` statement and not all users will want or need them. Setting up a volume on your container with a `-v` allows you to use `--volumes-from` just as easily.

```console
$ docker run -d -v /etc/nginx/htpasswd -v /etc/nginx/vhost.d -v /var/run/docker.sock:/tmp/docker.sock -p 80:80 -p 443:443 --name nginx-proxy jwilder/nginx-proxy
165a3df710e84a3d53b7ff502f3f7e19c8c414bb73ecbb3b83629404771281eb
```
```console
$ docker run --rm --volumes-from nginx-proxy debian:jessie mount | grep nginx
/dev/sda1 on /etc/nginx/certs type ext4 (rw,relatime,data=ordered)
/dev/sda1 on /etc/nginx/htpasswd type ext4 (rw,relatime,data=ordered)
/dev/sda1 on /etc/nginx/vhost.d type ext4 (rw,relatime,data=ordered)
/dev/sda1 on /var/cache/nginx type ext4 (rw,relatime,data=ordered)
```

RobinBressan:
I don't understand why this is a problem. IMO there is no need to undo, it is just in case you need to access it from an another container (otherwise the `VOLUME` command in a dockerfile would be useless).

And if we do what you suggest, why should we keep the `/etc/nginx/certs` into the Dockerfile? It would not be very consistent.

md5:
In your opinion there's no need to undo, but for people who don't want the volume there would be a need (and it wouldn't be possible).

I personally think that `VOLUME` should be restricted to things that are necessary to the operation of the image and have a need to read and write significant amounts of data. Other than that, documentation generally should suffice (possibly including a sample `docker-compose.yml` with some added `volumes`).

FWIW, I'm not a fan of the current volume for `/etc/nginx/certs` either, but it's already there. And ultimately, it's up to @jwilder whether there should be more volumes since it's his project.

brikou:
@md5 Can you explain us what is the benefice of undoing a volume? Is there IO benefits? If so we should definitively remove all `VOLUME` from all dockerfiles, don't you think so?

@RobinBressan totally agree with your point of view (unless @md5 make me change my mind with clever explanations) lol

thaJeztah:
I'm -1 on this change

I *do* understand the use-case, but this can be easily achieved, either by the example that @md5 showed above or a simple two-line Dockerfile;

```Dockerfile
FROM jwilder/nginx-proxy:latest
VOLUME ["/etc/nginx/certs", "/etc/nginx/htpasswd", "/etc/nginx/vhost.d"]
```

The *opposite*, however, won't be possible (disabling the volume), which would prevent users from extending the image to ship it with pre-defined configurations in those directories.

In addition, even though volumes are very useful in many cases, they can also be a major PITA; removing a container *without* the `-v` option causes the volumes to be left behind after the container was removed (and are difficult to get rid of).

I'm not a big fan of the current `VOLUME` either, but removing that would break BC.

jwilder:
The current volume that is there was a mistake and I think it would be better to remove it rather than add others to it.

Issue #47, opened 2014-10-26
Using multiple path


munjalpatel:
Hello,

I have modified your nginx template a bit so that I can use different docker containers for different paths of the same domain.

For example: 
    www.mysite.com with /        ---> web-main container
    www.mysite.com with /auth ---> web-auth container

But I am not sue how I can control order in which these values are registered in sites-emabled nginx config file.

Currently, config block for "/" gets defined before "/auth" block and so all of my requests ends up going to the "web-main" docker container.

Basically, it would need to sort the collection of the "path" in descending order before looping over it to generate the config file.

I would appreciate if anyone can point me into the right direction here.

Here is the modified nginx template:

```
server {
    listen 80 default_server;
    server_name _; # This is just an invalid value which will never trigger on a real hostname.
    error_log /proc/self/fd/2;
    access_log /proc/self/fd/1;
    return 503;
}

{{ range $host, $containers := groupByMulti $ "Env.VIRTUAL_HOST" "," }}
{{ range $path, $containers := groupByMulti $ "Env.VIRTUAL_PATH" "," }}
{{ range $index, $value := $containers }}

upstream {{ $value.Name }} {
    {{ $addrLen := len $value.Addresses }}
    {{/* If only 1 port exposed, use that */}}
    {{ if eq $addrLen 1 }}
        {{ with $address := index $value.Addresses 0 }}
           server {{ $address.IP }}:{{ $address.Port }};
        {{ end }}

    {{/* If more than one port exposed, use the one matching VIRTUAL_PORT env var */}}
    {{ else if $value.Env.VIRTUAL_PORT }}
        {{ range $i, $address := $value.Addresses }}
           {{ if eq $address.Port $value.Env.VIRTUAL_PORT }}
           server {{ $address.IP }}:{{ $address.Port }};
           {{ end }}
        {{ end }}

    {{/* Else default to standard web port 80 */}}
    {{ else }}
        {{ range $i, $address := $value.Addresses }}
            {{ if eq $address.Port "80" }}
            server {{ $address.IP }}:{{ $address.Port }};
            {{ end }}
        {{ end }}
    {{ end }}
}

server {
    listen 443 ssl;

    ssl_certificate /etc/nginx/certs/nginx.crt;
    ssl_certificate_key /etc/nginx/certs/nginx.key;

    gzip_types text/plain text/css application/json application/x-javascript text/xml application/xml application/xml+rss text/javascript;

    server_name {{ $host }};
[...]

PR #145, opened 2015-04-14
Support static file serving

iddo:
Easy to use static file serving alongside the reverse proxy using the 'try_files' directive with named location technique

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,3 +158,19 @@ If you are using multiple hostnames for a single container (e.g. `VIRTUAL_HOST=e
 
     $ { echo 'server_tokens off;'; echo 'client_max_body_size 100m;'; } > /path/to/vhost.d/www.example.com
     $ ln -s www.example.com /path/to/vhost.d/example.com
+
+### Static files serving
+
+In order to server static files in conjuction with reverse proxying the virtual host simply mount the static files directory to `/srv`
+
+For example, if you have a virtual host named `app.example.com`, you can server the static files from the host folder `/home/ubuntu/app/static` using:
+
+    $ docker run -d -p 80:80 -p 443:443 -v /srv/app.example.com:/home/ubuntu/app/static:ro -v /var/run/docker.sock:/tmp/docker.sock jwilder/nginx-proxy
+
+If you need to serve multiple hosts simply mount the /srv folder and use symlinks on the host:
+
+    $ mkdir -p /home/ubuntu/static
+    $ cd /home/ubuntu/static
+    $ ln -s app.example.com /home/ubuntu/app/static
+    $ ln -s app2.example.com /home/ubuntu/app2/static
+    $ docker run -d -p 80:80 -p 443:443 -v /srv:/home/ubuntu/static:ro -v /var/run/docker.sock:/tmp/docker.sock jwilder/nginx-proxy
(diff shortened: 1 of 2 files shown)


kytwb:
Can this get merged?

iddo:
It's pretty straight forward, so I definitely think so, but I'm not the one calling the shots around here :-)

kytwb:
Ping @jwilder

thomasleveil:
shouldn't this proxy project stay focused on its mission: _Automated nginx proxy for Docker containers using docker-gen_?

This would add a new role (webserver) to this project. I like projects that do less but do it right (UNIX philosophy). 

If one needs to serve static files, I suggest the use of a container that does the job of serving file in combination with nginx-proxy and not changing nginx-proxy

iddo:
Normally I would agree with what @thomasleveil wrote, "use specialized tools for every need", but there aren't that many static file serving options out there (https://nbonvin.wordpress.com/2011/03/14/apache-vs-nginx-vs-varnish-vs-gwan/) and this is quite a common scenario.
Achieving this kind of configuration (static files with fallback to the reverse proxy) will still require specialized nginx (or other) configuration even if we were to use a specialized static file server.

thaJeztah:
Now that nginx-proxy supports per virtual-host configuration files, I don't really see a reason to add this to the default template. Also, it's possible to replace the default template either by bind-mounting it as a volume, or creating a simple Dockerfile that extends the image.

I'm sure there are people that want this, but not *everyone*. How about an "examples" directory in the repository, accompanied by a README.md that explains what each example does?

iddo:
@thaJeztah, I'm using the bind-mount technique with my suggested template modification, but given the default behavior does not change for people who don't need/use this feature I think it will be a welcomed addition.

md5:
I'm on the fence on this one. Setting up `nginx-proxy` + an app + another `nginx` can be kind of tricky in terms of getting the virtual host stuff right. Without multiple paths supports (cf. #47), you'd either have to only use `VIRTUAL_HOST` on the `nginx` container that is itself proxying to the app or you'd have to arrange for the static assets to be served from a separate hostname.

jwilder:
I agree with @thaJeztah and @thomasleveil .  Since it is possible to do this with the current features, I think it's better to keep the project focused on a being a proxy for now and add examples or documentation on how to set this up if needed.  Static file serving will likely require other customization hooks for cache control, compression, etc. that I would rather not support right now.

tanzaho:
Can anyone give some hints about how to proceed ?

thomasleveil:
See https://github.com/jwilder/nginx-proxy/blob/master/README.md#per-virtual_host

tanzaho:
I followed the instructions, and I tried to add the following to nginx.
``` 
location /static/ {
	/usr/src/app/static/;
    }
``` 
These where added as a volume named as the host.

When I restarted nginx_proxy, the file was used, but the behavior is not the one I expected : 
* the index page is not showing my app anymore, instead it shows the "Welcome to nginx". 
* the /static/ folder returns a 404.

I know that my question might be related to nginx, but if anyone can point me toward the good direction, that'd very nice.

PS : I am using docker-compose to run my containers.

pascalandy:
@tanzaho Hey Jean-Christophe! May I asked you back to question you asked a year ago ?

> I know that my question might be related to nginx, but if anyone can point me toward the good direction, that'd very nice.

Issue #121, opened 2015-03-12
support both: http and https when valid SSL cert is available


Finkes:
I'm using a self-signed SSL certificate and I want to support both: http and https to access my websites. The reason for that is that I have a jenkins server running inside a dicker container and it is not receiving any hook from github or bitbucket. Currently I always geht redirected to https when accesing my server. 
Maybe this could be done with a env variable (e.g. FORCE_HTTPS=FALSE) when starting nginx reverse proxy.

Finkes:
For now I just use two reverse proxy: one is listening on port 80 and one only on port 443
>docker run --name reverseproxy_ssl -d -p 443:443 -v [...]
>docker run --name reverseproxy -d -p 80:80 -v [...]

jakeslee:
Yes, I think this is very necessary. If I add a wildcard certificate to root domain, all subdomain will be forced to redirect to HTTPS. It is not my intent. Maybe setting a ENV to support two situations is better.

atom2ueki:
@Finkes May I know how you let your new deployed container choose one of your reverse proxy ?

Finkes:
I just serve all the containers via both proxies (http and https). I don't know how to achieve that only one reverseproxy will serve your container.

ryneeverett:
Resolved #298's `HTTPS_METHOD=noredirect`?

PR #155, opened 2015-05-02
Added SSL_REDIRECT environment option to allow http and https to run in ...

itsafire:
...parallel. Merged http and https server block to one according to nginx documentation. See #121


jwilder:
This removes http auth and custom includes.  I'm also not in in favor of adding more env vars since there are other ideas being discussed that could handle this case.

PR #156, opened 2015-05-02
configure container as backup server

christophwitzko:
Hi Jason,

thank you very much for this awesome docker image.
I found out that it is not possible to configure a container as [backup server](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#server).
I added this feature in this pull request.

have a nice day
Christoph


christophwitzko:
Ping @jwilder

jwilder:
I'd rather not add new env vars because it can easily turn into a mess. You should be able to use a custom template to do this right now though.

christophwitzko:
ok. thank you.

PR #161, opened 2015-05-08
added info how to modify nginx config in startup command

schmunk42:
As described in the paragraph, this is a temporary solution, only for debugging.

If it's too hacky, just close the pull-request - I don't mind.

...but this was a real life saver for me ;)


md5:
This looks like a good candidate for the wiki. Looks like there aren't any pages there yet.

Once there are some wiki pages, I think that adding a link to the wiki in `README.md` would be great.

kevin-coyle:
Just came across an issue where this fix would have been nice. Any ETA on when it can be merged in?

PR #163, opened 2015-05-09
Close marker instead of empty

mko-x:
If a client is not providing the Upgrade HTTP header with the request, the connection will now be marked as 'to close' when finished.
See [#46](https://github.com/jwilder/nginx-proxy/pull/46) for more info.

Added a .gitignore by accident or out of habit as well.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -9,7 +9,7 @@ map $http_x_forwarded_proto $proxy_x_forwarded_proto {
 # Connection header that may have been passed to this server
 map $http_upgrade $proxy_connection {
   default upgrade;
-  ''      '';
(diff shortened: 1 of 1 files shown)


md5:
I'd suggest using a global `core.excludesFile` for your own machine that ignores `.DS_Store`. See here: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

jwilder:
@mko-x can you drop the gitignore commit?

mko-x:
@jwilder .gitignore is dropped

@md5 thank you, I will try host global ignore

jwilder:
Thanks!

PR #168, opened 2015-05-19
Fix a couple things in the README.

bprodoehl:
As it is written, the "Separate Containers" section of the README doesn't work.  I added the jwilder repository to the docker-gen image name, assuming that the user won't have a local build named docker-gen, and also changed the docker-gen behavior from -only-published to -only-exposed, to match the nginx-proxy container's behavior.


jwilder:
Thanks!

PR #170, opened 2015-05-21
Add support for ssl_dhparam to prevent 'Logjam' attack

kcyeu:
# Description
Added detection for DH group by site in the format ```<hostname>.dhparam.pem```, if file exists then use it.

# Reference 
* [The Logjam Attack](https://weakdh.org/)
* [Guide to Deploying Diffie-Hellman for TLS](https://weakdh.org/sysadmin.html)
* [CVE-2015-1716](http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-1716)

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,8 +86,8 @@ hosts in use.  The certificate and keys should be named after the virtual host w
 #### Diffie-Hellman Groups
 
 If you have Diffie-Hellman groups enabled, the files should be named after the virtual host with a
-`dhparams` suffix and `.pem` extension. For example, a container with `VIRTUAL_HOST=foo.bar.com`
-should have a `foo.bar.com.dhparams.pem` file in the certs directory.
+`dhparam` suffix and `.pem` extension. For example, a container with `VIRTUAL_HOST=foo.bar.com`
+should have a `foo.bar.com.dhparam.pem` file in the certs directory.
 
 #### Wildcard Certificates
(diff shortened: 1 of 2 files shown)


md5:
It looks like the indentation is off in that diff. Also, I think being consistent with the parameter name and using `dhparam.pem` instead of `dhparams.pem` would be better.

I think we may also want to make sure that export ciphers are not turned on by default, though it may be best for that fix to occur in the `nginx` container.

Otherwise, this LGTM.

viranch:
I'd also like to see the support for a shared `dhparams.pem` file, just like the `CERT_NAME` option that the image offers for shared cert.

Also, shouldn't `ssl_prefer_server_ciphers on;` be included too?

md5:
@viranch It already is: https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl#L101

viranch:
Oops, missed that one.

md5:
Looks like `ssl_ciphers` already has `!EXPORT` as well. The cipher list should probably still be reviewed, but that seems like it could be a separate PR.

LGTM

md5:
BTW, @kcyeu, you might want to leave a comment when you update a PR since GitHub doesn't send a notification.

jwilder:
Thanks @viranch @md5!

PR #174, opened 2015-06-05
Encourage using read-only mount for docker.sock

### LINKED ISSUE #173: Mount docker.sock readonly?
Is there any reason why we wouldn't want to mount the docker.sock as Read Only? 

`docker run -d -p 80:80 -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy`

rtrauntvein:
Changes to README.md to encourage use of a read-only mount for /var/run/docker.sock

Closes #173

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ See [Automated Nginx Reverse Proxy for Docker][2] for why you might want to use 
 
 To run it:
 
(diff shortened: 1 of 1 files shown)


jwilder:
Thanks!

PR #178, opened 2015-06-09
Cleaner template for default host

viranch:
Quick commit to clean up the long `if-else` block added to support `DEFAULT_HOST`.


jwilder:
This whole block might get removed via #202

PR #186, opened 2015-06-24
Prevent generating broken config

### LINKED ISSUE #115: Prevent Generating Broken Config
https://github.com/jwilder/nginx-proxy/issues/107 resulted in a invalid nginx.conf because a container exposed two ports but neither of them were port 80 and a `VIRTUAL_PORT` env var was not specified.  The generated nginx.conf had a backend section w/ no endpoints listed causing nginx to fail to reload.

The issue is that https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl#L62 should have an else clause that generates `server {{ $address.IP }} down;` per https://github.com/jwilder/nginx-proxy/issues/107#issuecomment-75837337 to make it more explicit that there is a problem.

If https://github.com/jwilder/docker-gen/issues/67 is implemented, it should also be used.

md5:
Fixes #115

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN echo "daemon off;" >> /etc/nginx/nginx.conf \
 RUN wget -P /usr/local/bin https://godist.herokuapp.com/projects/ddollar/forego/releases/current/linux-amd64/forego \
  && chmod u+x /usr/local/bin/forego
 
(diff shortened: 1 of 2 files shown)


jwilder:
Thanks @md5!

md5:
:+1:

PR #192, opened 2015-07-04
Use container host's IP:port if we're connected to a swarm master

viranch:
This PR completes support for Swarm cluster.

This will also pick up any containers running on a remote host that is in the Swarm cluster and uses the host address/port instead of container address/port.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -44,24 +44,40 @@ upstream {{ $host }} {
 	{{ $addrLen := len $container.Addresses }}
 	{{/* If only 1 port exposed, use that */}}
 	{{ if eq $addrLen 1 }}
-		{{ with $address := index $container.Addresses 0 }}
-		   # {{$container.Name}}
-		   server {{ $address.IP }}:{{ $address.Port }};
+		{{ $address := index $container.Addresses 0 }}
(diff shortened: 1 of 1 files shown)


md5:
LGTM

That being said, this section of the template is getting to be pretty unmaintainable. It might be nice to use `{{define}}` and `{{template}}` to streamline some of the repetitive parts, though I'm not sure how much that would help.

viranch:
@md5 +1.

I started off by assigning the address object to a `chosenAddr` variable and doing swarm logic outside the if-else, but couldn't get it to work.

viranch:
Ping.

Added `{{define}}` & `{{template}}` for upstream block. Also a simplification to `coalesce .Env.VIRTUAL_PORT "80"` to reduce an `else` block (check commit diffs).

md5:
:+1:

jwilder:
@viranch Haven't had a chance to test this yet but want to get it merged. @md5 Thanks for reviewing.  If anyone else could help test this PR out with their existing setup that would be very useful.

viranch:
If it helps, I'm using this patch in production, after thorough testing of course :)

jwilder:
Thanks @viranch!

PR #202, opened 2015-07-24
set default_server also for https

webner:
DEFAULT_HOST was not respected when using https

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -72,7 +72,8 @@ upstream {{ $host }} {
 {{ end }}
 }
 
-{{ $default_server := and ($.Env.DEFAULT_HOST) (eq $.Env.DEFAULT_HOST $host) }}
+{{ $default_host := or ($.Env.DEFAULT_HOST) "" }}
+{{ $default_server := index (dict $host "" $default_host "default_server") $host }}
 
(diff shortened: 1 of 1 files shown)


md5:
:+1:

jwilder:
This fails for me when running just one container with a `VIRTUAL_HOST` but not starting `nginx-proxy` with `-e DEFAULT_HOST=foo.com`.

```
dockergen.1 | 2015/07/25 22:50:01 template error: template: nginx.tmpl:75:48: executing "nginx.tmpl" at <eq $.Env.DEFAULT_HOS...>: error calling eq: invalid type for comparison
```

nginx.tmpl
```diff
@@ -72,6 +72,9 @@ upstream {{ $host }} {
 {{ end }}
 }
 
+{{ $default_host := or ($.Env.DEFAULT_HOST) "" }}
+{{ $default_server := index (dict $host "" $default_host "default_server") $host }}
```
webner: this is little bit hacky, but I didn't find another way to set a variable depending on another variable value.

Perhaps we could add a "when" function to docker-gen?
$default_server := when (eq $host $default_host) "default_server" ""

Informatic:
Are there plans on merging that in near future?

jwilder:
Thanks @webner !

PR #206, opened 2015-07-28
Fix boot2docker instructions

md5:
The volume mount for `/var/run/docker.sock` was missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,10 @@ Provided your DNS is setup to forward foo.bar.com to the a host running nginx-pr
 
 If your are using `boot2docker` start `nginx-proxy` with:
 
-    $ $(boot2docker shellinit)
-    $ docker run -p 80:80 -e DOCKER_HOST -e DOCKER_CERT_PATH -e DOCKER_TLS_VERIFY -v $DOCKER_CERT_PATH:$DOCKER_CERT_PATH -it jwilder/nginx-proxy
+    $ eval "$(boot2docker shellinit)"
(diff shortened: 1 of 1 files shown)


md5:
Acutally, I just noticed that `DOCKER_HOST` is being passed... That's not the way I run under b2d, but I suppose it works.

md5:
In fact, I don't do anything special to run under b2d.

@jwilder Why did you introduce the previous commit? Using TCP to connect in the b2d case is unnecessary. The unix socket connection works fine.

jwilder:
@md5 Didn't even realize that b2d was also using the unix socket.

md5:
Under `boot2docker`, all `--volume` mounts except those from the use home directory (`/Users` under OS X and `C:\Users` on Windows) will mount files from the VM's filesystem. This includes `/var/run/docker.sock`.

md5:
Even in the case of the user home directories, it's a VM directory that's being mounted. It just happens to be a VM directory that exposes a host directory via the VirtualBox shared folders extension.

thaJeztah:
Also note that boot2docker (the CLI tools, not the Linux distro), is on its way out and will be replaced by docker machine.

md5:
I'm closing this in favor of #224.

codepushr:
@md5 I'm struggling with using docker-machine and nginx-proxy. Right now I'm using port forwarding to port from inside the virtual machine :80 (nginx) to my host :8080 to be able to use localhost:8080. Then I used pf/pfctl to forward from 8080 on my host to 80 on my host to be able to use only localhost. The problem is that I can't access the virtual hosts that I configured (like foo.localhost). Any idea?

PR #209, opened 2015-07-29
removed SSL V1

uptownhr:
considered to be unsafe.


jwilder:
TLSv1 is still part of the intermediate compatibility configuration that the other SSL settings are based off of.  See: https://wiki.mozilla.org/Security/Server_Side_TLS

PR #215, opened 2015-08-14
customizability improvements

taxilian:
I'm still testing this, but it's mainly template fixes.  I thought I'd go ahead and make the pull request so you can look it over; if you're comfortable merging before I finish my tests then go for it, otherwise I'll let you know when I'm sure it's all working.

I also would like feedback if there is anything here you think should be done a different way.  All of these customizations are things I've found need for while using it

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,3 +193,8 @@ If you are using multiple hostnames for a single container (e.g. `VIRTUAL_HOST=e
 
     $ { echo 'server_tokens off;'; echo 'client_max_body_size 100m;'; } > /path/to/vhost.d/www.example.com
     $ ln -s /path/to/vhost.d/www.example.com /path/to/vhost.d/example.com
+
+#### Per-VIRTUAL_HOST default configuration
+
+If you want most of your virtual hosts to use a default single configuration and then override on a few specific ones, add those settings to the `/etc/nginx/vhost.d/default` file. This file
+will be used on any virtual host which does not have a `/etc/nginx/vhost.d/{VIRTUAL_HOST}` file associated with it.
--- a/README.md
+++ b/README.md
@@ -198,3 +198,24 @@ If you are using multiple hostnames for a single container (e.g. `VIRTUAL_HOST=e
 
 If you want most of your virtual hosts to use a default single configuration and then override on a few specific ones, add those settings to the `/etc/nginx/vhost.d/default` file. This file
 will be used on any virtual host which does not have a `/etc/nginx/vhost.d/{VIRTUAL_HOST}` file associated with it.
+
+#### Per-VIRTUAL_HOST location configuration
+
+To add settings to the "location" block on a per-`VIRTUAL_HOST` basis, add your configuration file under `/etc/nginx/vhost.d`
+just like the previous section except with the suffix `_location`.
+
+For example, if you have a virtual host named `app.example.com` and you have configured a proxy_cache `my-cache` in another custom file, you could tell it to use a proxy cache as follows:
+
+    $ docker run -d -p 80:80 -p 443:443 -v /path/to/vhost.d:/etc/nginx/vhost.d:ro -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy
+    $ { echo 'proxy_cache my-cache;'; echo 'proxy_cache_valid  200 302  60m;'; echo 'proxy_cache_valid  404 1m;' } > /path/to/vhost.d/app.example.com_location
+
+If you are using multiple hostnames for a single container (e.g. `VIRTUAL_HOST=example.com,www.example.com`), the virtual host configuration file must exist for each hostname. If you would like to use the same configuration for multiple virtual host names, you can use a symlink:
+
+    $ { echo 'proxy_cache my-cache;'; echo 'proxy_cache_valid  200 302  60m;'; echo 'proxy_cache_valid  404 1m;' } > /path/to/vhost.d/app.example.com_location
+    $ ln -s /path/to/vhost.d/www.example.com /path/to/vhost.d/example.com
+
+#### Per-VIRTUAL_HOST location default configuration
+
+If you want most of your virtual hosts to use a default single `location` block configuration and then override on a few specific ones, add those settings to the `/etc/nginx/vhost.d/default_location` file. This file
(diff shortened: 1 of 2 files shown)


nginx.tmpl
```diff
@@ -121,6 +125,8 @@ server {
 
 	{{ if (exists (printf "/etc/nginx/vhost.d/%s" $host)) }}
 	include {{ printf "/etc/nginx/vhost.d/%s" $host }};
+	{{ else if (exists /etc/vhost.d/defaults) }}
```
jwilder: Should be `default` to be consistent with other existing naming conventions.

nginx.tmpl
```diff
@@ -129,6 +135,11 @@ server {
 		auth_basic	"Restricted {{ $host }}";
 		auth_basic_user_file	{{ (printf "/etc/nginx/htpasswd/%s" $host) }};
 		{{ end }}
+		{{ if (exists (printf "/etc/nginx/vhost.d/%s_location" $host)) }}
+		include {{ printf "/etc/nginx/vhost.d/%s_location"}}
+		{{ else if (exists /etc/vhost.d/defaults_location) }}
```
jwilder: Same here `default_location` for consistency.

jwilder:
@taxilian A few small things but looks pretty good to me.

taxilian:
Your comment about positioning so that previous ones aren't broken came without a reference to the code you're referring to; I assume you mean the proxy_defaults.conf file.  That's in an else block, so you only provide that file if you want to replace the default options entirely. I considered adding a similar piece for ssl, but I don't personally actually need it so decided not to for now, given that you're using very sensible defaults.

taxilian:
The specific purpose of proxy_settings.conf is to replace the default proxy_settings line; calling it proxy.conf might be okay, but global.conf would lose the implication that it replaces the existing settings.  note that it's in an if / else block so if the file exists it will use that instead of doing the default options.

Perhaps proxy_config.conf?  It's not just intended to provide global context where you can add things, it's intended to allow you to change if desired the global proxy configuration.  I believe you are correct that you can't just redefine previous settings, you need to replace them.

README.md
```diff
@@ -143,6 +143,23 @@ You'll need apache2-utils on the machine where you plan to create the htpasswd f
 
 If you need to configure Nginx beyond what is possible using environment variables, you can provide custom configuration files on either a proxy-wide or per-`VIRTUAL_HOST` basis.
 
+#### Overriding default proxy settings
+
+If you want to override the default proxy settings for the nginx container, add a configuration file at `/etc/nginx/proxy_settings.conf`. A file with the default settings would
+look like this:
+
+```/etc/nginx/proxy_settings.conf
```
thaJeztah: I think there should be a new line before `/etc/...`

jwilder:
@taxilian Ah..  See: https://github.com/gradecam/nginx-proxy/commit/51c44fe7f8951ff9ebb8cd80cfbb1eb73aeea337

I see it's in an else block but my concern is that it's not overriding existing behavior, it's replacing it.  (e.g. if you do not add `proxy_set_header Host $http_host;`, etc.., the proxy will not work.

README.md
thaJeztah: Oh, or just removed, looking at it :-)

taxilian: Honestly I just put that there to mimic the block a little later in the file, assuming that it was done that way for a reason.  Sounds like they are both broken =]

thaJeztah: Well, on github, you can add a "language hint", e.g. bash, not sure what should be put here, think there's a list somewhere, I'll have a look

taxilian:
Ahh, I misread your comment.  I definitely see the concern, but it is an advanced customization action; it should be possible to remove unwanted options if that's what the user needs. Also, I think you're right that you can't override previous settings.  I'll add a warning in the docs about that, would that suffice?

README.md
thaJeztah: Complete list can be found here; https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

taxilian: Got it; that's new to me, thanks! I'll fix it.

README.md
```diff
@@ -143,6 +143,25 @@ You'll need apache2-utils on the machine where you plan to create the htpasswd f
 
 If you need to configure Nginx beyond what is possible using environment variables, you can provide custom configuration files on either a proxy-wide or per-`VIRTUAL_HOST` basis.
 
+#### Overriding default proxy settings
+
+If you want to override the default proxy settings for the nginx container, add a configuration file at `/etc/nginx/proxy_config.conf`. A file with the default settings would
```
jwilder: `_config` is still redundant and superfluous.  I'd rather use `proxy.conf` since it's shorter and matches up with the nginx examples: http://wiki.nginx.org/FullExample

taxilian:
Your project, so you get to choose the naming conventions =] I've changed it to proxy.conf

I think this pull request has now been updated to reflect all of your concerns; anything I missed?

README.md
```diff
@@ -143,6 +143,25 @@ You'll need apache2-utils on the machine where you plan to create the htpasswd f
 
 If you need to configure Nginx beyond what is possible using environment variables, you can provide custom configuration files on either a proxy-wide or per-`VIRTUAL_HOST` basis.
 
+#### Overriding default proxy settings
```
jwilder: I think this should say "Replacing default proxy settings".

README.md
```diff
@@ -143,6 +143,25 @@ You'll need apache2-utils on the machine where you plan to create the htpasswd f
...
+
+If you want to override the default proxy settings for the nginx container, add a configuration file at `/etc/nginx/proxy.conf`. A file with the default settings would
```
jwilder: s/override/replace/

jwilder:
@taxilian Sorry for the nitpicking on naming but it's hard to change later.  Thanks for the PR though!  This can address a number of outstanding issues.

I need to test it out but looks good now.

taxilian:
I understand completely. You have to be picky with a project to keep it clean. I'll push back on anything that I think will keep it from fixing my use cases, but yours are just stylistic and preference things, and for a small project it's important to be consistent.

thaJeztah:
Thanks @taxilian :+1:

jwilder:
Thanks @taxilian!

PR #216, opened 2015-08-15
Remove duplicate access log

wader:
For each proxied request nginx-proxy currently logs:
```
nginx_1 | nginx.1    | 192.168.59.3 - - [15/Aug/2015:10:03:48 +0000] "GET / HTTP/1.1" 502 172 "-" "curl/7.37.1" "-"
nginx_1 | nginx.1    | test.tld 192.168.59.3 - - [15/Aug/2015:10:03:48 +0000] "GET / HTTP/1.1" 502 172 "-" "curl/7.37.1"
```

This PR removes the log line without hostname. Is there a good reason for keeping that access log? non-proxied requests somehow?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,8 @@ RUN apt-get update \
 
 # Configure Nginx and apply fix for very long server names
 RUN echo "daemon off;" >> /etc/nginx/nginx.conf \
- && sed -i 's/^http {/&\n    server_names_hash_bucket_size 128;/g' /etc/nginx/nginx.conf
+ && sed -i 's/^http {/&\n    server_names_hash_bucket_size 128;/g' /etc/nginx/nginx.conf \
+ && sed -i 's/access_log/#access_log/g' /etc/nginx/nginx.conf
 
 # Install Forego
 RUN wget -P /usr/local/bin https://godist.herokuapp.com/projects/ddollar/forego/releases/current/linux-amd64/forego \


md5:
Good catch. I'd probably change the `sed` line to just drop the `access_log` line in the main conf file instead of commenting it out. You also may want to drop the `error_log` line at the same time and combine the two sed commands into one.

md5:
i.e. `sed -i 's/^http {/&\n    server_names_hash_bucket_size 128;/g; /access_log/d; /error_log/d' /etc/nginx/nginx.conf`

wader:
Ok can do. But could there be a good reason for keeping the error log (but redirect to fd 2?) e.g. not losing important error messages or will there always be a default server with a error_log?

md5:
It would be a breaking change in the parent `nginx` image that I don't think they're going to make: https://github.com/nginxinc/docker-nginx/blob/d3d1bcf9227e28e6ccc421090462faa98d5c62b4/Dockerfile#L16

wader:
Aha the error_log in the "global" scope already ends up in stderr. Hmm not sure i follow about the breaking change.

md5:
What I mean is that if the `nginx` image were to remove the mapping of the default error log file to `STDERR`, it would be a breaking change for their users. Hence, I think it's very unlikely that they would change this.

wader:
Ok but do we really want to remove the global error_log altogether?

md5:
Looking closer at the upstream `nginx.conf`, it looks like the `access_log` directive is inside the `http` block, but the `error_log` is outside of it.

Given that, I think what makes the most sense is to remove the `access_log` from the stock `nginx.conf` in this image since it's added in `nginx.tmpl`/`default.conf`, but to leave `error_log` in the main `nginx.conf` and remove it from `nginx.tmpl`.

I'm also starting to think that opening an issue against [`nginxinc/docker-nginx`](https://github.com/nginxinc/docker-nginx) asking for some advice on the right way to handle this. There may be a way to avoid the duplicate access log entries without having to modify the main `nginx.conf`, though my reading of the docs didn't reveal any. The only reason we added `access_log` in the first place was to make sure the vhost name shows up in the logs. If there were a way to change the `main` log format that's declared in the upstream `nginx.conf` instead of adding a new `access_log` in `nginx.tmpl`, that would be cool.

Also, it's possible that nginx is getting confused because one `access_log` directive refers to `/var/log/nginx/access.log` while the other one refers to `/proc/self/fd/1`.

Sorry if this reply is too stream-of-consciousness :stuck_out_tongue:

md5:
One more thing to consider is that other images that derive from `nginx` may run into this same problem if they want to customize their access logs and still have them go to `STDOUT`, so that may be another reason to open an upstream issue.

jgallen23:
I'm running into a similar issue because I want my logs in json format. I couldn't figure out a way to override the output without updating the nginx.conf. would be great to post an issue on the other repo to find out the suggested way to handle

md5:
See https://github.com/nginxinc/docker-nginx/issues/39

wader:
Nice, maybe chill a bit with this then and see what the base image people say

PR #224, opened 2015-08-25
Remove boot2docker-specific instructions

md5:
Both boot2docker and docker-machine will work fine with the normal `docker run`
command used in the rest of the docs.


jwilder:
Thanks!

Issue #188, opened 2015-06-26
Missing default server declaration for SSL returning 503


montgomery1944:
In our configuration nginx-proxy adds SSL to non secured multiple containers. Our domain is configured, so that it allows subdomains. The problem is, that entering non existing url, which means not mapped to any container, is returning the server from the first SSL declaration, which is the normal nginx behaviour. This causes many problems, when users makes mistakes in the URL and nginx-proxy is returning normal, invalid page instead of 503. This problem does not exist for HTTP URLs, because nginx-proxy contains default declaration returning 503 for them.

Currently we have to append this declaration to the config generated by nginx-proxy:
```
server {
        listen 443 ssl default_server;
        server_name _; # This is just an invalid value which will never trigger on a real hostname.
        return 503;
        ssl_certificate /etc/nginx/certs/default.crt;
        ssl_certificate_key /etc/nginx/certs/default.key;
}
```
which is very similar to the default config provided for HTTP servers.

It would be good if you could fix this issue internally in the nginx-proxy.

jwilder:
That config is already in the default template: https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl#L153-L162

Are you configuring SSL as described here: https://github.com/jwilder/nginx-proxy#how-ssl-support-works

Is this not working?

montgomery1944:
The config you mentioned is going to prevent from accessing https://existing.com when there is only http://existing.com mapping, but it's not going to prevent from accessing https://non-existing.com. Currently in the last case nginx-proxy is going to return the first SSL mapping from the config.

The config I provided in my issue was based on: https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl#L34-L38, which is preventing from accessing http://non-existing.com.

montgomery1944:
@jwilder, is my explanation clear? I can provide you with further details if you need.

funkygibbing:
Thankyou @montgomery1944 for this ticket and your workaround. I found your example to be clear and accurate, and the solution appears to work perfectly.

@jwilder I believe you simply need to replace `{{ $host }};` with `_;` in `nginx.tmpl` in order for this fix to work as required.

montgomery1944:
It seems that this issue was fixed in master branch.

PR #225, opened 2015-08-26
Fix for #188 - remove hostname from default SSL block

funkygibbing:
This resolves the issue where accessing a non-existent hostname over HTTPS returns the first SSL-enabled virtual host instead of correctly responding with a 503 error.


jwilder:
Thanks @funkygibbing !

PR #227, opened 2015-08-27
allow FORCE_SSL var to set SSL preference

briceburg:
we needed to allow :80 and :443 at the client discretion, so introduced FORCE_SSL env var to allow turning off the SSL-only preference. 

setting FORCE_SSL to any value but "yes" will disable the current default behavior and allow :80


sdurrheimer:
:+1: I need this too.

Skrath:
ditto

PR #228, opened 2015-09-01
Bumping to latest Docker Gen version

berfarah:
As per [this](https://github.com/jwilder/docker-gen/issues/114) ticket on docker-gen. Thanks again!


jwilder:
Thanks @berfarah!

PR #232, opened 2015-09-10
Added a higher client_max_body_size than standard

TheKnarf:
This allows you to among other things upload images and files which are bigger than 1mb if you use nginx infront of wordpress


md5:
You can add this yourself by either creating a derived image or by mounting in an additional config file with `-v`. Changing this setting for all users is not necessary.

See the docs here: https://github.com/jwilder/nginx-proxy#proxy-wide

PR #243, opened 2015-09-25
Allow vhosts to serve both http and https.

darrint:
I needed this for testing a site that handled both http and https.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,8 +112,9 @@ session caches.
 
 The behavior for the proxy when port 80 and 443 are exposed is as follows:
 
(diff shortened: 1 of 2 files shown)


sdurrheimer:
+1 for this kind of feature

patakijv:
+1 : this seems like it is a must have unless there is another way to avoid having 100% of your traffic be routed through HTTPS.

genezys:
PR #298 seems to have a more complete solution.

sdurrheimer:
#298 has been merged, this PR can be closed.

PR #244, opened 2015-09-29
Fix vhost_default path

rschmukler:


Cryszon:
This problem also seems to appear on lines 143 and 167 and makes defaults completely unusable.

wader:
the exists checks for default server blocks config looks broken too.

https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl#L141
https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl#L167

wader:
doh @Cryszon just said that

jwilder:
Fixed via #252

PR #245, opened 2015-09-29
feat(ssl): Allow using .pem for public key file

randombk:
Try using $cert.pem as the certificate before falling back to
  $cert.crt, as .pem is the standard file extension for
  certificates. (RFC1421, RFC1424). This is different from the
  current behaviour, where only the .crt is used.

  NOTE: This could cause a regression if users have both a .pem and
  a .crt file in the /etc/nginx/certs directory. In that
  situation, the .pem will be used instead of the .crt, which
  can be problematic if the files contain different information
  (i.e. one contains the entire certificate chain, while the other
  does not). If this is an issue, the order can be changed so that
  the current behaviour is default.

Signed-off-by: David Li jiawei.davidli@gmail.com

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -58,13 +58,18 @@ server {
 	return 503;
 }
 
-{{ if (and (exists "/etc/nginx/certs/default.crt") (exists "/etc/nginx/certs/default.key")) }}
+{{ if (and (or (exists "/etc/nginx/certs/default.pem") (exists "/etc/nginx/certs/default.crt")) (exists "/etc/nginx/certs/default.key")) }}
 server {
 	server_name _; # This is just an invalid value which will never trigger on a real hostname.
 	listen 443 ssl spdy;
 	return 503;
 
-	ssl_certificate /etc/nginx/certs/default.crt;
+    {{ if (exists "/etc/nginx/certs/default.pem") }}
+    ssl_certificate /etc/nginx/certs/default.pem;
+    {{ else }}
+    ssl_certificate /etc/nginx/certs/default.crt;
+    {{ end }}
+
 	ssl_certificate_key /etc/nginx/certs/default.key;
 }
 {{ end }}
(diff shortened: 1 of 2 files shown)


jwilder:
Would you be able to point me to the reference that says that `.pem` is the standard extension?

`.crt` seems pretty common.  https://en.wikipedia.org/wiki/X.509#Extensions_informing_a_specific_usage_of_a_certificate

randombk:
Hmm. Looking closer, I see that the specification doesn't actually dictate a particular file extension to be used. However, I still believe we should default to `.pem`, for the following reasons:

* `.pem` files are unambiguously in PEM format, which is what Nginx requires. `.crt` files may be in either PEM or DER format.
* `.pem` is what the Nginx documentation uses in its examples [1]
* There is quite a lot of literature out there that tends to claim `.pem` as the default [2] [3] [4]

Thanks.

[1] http://nginx.org/en/docs/http/ngx_http_ssl_module.html#example

[2] https://www.digicert.com/ssl-certificate-installation-nginx.htm#edit_nginx

[3] https://support.ssl.com/Knowledgebase/Article/View/19/0/der-vs-crt-vs-cer-vs-pem-certificates-and-how-to-convert-them

[4] http://serverfault.com/a/9717

md5:
I'm not sure how much @jwilder cares, but it looks like this PR does not maintain the existing whitespace. I know the files is currently a mix of hard tabs and spaces, but the lines being changed originally had tabs. I personally think the whole template should use tabs.

nginx.tmpl
```diff
@@ -97,16 +102,17 @@ upstream {{ $host }} {
 {{ $certName := (first (groupByKeys $containers "Env.CERT_NAME")) }}
 
 {{/* Get the best matching cert  by name for the vhost. */}}
-{{ $vhostCert := (closest (dir "/etc/nginx/certs") (printf "%s.crt" $host))}}
+{{ $vhostCert := (closest (dir "/etc/nginx/certs") (printf "%s.key" $host))}}
```
md5: This line seems weird. I think I understand why you changed it to look for `.key`, but it may need a comment.

randombk: This entire vhost matching section is a bit off, tbh. I'll write up a separate issue/PR once I get the chance.

The three `{{ $vhostCert := replace $vhostCert ".[....] " "" -1 }}` lines below this look like they would break if the vhost name ends in .pem, .crt, or .key (for whatever reason). I.e. if the vhost name is `foo.key`, and the certificate file name is `foo.key.crt`, then the name could get stripped down to `foo`. I'll need to do a bit of testing to see if this is actually an issue, but I think the overall section needs a rework to cover these edge cases.

md5: Luckily, `.crt`, `.key`, and `.pem` are not actual TLDs :) (see https://www.icann.org/resources/pages/listing-2012-02-25-en)

randombk: Haha, true. Unfortunately, I know more than one network that uses those internally... It's definitely a rare edge case, but an edge case all the same.

deini:
Any updates on this? I would also like to use `.pem`s

Issue #197, opened 2015-07-16
Add unit tests


jwilder:
The functionality in this image is complicated enough that some unit tests would be useful to test PRs against.

Something simple where each test could start an `nginx-proxy` container and sub-containers in various configuration and issue curl commands to return to verify result would be nice.

chaos95:
The unit and integration test suite over at [docker-compose](https://github.com/docker/compose) is pretty solid, and built around standing up test containers in various configurations. Might be worth looking to them for inspiration?

dmp1ce:
Should this be closed? Tests are running. I did notice that the tests do not clean up containers they create. I'll create another issue for that.

buchdag:
Closed by #197 and #655

PR #246, opened 2015-09-30
add test suite. See #197

thomasleveil:
This test suite is implemented using [bats](https://github.com/sstephenson/bats).

Not all features are tested. For instance ssl features and custom nginx config are missing. Probably others. But I hope this is enough to get contributors involved in writing tests.

This test suite won't work with TravisCI. Too many evenings were wasted trying to overcome [issues](http://stackoverflow.com/questions/32846800/travis-fails-to-stop-docker-containers) that only arise on the TravisCI platform. However, it runs on [CircleCI](https://circleci.com) which is also free for opensource projects.

An example of test result from CircleCI for this test suite is available at https://circleci.com/gh/thomasleveil/nginx-proxy/45

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,21 @@
+machine:
+  pre:
+    # install docker 1.7.1
+    - sudo curl -L -o /usr/bin/docker 'https://s3-external-1.amazonaws.com/circle-downloads/docker-1.7.1-circleci'; sudo chmod 0755 /usr/bin/docker; true
+  services:
+    - docker
+
+dependencies:
+  override:
+    - sudo add-apt-repository ppa:duggan/bats --yes
+    - sudo apt-get update -qq
+    - sudo apt-get install -qq bats
+    - docker pull jwilder/docker-gen
+    - docker pull nginx
+    - docker pull python:3
+    - docker pull rancher/socat-docker
+
+test:
+  override:
+    - docker build -t jwilder/nginx-proxy:bats .
+    - bats test
--- a/test/README.md
+++ b/test/README.md
@@ -1,0 +1,14 @@
+Test suite
+==========
+
+This test suite is implemented on top of the [Bats](https://github.com/sstephenson/bats/blob/master/README.md) test framework.
+
+It is intended to verify the correct behavior of the Docker image `jwilder/nginx-proxy:bats`.
+
+Running the test suite
+----------------------
+
+Make sure you have Bats installed, then run:
+
+    docker build -t jwilder/nginx-proxy:bats .
+    bats test/
--- a/test/default-host.bats
+++ b/test/default-host.bats
@@ -1,0 +1,38 @@
+#!/usr/bin/env bats
+load test_helpers
+
+function setup {
+	# make sure to stop any web container before each test so we don't
+	# have any unexpected contaiener running with VIRTUAL_HOST or VIRUTAL_PORT set
+	docker ps -q --filter "label=bats-type=web" | xargs -r docker stop >&2
+}
+
+
+@test "[$TEST_FILE] DEFAULT_HOST=web1.bats" {
+	SUT_CONTAINER=bats-nginx-proxy-${TEST_FILE}-1
+
+	# GIVEN a webserver with VIRTUAL_HOST set to web.bats
+	docker_clean bats-web
+	run docker run -d \
+		--label bats-type="web" \
+		--name bats-web \
+		-e VIRTUAL_HOST=web.bats \
+		--expose 80 \
+		-w /var/www \
+		python:3 \
+		python -m http.server 80
+	assert_success
+
+	# WHEN nginx-proxy runs with DEFAULT_HOST set to web.bats
+	run nginxproxy $SUT_CONTAINER -v /var/run/docker.sock:/tmp/docker.sock:ro -e DEFAULT_HOST=web.bats
+	assert_success
+	docker_wait_for_log $SUT_CONTAINER 3 "Watching docker events"
+
+	# THEN querying the proxy without Host header → 200
+	run curl_container $SUT_CONTAINER / --head
+	assert_output -l 0 $'HTTP/1.1 200 OK\r'
+
+	# THEN querying the proxy with any other Host header → 200
+	run curl_container $SUT_CONTAINER / --head --header "Host: something.I.just.made.up"
+	assert_output -l 0 $'HTTP/1.1 200 OK\r'
+}
--- a/test/default-host.bats
+++ b/test/default-host.bats
@@ -12,16 +12,7 @@ function setup {
 	SUT_CONTAINER=bats-nginx-proxy-${TEST_FILE}-1
 
 	# GIVEN a webserver with VIRTUAL_HOST set to web.bats
-	docker_clean bats-web
-	run docker run -d \
-		--label bats-type="web" \
-		--name bats-web \
-		-e VIRTUAL_HOST=web.bats \
-		--expose 80 \
-		-w /var/www \
-		python:3 \
-		python -m http.server 80
-	assert_success
+	prepare_web_container bats-web 80 -e VIRTUAL_HOST=web.bats
 
 	# WHEN nginx-proxy runs with DEFAULT_HOST set to web.bats
 	run nginxproxy $SUT_CONTAINER -v /var/run/docker.sock:/tmp/docker.sock:ro -e DEFAULT_HOST=web.bats
--- a/test/docker.bats
+++ b/test/docker.bats
@@ -62,7 +62,7 @@ load test_helpers
 		-v /etc/nginx/certs/ \
 		nginx:latest
 	assert_success
-	run retry 5 1s curl --silent --fail -A "before-docker-gen" --head http://$(docker_ip bats-nginx)/
+	run retry 5 1s curl --silent --fail --head http://$(docker_ip bats-nginx)/
 	assert_output -l 0 $'HTTP/1.1 200 OK\r'
 
 	# WHEN docker-gen runs on our docker host
--- a/test/docker.bats
+++ b/test/docker.bats
@@ -1,0 +1,117 @@
+#!/usr/bin/env bats
+load test_helpers
+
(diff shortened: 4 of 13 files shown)


jwilder:
Thanks @thomasleveil!  I'll give a try.  I don't have any issue using CircleCI for the project as well.

wader:
#244 is a good test case candidate

jwilder:
Finally got a chance to test this out.  This is great. I wasn't able to get it to work on OSX/docker-machine, but seems to run fine on linux.

jwilder:
Thanks @thomasleveil!

md5:
@jwilder How did you try running this on OSX/docker-machine? Perhaps I can take a look at those failures when I get a chance.

Off the top of my head, it looks like parts of the test are trying to mount in local directories into the test. When I've run across similar issues with the Official Images test suites, I've had to do things like `-v "$(pwd):$(pwd)" to get the tests to work.

jwilder:
@md5 I manually installed `bats` and also had to install `findutils` and tweak some tests to use `gxargs`, fix some command-line options, etc..  The first test seemed to hang and I wasn't able to see why easily so I just switched to running them under vagrant which worked.

md5:
I started working to fix these issues and made a bit of progress.

The `xargs -r` stuff I just replaced with a direct call to `docker stop` if there are any container ids found with `bats-type=web`.

The issue with the hanging was caused by the fact that the test was trying to connect to a container IP with `curl` from the host. I changed those to use my `appropriate/curl` image to run curl from a sibling container instead.

Now, I'm getting this really weird error on the "start 2 containers" test. It looks like this:

```console
$ bats test/docker.bats 
 ✗ [docker] start 2 web containers
   (from function `assert_success' in file test/lib/bats/batslib.bash, line 118,
    from function `prepare_web_container' in file test/test_helpers.bash, line 121,
    in test file test/docker.bats, line 7)
     `prepare_web_container bats-web2 82 -e VIRTUAL_HOST=web2.bats' failed
   container_name: bats-web1
   ports: 81
   options: -e VIRTUAL_HOST=web1.bats
   expose_option: --expose=81 
   container_name: bats-web2
   ports: 82
   options: -e
   VIRTUAL_HOST=web2.bats
   expose_option: --expose=82 
   
   -- command failed --
   status : 1
   output (2 lines):
     docker: Invalid range format for --expose: 82 , error: strconv.ParseUint: parsing "82 ": invalid syntax.
     See 'docker run --help'.
   --
```

I have no idea why the first call to `prepare_web_container` with port 81 works, but the second call thinks there's a trailing space on the `--expose=82` argument for some reason. When I switch the order of the "81" and "82" containers, it's still the second container that fails to start...

md5:
Here's the branch BTW: https://github.com/appropriate/nginx-proxy/tree/osx-fix-test-suite

thomasleveil:
what's also weird is that `-e VIRTUAL_HOST=web2.bats` seems to be on two lines instead of one.

for the space in `"80 "`, it comes from https://github.com/appropriate/nginx-proxy/blob/osx-fix-test-suite/test/test_helpers.bash#L91 and should not be an issue (on Linux). 
Maybe forcing the integer type for the port would help:

	for port in $ports; do
		local -i port_as_int=$port
		expose_option="${expose_option}--expose=$port_as_int "
	done

md5:
@thomasleveil I get where the extra space comes from, but what I don't understand is why it's causing a problem. Since `$expose_option` isn't double-quoted, the parameter expansion should make that extra space insignificant. For some reason, the first call to `prepare_web_container` in that test works as expected, but the second one has a problem with the extra space.

Good catch on the two lines for `-e VIRTUAL_HOST=web2.bats`. I'm not sure what that's about either... I checked `test/docker.bats` and it only has a space in its arguments to `prepare_web_container` for both the first and the second call.

md5:
Here's some more unexpected behavior I'm seeing in the `multiple-ports` tests:

```console
 ✗ [multiple-ports] nginx-proxy defaults to the service running on port 80
   (from function `assert_output' in file test/lib/bats/batslib.bash, line 355,
    from function `prepare_web_container' in file test/test_helpers.bash, line 127,
    in test file test/multiple-ports.bats, line 25)
     `prepare_web_container bats-web-${TEST_FILE}-1 "80 90" -e VIRTUAL_HOST=web.bats' failed
   container_name: bats-web-multiple-ports-1
   ports: 80 90
   options: -e VIRTUAL_HOST=web.bats
   expose_option: --expose=80 --expose=90 
   
   -- output differs --
   expected : answer from port 80 90
   actual   : answer from port 80
   --
   
 ✗ [multiple-ports] VIRTUAL_PORT=90 while port 80 is also exposed
   (from function `assert_output' in file test/lib/bats/batslib.bash, line 355,
    from function `prepare_web_container' in file test/test_helpers.bash, line 127,
    in test file test/multiple-ports.bats, line 34)
     `prepare_web_container bats-web-${TEST_FILE}-2 "80 90" -e VIRTUAL_HOST=web.bats -e VIRTUAL_PORT=90' failed
   88f58d0c9692
   container_name: bats-web-multiple-ports-2
   ports: 80 90
   options: -e VIRTUAL_HOST=web.bats -e VIRTUAL_PORT=90
   expose_option: --expose=80 --expose=90 
   
   -- output differs --
   expected : answer from port 80 90
   actual   : answer from port 80
   --
```

md5:
BTW, I'm using `bats` 0.4.0 installed with `brew install bats`. It looks like the Circle CI setup is using the same version installed from `ppa:duggan/bats`.

md5:
I managed to figure out that the `multiple-ports` failure is happening because `IFS` is getting set to `$' \t\n'` in `run` and not getting set back. I'm not sure why this isn't causing a failure under Linux though...

Looks like this was patched in an unreleased version of `bats`: https://github.com/sstephenson/bats/commit/1735a4fcd26bf34d5e3118185ff3b84f4bcd4258

However, there was also discussion about reverting the fix...

See https://github.com/sstephenson/bats/issues/89, https://github.com/sstephenson/bats/pull/90, and https://github.com/sstephenson/bats/pull/92

md5:
I just pushed a commit to my branch that makes 15/16 tests pass on OS X: https://github.com/appropriate/nginx-proxy/commit/b1a08843da6cf2983dec7cebb490aee2cf086273

The only remaining failure is this one:

```console
 ✗ [docker] -e DOCKER_HOST=tcp://...
   (from function `assert_success' in file test/lib/bats/batslib.bash, line 118,
    in test file test/docker.bats, line 41)
     `assert_success' failed
   
   -- command failed --
   status : 1
   output (9 lines):
     4b2a47217707471c0f9a18f49982c16a86bb4fa5dec43f425cca448fb1028a00
     Client:
      Version:      1.8.2
      API version:  1.20
      Go version:   go1.5.1
      Git commit:   0a8c2e3
      Built:        Fri Sep 11 01:46:35 UTC 2015
      OS/Arch:      darwin/amd64
     An error occurred trying to connect: Get https://172.17.2.65:2375/v1.20/version: dial tcp 172.17.2.65:2375: i/o timeout
   --
```

md5:
That last failure was for the same reason as the `curl` failure earlier, i.e. an attempt to access the container IP from the OS X host. I fixed that one in https://github.com/appropriate/nginx-proxy/commit/30e565f0ed62a2bf34ab62a356e432957f12ac10

I'm going to open a PR with these changes.

md5:
I figured out why this doesn't fail with the PPA-installed `bats` "0.4.0" under Ubuntu. The reason is that it's already patched...

```console
# grep -C5 oldIFS /usr/lib/bats/bats-exec-test

  source "${filename}"
}

run() {
  local e E T oldIFS
  [[ ! "$-" =~ e ]] || e=1
  [[ ! "$-" =~ E ]] || E=1
  [[ ! "$-" =~ T ]] || T=1
  set +e
  set +E
  set +T
  output="$("$@" 2>&1)"
  status="$?"
  oldIFS=$IFS
  IFS=$'\n' lines=($output)
  [ -z "$e" ] || set -e
  [ -z "$E" ] || set -E
  [ -z "$T" ] || set -T
  IFS=$oldIFS
}

setup() {
  true
}
root@9917664d4598:/# 
```

See http://bazaar.launchpad.net/~duggan/bats/trunk/revision/90?start_revid=90

thomasleveil:
I'm curious about your feedback regarding _bats_. 

here's mine:
- Shell scripting is supposed to be portable across platforms → it is not that easy in the end
- Shell scripting is tricky and prone to bugs (loops, variable expansion, strings, escaping, ...)
- Debugging bats tests is difficult (cannot debug step-by-step, cannot inspect variables easily, etc)
- Bats (when using helper functions) allows to write tests that can be easy to understand by newcomers (that's my opinion, what about yours?)

I have other test frameworks in mind I could test:
- [py.test](http://pytest.org/latest/) + [docker-py][1]
- [pea](https://github.com/gfxmonk/pea) + [docker-py][1]
- [lettuce](http://lettuce.it/) + [docker-py][1]

Let me know if you think it is worth the effort to explore new test frameworks

[1]: http://docker-py.readthedocs.org/en/latest/

md5:
@thomasleveil I agree with your points about Bats. One thing I would add is that it's good to test shell commands as opposed to something else since shell commands are what users will actually be using in the vast majority of cases.

I think it's probably good to stick with Bats, despite its shortcomings.

jwilder:
`bats` is some pretty sophisticated bash scripting.  Honestly, it seems well done but, I'm guessing there is a lot bash hacking that has to be done to get simple tests going.  I have not tried to write any tests yet but getting them to run consistently has been challenging (even under linux).  

Originally, I was thinking of some kind of test setup based on `docker-compose` where each dir would be a test with a compose file and any necessary test files (SSL certs, templates, etc..).  The tests would use `docker-compose` to spin up containers then make assertions, etc.. I was thinking of trying out https://github.com/docker/libcompose and just using `go test` potentially as well since it already includes a test framework and doesn't require a lot of dependencies.  Never got a chance to try this or flush out the idea but might be worthwhile at some point.

PR #257, opened 2015-10-11
Fix running test suite under Docker Machine

md5:
This PR makes the test suite run correctly under Docker Machine, including under OS X.

Fixes include:

* Avoiding the use of `xargs -r` (OS X fix)
* Testing container connectivity from sibling containers instead of directly
* Reset `IFS` to the default due to https://github.com/sstephenson/bats/issues/89


md5:
Looks like my `appropriate/curl` image locally differed from the one on Docker Hub... Lemme fix that.

md5:
I pushed a fix for `appropriate/curl` here: https://github.com/appropriate/docker-curl/commit/af2cc9f44f7a62666fa667d0c5664e26664fa5d7

I also pushed a new commit to this branch to remove a commented out line that should trigger a (hopefully successful) test run.

md5:
Now there seems to be some sort of error with Circle CI...

```console
 Error deleting container: Error response from daemon: Cannot destroy container a8bc77c4639069aebb72210eeadc18b6b50fa24c3f70670547cfc3031a92c38b: Driver btrfs failed to remove root filesystem a8bc77c4639069aebb72210eeadc18b6b50fa24c3f70670547cfc3031a92c38b: Failed to destroy btrfs snapshot: operation not permitted
```

md5:
BTW, reading the diff for this PR is much easier if whitespace is ignored since I changed the line endings on two of the files from DOS to Unix: https://github.com/jwilder/nginx-proxy/pull/257/files?w=1

jwilder:
I'd like to merge this but, I have not been able to get circleci to run the tests succesfully yet.  Seems to be complaining about `btrfs` snapshots which might be related to this: https://github.com/jwilder/nginx-proxy/pull/257/files#diff-29944324a3cbf9f4bd0162dfe3975d88R18.

md5:
I don't think that the failure is related to `docker pull docker`. I'm pretty sure it's coming from the `docker_clean` helper here: https://github.com/appropriate/nginx-proxy/blob/osx-fix-test-suite/test/lib/docker_helpers.bash#L7

I'm not sure why it suddenly started being reported as a problem, however.

md5:
Alright @jwilder 

Got it green with a couple of changes! :money_with_wings: :green_heart:

jwilder:
Nice @md5! Thanks for this PR!

md5:
Sure thing. It might be nice to still have the `--rm` when not running on CircleCI. I may have another PR for that at some point, since I don't like how many containers this leaves around when I test it locally.

I've also got another PR that I'm going to submit in the next few minutes that cuts the test time in half. Hopefully CircleCI doesn't choke on it...

thomasleveil:
I left the test containers hanging around on purpose to allow manual tests to investigate a failing test case.

My workflow is:
- start the containers for the failing test case
- do the curl queries to reproduce the issue and find a fix

we could provide a shell script to remove all containers named 'bats-*' to clean up after bats if no failing case were found

md5:
I was talking about the sibling containers used to run the Curl commands
now.

thomasleveil:
my mistake, then what about cleaning them in the bats `teardown` function?

For instance, they could be started with the docker label `bats-type=teardown` and the teardown function would be:

    function teardown {
        CIDS=( $(docker ps -q --filter "label=bats-type=teardown") )
        if [ ${#CIDS[@]} -gt 0 ]; then
                docker rm -vf ${CIDS[@]} >&2
        fi
    }

md5:
Perhaps. It's not really clear to me whether that would work under
CircleCI. It seems odd to me that “docker rm -vf" would work but not
“docker run --rm".

I was thinking of making the omission of “--rm" conditional on an
environment variable that's only set in circle.yml.

PR #260, opened 2015-10-13
Trap TERM to allow quicker exit of web containers

md5:
This cut the test time roughly in half for me locally.


md5:
This doesn't seem to have had quite as dramatic an effect on CircleCI as it had locally... Let me know if you just want me to close it.

jwilder:
I see the same speed ups locally as well so definitely a worthwhile change.

jwilder:
Thanks @md5!

md5:
Cool. :+1:

PR #262, opened 2015-10-13
Use trim to handle VIRTUAL_HOST values with spaces

mariusGundersen:
based on latest docker-gen


jwilder:
Thanks @mariusGundersen!

md5:
Even with this change, there will still be problems with spaces if any custom config files are used since the `$host` variable will still have extraneous space in those `exists` checks. I think it would be best do do something like `{{ $host := trim $host }}` (and the same for `$proto`).

jwilder:
Good point @md5

PR #265, opened 2015-10-14
makefile: add target to update docker images tests depend on

thomasleveil:


jwilder:
Thanks @thomasleveil!

PR #230, opened 2015-09-03
Remove duplicate access log entries

md5:
After getting some advice over at https://github.com/nginxinc/docker-nginx/issues/39#issuecomment-137470214, I managed to fix this issue.

Fixes #216

PR #272, opened 2015-10-18
Move SSL configuration items out of server block, into http block

27Bslash6:
This enables more control over HTTPS hosts behind the proxy, by adding different configuration options in the hosts.d folder for the cipher suite, protocols or timeouts.

Cosmetically, this also resolves the session cache not detected issue with Qualys SSL Labs test - see http://stackoverflow.com/questions/22732045/session-cache-not-detected-in-nginx


md5:
LGTM

I checked that `ssl_ciphers` is the same before and after.

@funkygibbing Just to be clear, is the issue that the `ssl_session_cache` is not being used for these two blocks?

1. https://github.com/jwilder/nginx-proxy/blob/1e0b9301747a48d7a51d5d088ac5df16bde878c0/nginx.tmpl#L62-L70
2. https://github.com/jwilder/nginx-proxy/blob/1e0b9301747a48d7a51d5d088ac5df16bde878c0/nginx.tmpl#L187-L195

funkygibbing:
Yes, I believe that's the reason SSL Labs can't reuse session tickets, but I don't know if that's a great concern in the wild that's why I think it's more of a cosmetic change

The bigger issue resolved here is the ability to tweak SSL config options per host without getting fatal config errors like: 
`nginx: [emerg] "ssl_ciphers" directive is duplicate in /etc/nginx/vhost.d/example.com:2`

md5:
@funkygibbing Thanks for the explanation. There was some discussion of SSL-specific parameters for virtual hosts when #106 was merged, but I think this may be the first time a user has reported a specific problem.

funkygibbing:
Changes just pushed accidentally to the same branch, which are not SSL related, but now that they're there... 

These changes are of a similar nature, reducing duplication and increasing customisation - access_log directives out of each server{} block, and add
`access_log /dev/stdout vhost; error_log /dev/stderr;` to the http{} block.

This enables per-vhost logging configuration in `vhost.d`, such as entirely disabling logging for a vhost or logging to a different location, such as a mounted volume specific to the vhost.

Also, since the nginx image log directory just symlinks `/var/log/nginx/access.log` and `...error.log` to `/dev/stdout` and `/dev/stderr` respecitvely, this possibly shortcuts a filesystem lookup.  I'm not sure about any performance gains be eliminating a symlink, I imagine it would be trivial.

md5:
Can you please revert the `access_log` changes? Nginx will only allow them to be overridden if they're at different levels of the configuration. Your original diff looked fine.

md5:
See https://github.com/jwilder/nginx-proxy/pull/230 for the discussion, along with the linked issues from the `nginx` image.

md5:
Go ahead and open a separate issue for how to allow per-Vhost configuration of logging and we can discuss it there.

funkygibbing:
Done! Sheepish-face.

md5:
Thanks! I know that it's slightly counter-intuitive how the Nginx configuration works here, but sorting out that `vhost` stuff was a pain. Best not to mix it up with the SSL changes, which are pretty straightforward and certainly useful. :+1:

md5:
Also, @jwilder isn't usually too picky about the Git history, but I'd probably drop the two most recent commits from this branch:

```console
$ git checkout -b old-master
$ git checkout master
$ git reset --hard 6018e25afca6abac756195d2acf0833e5d27217f
$ git push -f origin master
```

In the future, I'd recommend using a branch other than `master` for proposing PRs to make this stuff easier/safer to manage.

md5:
P.S. the `git checkout -b old-master` is just to preserve your old `master`, though you might want to just kill those commits entirely.

If you want to preserve the `access_log` commits in a different branch for discussion, you could do something like this:

```console
$ git branch access-log-consolidation da36919
``

Still, I think those changes are a non-starter since the way that Nginx handles the `access_log` directive makes the seemingly duplicative entries in the current template actually necessary.

funkygibbing:
Absolutely taken on board. It was meant to be a quick throwaway fork without future changes, as I'm applying more radical changes in a different fork.  No need to keep changes, they're minor.  Hard reset pushed.

I will shortly link a new PR though, and continue the chat there.  I'm using the changes seemingly without problem and am interested in your take on why they aren't good.

md5:
:+1:

buchdag:
This is and old PR which is no longer relevant, closing.

PR #273, opened 2015-10-20
Enable access_log customisation per virtual host

27Bslash6:
Following on from #216 and #230 

Removing `access_log` from default nginx.conf enables per-virtual host logging configuration.

Is there any argument against modifying default nginx.conf in this way, considering it's already being done with the `server_names_hash_bucket_size`? If so, why is `server_names_hash_bucket_size` added in the Dockerfile instead of in `nginx.tmpl`?

Paging Dr @md5

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,9 +9,11 @@ RUN apt-get update \
  && apt-get clean \
  && rm -r /var/lib/apt/lists/*
 
-# Configure Nginx and apply fix for very long server names
+# Update default nginx.conf
+# - apply fix for very long server names
+# - delete access_log to allow for per-virtual host customisation in nginx.tmpl
 RUN echo "daemon off;" >> /etc/nginx/nginx.conf \
- && sed -i 's/^http {/&\n    server_names_hash_bucket_size 128;/g' /etc/nginx/nginx.conf
+ && sed -i 's/^http {/&\n    server_names_hash_bucket_size 128;/g; /access_log/d' /etc/nginx/nginx.conf
 
 # Install Forego
 RUN wget -P /usr/local/bin https://godist.herokuapp.com/projects/ddollar/forego/releases/current/linux-amd64/forego \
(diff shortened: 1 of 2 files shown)


md5:
These changes seem reasonable to me.

To be honest, it probably makes sense to just replace the entire `nginx.conf`. That's what the Nginx folks suggested. I'm not sure that there's much being gained by relying on the default configuration of the `nginx` image.

funkygibbing:
Agreed. I'd say there's much to be gained by writing an essentially blank nginx.conf with the interesting bits imported in the nginx.tmpl or user entered conf.d / vhost.d entries.

almereyda:
I've just stumbled into this while debugging the upload of a big file.

> # 413 Request Entity Too Large

But that's clearly not the point here.

One should probably mirror all mounting options into this README from the nginx base image, including the HTML directory for standard splash pages and the log directory for getting access to the logs

https://github.com/dockerfile/nginx#attach-persistentshared-directories

md5:
@almereyda There isn't a log directory since the logs go to `STDERR` and `STDOUT`. In the base `nginx` image, `/var/log/nginx/access.log` and `/var/log/nginx/error.log` are symlinks.

md5:
Also, there shouldn't be a standard splash page for this image when it's working properly.

subdavis:
@27Bslash6 Sorry to necro this, but since I think this repo is dead, I need to merge these changes into my own fork.

How does this enable per-host loggin?  I see that you've factored `access_log` out to the root, but this doesn't add support for any new environment variables in other containers? 

I would expect something like `ACCESS_LOG_FORMAT` and `ACCESS_LOG_PATH` get added?  Or is your intention simply to lay the groundwork for that later?

27Bslash6:
@subdavis Since access_log isn't defined in the server block any more, it enables you to define per-virtual-host access logging in any `/etc/nginx/vhost.d/*` files as per normal nginx server {} syntax

PR #254, opened 2015-10-09
Support proxying multiple paths to different backends

md5:
I'm finally opening a PR to review my branch for supporting multiple paths with `VIRTUAL_PATH`. I think that some of the duplicated code could probably be made a bit more DRY, but this code seems like a good starting point for a discussion to me.

A couple things are happening in this patch.
1. Detect if `VIRTUAL_PATH` is being used within the containers with a shared `VIRTUAL_HOST`
2. If so, take the SHA-1 hash of the path and use it as a suffix for both the `upstream` name and the `proxy_pass` statement

In the event that `VIRTUAL_PATH` is not set for `/` within the context of a particular `VIRTUAL_HOST`, a 404 is returned. This should possibly be a 503 instead, but I didn't want to hold up the discussion of this patch any longer (it's been over a year since I first created this branch).

Known issues that might need to be addressed before merging:
1. Further consolidate duplicate code into the shared "location" template
2. Figure out if 503 should be returned for a missing `/` instead of 404
3. Handle the case where some containers within a `VIRTUAL_HOST` have `VIRTUAL_PATH` set but others do not

Fixes #33

PR #283, opened 2015-11-06
Created a "Location" template to remove duplicate code.

Savageman:
I have no knowledge about this templating language, but I saw the "upstream" template reused at several places, so I think it should also work to share the part inside `location { ... }` which is responsible to allow per VirtualHost configuration.

I hope this is useful.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -15,6 +15,27 @@
 	{{ end }}
 {{ end }}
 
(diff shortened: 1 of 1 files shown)


md5:
:+1: 

I have a very similar change in the branch I used for the stalled "multiple paths" PR: #254

buchdag:
Done in #1901

Issue #113, opened 2015-03-04
connection refused when docker daemon is secured


ArtworkAD:
I have secured my docker daemon like suggested by docker: https://docs.docker.com/articles/https/ The nginx-proxy stopped working and outputs:
```
dockergen.1 | 2015/03/04 13:48:01 error listing containers: dial unix /tmp/docker.sock: connection refused
dockergen.1 | 2015/03/04 13:48:01 Unable to ping docker daemaon: dial unix /tmp/docker.sock: connection refused
````

How can I run nginx-proxy with a secured daemon?

**Comment by md5** [comment]
You'll need to start your Docker daemon with both a TLS socket and a Unix socket.

How are you starting your Docker daemon?

**Comment by ArtworkAD** [comment]
I run docker via systemd. In /etc/default/docker I have configured docker like this:
```
DOCKER_OPTS='--tlsverify --tlscacert=/etc/docker/ca.pem --tlscert=/etc/docker/server-cert.pem --tlskey=/etc/docker/server-key.pem -H=0.0.0.0:2376'
```

After that nginx proxy stopped working. @md5 any ideas?

jwilder:
When you switched to TLS, you removed the unix domain socket that nginx-proxy uses to connect to docker.  nginx-proxy does not currently support TLS although you should be able to get it to work by mounting in your docker key/cert and starting docker-gen w/ the -tls* flags.  If you don't do that, you'll have to add back the `-H unix:///var/run/docker.sock` to your `DOCKER_OPTS`

md5:
:100:

ArtworkAD:
@jwilder any plans on supporting tls? So I need to create separate containers, one for nginx and one for docker gen?

jwilder:
@ArtworkAD No immediate plan but I added #114.

md5:
@ArtworkAD What's wrong with doing as @jwilder suggested an having Docker listen on both TLS and a Unix socket?

md5:
FWIW, that's how Docker Machine configures the Docker daemons it provisions.

ArtworkAD:
How can I configure this?

md5:
DOCKER_OPTS='--tlsverify --tlscacert=/etc/docker/ca.pem --tlscert=/etc/docker/server-cert.pem --tlskey=/etc/docker/server-key.pem -H=tcp://0.0.0.0:2376 -H unix:///var/run/docker.sock'

ArtworkAD:
[...]

PR #286, opened 2015-11-08
add more descriptive error message when unix mount fails

dylanbfox:
Thought it might be helpful for people running into this issue to print this tip to users as well. I know it's been discussed in issues like https://github.com/jwilder/nginx-proxy/issues/113, but it took me a bit of digging to figure out this was actually the mistake I had made.


jwilder:
This doesn't seem accurate.  You can use TLS over TCP w/ docker.

Issue #295, opened 2015-11-17
Option to disable "SSL preference"


bluepuma77:
For some tests I needed to reach a subdirectory via SSL. Now suddenly every access is redirected to SSL, which really breaks some things.

I found in the documentation "If a container has a usable cert, port 80 will redirect to 443 for that container so that HTTPS is always preferred when available".

In general it's a good idea to prefer the secure connection, but in some cases it's not. So it would be great to have an option to disable the preference and just stick with the original connection type.

kamermans:
I second this request.  In my use case, I have several subdomains that are served, all of which are covered by a wildcard SSL cert, but only some of them should be served via SSL by default.  It would be nice to turn the auto-SSL redirect on or off via an environment variable.

tkw1536:
Resolved by solved by [HTTPS_METHOD=noredirect](https://github.com/nginx-proxy/nginx-proxy#how-ssl-support-works), closing.

Issue #291, opened 2015-11-12
301 Moved Permanently


jareware:
[According to the docs](https://github.com/jwilder/nginx-proxy#how-ssl-support-works):

> If a container has a usable cert, port 80 will redirect to 443 for that container so that HTTPS is always preferred when available

Preferring HTTPS over HTTP makes a lot of sense, but that redirect is implemented as 301 Moved Permanently. I just landed in a bit of hot water after temporarily enabling HTTPS support, and moving back to HTTP when things didn't work out, but several clients had already seen the permanent redirect and wouldn't even consult the HTTP endpoint anymore.

Given the dynamic nature of containers coming and going, would it be a safer option to redirect with 302 Moved Temporarily?

wader:
Also note that a Strict-Transport-Security header is added by nginx-proxy which could cause this

jareware:
Just did a factory reset for one embedded device that would otherwise never even try the HTTP endpoint again... :)

wader:
Sounds like HSTS

jareware:
Wouldn't a Moved Permanently have that same effect, though?

md5:
@jareware Browsers don't pay attention to 301 v. 302 as far as I know. The only thing I've ever heard of paying attention to that distinction is search engine crawlers. And even there, I'd be surprised if they actually pay attention since developers get it wrong too often.

jareware:
OK, in that case HSTS sounds like a lot more likely explanation... and I don't think anyone'd want to give that up?

I don't suppose there's any way to keep a specific virtual host on plain old HTTP, and serve other virtual hosts through HTTPS?

md5:
The redirect to HTTPS and the HSTS logic are only triggered if you provide a cert for a `VIRTUAL_HOST` by hostname or if you set `CERT_NAME` on the container. The only other way that HTTPS would be enabled is if you provided `default.crt`, in which case you wouldn't get the redirect to port 443 and if you did hit port 443 it would give a 503 error with no `Strict-Transport-Security` header.

So to keep a specific virtual host on plain HTTPS, you just don't provide a cert for that host. If you can explain your virtual host setup a little more fully, perhaps there's something I'm missing about your situation.

jareware:
[...]

Issue #239, opened 2015-09-21
301 redirect to HTTPS without a cert


danmandle:
Even with -e CERT_NAME="", nginx-proxy is still is 301 redirecting http requests to https.

tkw1536:
Closed by #298. Use [`HTTPS_METHOD=noredirect`](https://github.com/nginx-proxy/nginx-proxy#how-ssl-support-works).

PR #298, opened 2015-11-20
Added env var to disable SSL redirect

kamermans:
This PR adds an environment variable option to disable the redirection of HTTP traffic to HTTPS.

The following possibilities exist:
   - `HTTPS_METHOD=redirect` This is the default behavior - redirect HTTP to HTTPS if available.
   - `HTTPS_METHOD=noredirect`  Instead of the redirect, the normal HTTP server block is rendered (although HSTS is still enabled).
   - `HTTPS_METHOD=nohttp`   No HTTP block is generated at all, so only HTTPS is available.

This solves the following tickets, from what I can tell (and possibly more): #295, #291, #239


mathieutu:
This function is really interesting !
I've actually some issues with the automatic redirection : default host, and wildcard certificate on the container which is concerned, but several domains are pointing on the reverse, and many without certificates. When I'm trying with http with an unsecure domain, the proxy try to go on https, and there is no certificate.

Is-it in production now ?
Thank's

kamermans:
@mathieutu yes, this is up and running in production on about 12 of my sites right now.  Everything is working great for me.

martinmaillard:
A way to disable SSL for one virtual host is necessary. I deployed this forked version and it works very well. I hope this gets merged.

vrehak:
@jwilder Is there a reason not to merge this PR? I also had to deploy @kamermans fork and it's running OK but I'd rather stick to the main branch.

jedi4ever:
@jwilder @kamermans @vrehak :+1:

FlimFlam:
+1

elliotfehr:
+1 @jwilder

SergiuCip:
Hi @kamermans I've tried to use the template on my own branch of nginx-proxy, however, it is still redirecting me to the htts. Looks like it doesn't want to use http for some reason. 

I was wondering where do you pass the HTTPS_METHOD=noredirect when you run the nginx-proxy or  when you run the client?

Thanks

kamermans:
Hi @SergiuCip, you need to add `HTTPS_METHOD=noredirect` as an environment variable when starting the `nginx-proxy` container, either via `-e "HTTPS_METHOD=noredirect"` if you're using docker-run, or via the `environment` section in a `docker-compose.yml` file.

If you have done this already, chances are that your requests are not even getting to the `nginx-proxy`.  The default `nginx-proxy` redirects to HTTPS and enables something called HTTP Strict Transport Security (HSTS), which tells your browser not to even try HTTP for a certain time period.  You should test it using CURL, which will not remember the HSTS directive like your browser.

SergiuCip:
Hello @kamermans! Thank you for your detailed explanation :+1: . Yes you were right it was my browser caching the connection and always was going to HTTPS instead of HTTP, I've open an incognito / private tab and everything worked. Thanks a lot for implementing this functionality :+1:

kamermans:
@jwilder: note that I've merged your master branch back in and I added bats tests for SSL/HTTPS that cover the default operation as well as both of the new HTTPS_METHOD options in this PR.

jwilder:
Thanks @kamermans!

kamermans:
You're welcome @jwilder, thanks for the pull!

sakoht:
This doesn't seem to be respected for me, whether I put the HTTPS_METHOD into the proxy or the delegate server:

I do this to setup the proxy,
```
docker run --name vproxy -e HTTPS_METHOD=noredirect -d -p 80:80 -p 443:443 -v /mysslcerts:/etc/nginx/certs -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy
```

```
docker run \
  -it \
  -e VIRTUAL_HOST=myhost \
  -e HTTPS_METHOD=noredirect \
  -p 8080:80 -p 8443:443 \
  --expose 80 \
  --expose 443 \
  --name rem \
  $IMAGE \
  startserver.sh
```

These work, so I can hit ports 80 and 443 on the server:
```
curl http://myserver:8080
# got on port 80
curl https://myserver:8443
# got on port 443
```

And this proxies to port 80:
```
curl http://myserver
# got on port 80
```

But this is still redirection 443 -> 80:
```
curl https://myserver
# got on port 443
```

I've verified that if I remove the proxy and let the server map -p to 80 and 443 directly everything works.

Am I doing something wrong, or is it impossible to have the proxy just pass 80 and 443 through to this server and let the server worry about the SSL stuff?

kamermans:
First, the correct way to use it is on the virtual host, not the `nginx-proxy` container, so your second example was correct.

`HTTPS_METHOD=noredirect` overrides the default behavior of sending an `HTTP 301` redirect to clients on port `80` to the secure port, `443`.  What you are seeing is not a redirection, but a port-forwarding / port address translation.  This is [described in the README.md](https://github.com/jwilder/nginx-proxy#multiple-ports) as follows:

> If your container exposes multiple ports, nginx-proxy will default to the service running on port 80. If you need to specify a different port, you can set a VIRTUAL_PORT env var to select a different one. If your container only exposes one port and it has a VIRTUAL_HOST env var set, that port will be selected.

I did not write that code, but I imagine there is a good reason for it.  Basically, `nginx-proxy` will only proxy traffic back to a single port, which you can choose with `VIRTUAL_PORT`.  Note that I recently contributed the `X-Forwarded-Port` update as well, so in your application, you can check if the HTTP header `X-Forwarded-Port: 443` is set - this traffic was answered by `nginx-proxy` on `443` and you could handle it differently if you so choose.

sakoht:
@kamermans so if my container listens has `HTTPS_METHOD=noredirect`, and just listens on port 80, it should get both the http and https requests, with your header distinguishing them.  Is that correct?

It seems that with `noredirect` the SSL requests make it through to the container, and the http requests get a 503 from nginx-proxy.

My container (happens to run apache, which is nice so I can see which layer gives an error):
```

**Comment by sakoht** [comment]
@kamermans thanks.

So if my container has `HTTPS_METHOD=noredirect`, and just listens on port 80, it should get both the http and https requests, with your header distinguishing them.  Is that correct?

It seems that when I have a wildcard cert with subdomains, with `noredirect` the SSL requests make it through to the container, but the _non-SSL_ requests get a 503 from nginx-proxy on sub-domains:

In my case, I have a wildcard cert for (*.rememberthis.io), and:
- https://rememberthis.io         # works
- https://foo.rememberthis.io  # works
- https://bar.rememberthis.io  # works

- http://rememberthis.io          # works
- http://foo.rememberthis.io   # FAILS with a 503 from nginx-proxy
- http://bar.rememberthis.io   # FAILS with a 503 from nginx-proxy

My container happens to run apache, and now has a dumb config of static files just for testing:
```   
<VirtualHost *:80>
     ServerName rememberthis.io
     ServerAdmin admin@rememberthis.io
     ServerAlias rememberthis.io
   
     <Directory /opt/rem/doc-root>
       Options Indexes FollowSymLinks
       AllowOverride None
       Require all granted
     </Directory>
     DocumentRoot /opt/rem/doc-root
   </VirtualHost>
```

The docker-compose for the container:
```
  myserver:
    restart: always
    image: XXXXX (apache image)
    environment:
      - VIRTUAL_HOST=rememberthis.io
      - HTTPS_METHOD=noredirect
    entrypoint: XXX(start apache)
    networks:
      - nginx-proxy
```

Am I mis-configuring this?

kamermans:
> So if my container has HTTPS_METHOD=noredirect, and just listens on port 80, it should get both the http and https requests, with your header distinguishing them. Is that correct?

Yes, that is how it is supposed to work.

I am not experiencing the problem that you mentioned, although I use subdomains covered by wildcards in this way regularly.  I'll check out issue #737.

PR #300, opened 2015-11-21
Add Let's Encrypt support

dmp1ce:
Using the [simp_le](https://github.com/kuba/simp_le) Let's Encrypt client a bash script runs and continuously tries to create/renew certificates if the `LETSENCRYPT_HOST` variable is set for a container.

Certificates are automatically downloaded and placed in the correct location for nginx-proxy to use them.


dmp1ce:
I'm not expecting that this actually gets merged because simp_le adds a lot of bloat to the container, I think it is worth a look for anyone wanting to integrate Let's Encrypt into their nginx-proxy.

hadim:
Must have feature !

md5:
@dmp1ce What if instead of merging this into the default image/tag or maintaining it as a fork, it were to be made available as a derived image or add-on somehow?

dmp1ce:
@md5 I thought about that but I think I couldn't come up with a solution where I wasn't overriding the nginx.tmpl and Procfile. So, I would be managing changes from nginx-proxy either way.

hadim:
@md5 is there is a possibility for you to merge this feature  ?

Letsencrypt is just a specific service but I think in a near future ACME client will become widely used and implemented by a several certificate authorities (not just letsencrypt). I really think this feaure is a must have. More over it's opt-in meaning that usual user won't see anything different from before.

I'll respect if you are against merging it. In this case I think @dmp1ce you should think about forking it.

jwilder:
@hadim I won't be merging this as is since it adds a lot of weight to the image and I haven't had time to look at LE.

dmp1ce:
@hadim This project is already forked and the Docker image is available here: https://hub.docker.com/r/dmp1ce/nginx-proxy-letsencrypt/

JrCs:
@jwilder In my PR request (https://github.com/dmp1ce/nginx-proxy-letsencrypt/pull/2) i have reduce the image size a lot: from 469 MB to 224.1 MB.
Only files needed by simp_le are add to your base image.

hadim:
@dmp1ce very nice ! Could you please open Github issues in your fork ? I have an issue when `$domain` contains multiple host such as `"example.com,www.example.com"`. `simp_le` fails with this error `ERROR:simp_le: ACME server returned an error: unauthorized :: The client lacks sufficient authorization :: Error creating new authz :: Syntax error`.

Please enable issues in your fork, I don't want to create unnecessary noise here.

hadim:
For the record, the following diff against `letsencrypt_service` file should fix it : 

```diff
@@ -17,7 +17,9 @@
         hosts_array=$host_varname[@]
         email_varname="LETSENCRYPT_${cid}_EMAIL"
 
-        for domain in "${!hosts_array}"; do
+        IFS=', ' read -r -a hosts_splitted <<< "${!hosts_array}"
+
+        for domain in "${hosts_splitted[@]}"; do
 
             # Create the domain directory
             mkdir -p /etc/nginx/certs/$domain
```

thaJeztah:
I wonder if generating and downloading certificates is something that belongs in the proxy image. Yes, I love to see more widespread use of letsencrypt, just not sure it needs to be in the same image.

I haven't looked into the details, but after the certificate has been generated and downloaded, it's just a "regular" host, and nginx-proxy does not need to know where the certificate came from (correct me if I'm over-simplifying)

Would it be possible to create a separate project/image (e.g. "letsencrypt-generator"), that listens to docker events, checks for `LETSENCRYPT_xx` env vars, writes the certificates to a shared volume, and triggers the nginx-proxy container to reload the configuration?

JrCs:
@hadim as the documentation said:
  *Specify multiple hosts with a space delimiter.*
Do you try to separate the domains with a space instead of comma ?

hadim:
+1 @JrCs

dmp1ce:
@hadim I open issues.

@thaJeztah The problem I had verifying to Let's Encrypt that the specified domain had control of port 443. This needs to be demonstrated every three months. nginx-proxy has control over that port.

georgwaechter:
Having letsencrypt plus the nginx proxy working automatically would be awesome! This would save many boring hours of renewing, configuring and testing certificates.

I'm wondering whether the following would also be a good alternative solution (that does not install additional software within the proxy image)?

* prepare the nginx proxy configuration to redirect the special ./well-known/acme-challenge location to the letsencrypt client, which
* is running in its own docker container in standalone mode (https://letsencrypt.readthedocs.org/en/latest/using.html#standalone)
* mount the certificate folder which is ready by the nginx proxy also into the letsencrypt container
* when the letsencrypt container is finished, the proxy automatically reloads and "sees" the new certificates

What do you think about this idea? Could it work, or have I forgotten something?

I have not tested this myself until now! Addtionally i have not verified whether the standalone mode also uses the special ./well-known/acme-challenge folder which is used by the webroot method.

All the best

Georg

dmp1ce:
@georgwaechter I had trouble automating the standalone client. The reason I use the simp_le client is because it is easy to automate.

My fork is modifying the nginx proxy configuration and using the `./well-known/` location for all VIRTUAL_HOSTs.

I felt it was simpler to modify the nginx.tmpl directly and put the Let's Encrypt service in the same container. However, I think it is possible to add all of the configuration to default.conf or VIRTUAL_HOST configuration in `/etc/nginx/vhost.d` and forward the request to a Let's Encrypt only container. You would need to duplicate the code in `location /`. This way you could use the nginx-proxy image and not a fork.

janeczku:
@dmp1ce 
> I'm not expecting that this actually gets merged because simp_le adds a lot of bloat to the container

Please have a look at https://github.com/xenolf/lego. It's a LE client in a single Go binary without any dependencies. This should reduce the overhead to a mere 10 MB.

dmp1ce:
@janeczku 

I believe @JrCs has gotten simp_le install down to 20 MB so the size isn't that big of a deal. Are there any other reasons to use lego over simp_le? simp_le is working well for this project so far.

janeczku:
@dmp1ce 
> I believe @JrCs has gotten simp_le install down to 20 MB

All good then! :smile:

JrCs:
I have create a new dedicated container: [letsencrypt-nginx-proxy-companion](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion) that create/renew Let's Encrypt certificates automatically but using the official nginx-proxy.

thaJeztah:
Awesome work, @JrCs!

jacobtomlinson:
I'm using @JrCs's  [letsencrypt-nginx-proxy-companion](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion) and it does the job perfectly.

tmiklas:
Thanks @JrCs for your work - it does the job perfectly!

@dmp1ce - if possible, it would be useful to configure vhosts in a way that if there's only HTTP version of the site (no LETSENCRYPT_HOST variable set == no keys generated), there is a dummy/default error page or simple HTTP error code returned instead of some other random HTTPS served by the same proxy instance. 

Scenario below:
- Site 1 with LETSENCRYPT_HOST set - available on HTTP (redir to HTTPS) and HTTPS
- Site 2 w/o LETSENCRYPT_HOST set - available on HTTP, when accessed over HTTPS shows Site 1

dmp1ce:
@tmiklas, If you set DEFAULT_HOST to "Site 1" then it might work. https://github.com/jwilder/nginx-proxy#default-host

You can also add your own custom nginx configuration if you need. https://github.com/jwilder/nginx-proxy#custom-nginx-configuration.

In what situation would you not just get a certificate for all sites?

tmiklas:
That would work @dmp1ce... will test shortly. 
The situation where site doesn't have certificate would mostly be when user decides to not have it (dunno why, but hey...) and effectively mixing HTTP and HTTP/HTTPS sites behind single proxy. Matter of applying common sense and best practices.
Thanks for quick advice!

18601673727:
I have serval proxies fighting each other for port 80 since @dmp1ce 's fork has own `letsencrypt` feature and rckrdstrgrd's one doesn't but his fork has own `docker-compose V2` feature, so is there someone could possible to merge their forks and give my proxies some peace?

dmp1ce:
@18601673727 No one is supporting my fork any longer. I recommend using another nginx-proxy. I use  `jwilder/nginx-proxy` and `JrCs/docker-letsencrypt-nginx-proxy-companion` for Let's Encrypt.

18601673727:
@dmp1ce Yeah, @JrCs really did a nice job, problem solved.

giovannicandido:
Awesome. This feature is great. Nice work guys :-)

corradio:
Any progress here?

joshmanders:
@corradio There is https://github.com/dokku/dokku-letsencrypt for letsencrypt. I use it and it works flawlessly.

jwilder:
Going to close this since https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion can be used for Let's Encrypt SSL support.

PR #310, opened 2015-12-05
Add letsencrypt service into nginx-proxy

### LINKED ISSUE #274: can you add Let's Encrypt
Our project is very nice.
It is possible to add the Let's Encrypt Script to get free and automated SSL-certificates for each container?
Look here: https://github.com/letsencrypt/letsencrypt/tree/master/letsencrypt-nginx

That can be the whipped cream. ;)
P.S. Let's Encrypt should be availability for production in  week of November 16, 2015
https://letsencrypt.org/2015/08/07/updated-lets-encrypt-launch-schedule.html

JrCs:
Add transparent integration of letsencrypt service into nginx-proxy.
Only add 40MB in the image size.

Close #274, #300

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-![nginx 1.9.6](https://img.shields.io/badge/nginx-1.9.6-brightgreen.svg) ![License MIT](https://img.shields.io/badge/license-MIT-blue.svg) [![Build](https://circleci.com/gh/jwilder/nginx-proxy.svg?&style=shield&circle-token=2da3ee844076a47371bd45da81cf27409ca7306a)](https://circleci.com/gh/jwilder/nginx-proxy)
+![nginx 1.9.6](https://img.shields.io/badge/nginx-1.9.6-brightgreen.svg) ![License MIT](https://img.shields.io/badge/license-MIT-blue.svg) [![Build](https://circleci.com/gh/jwilder/nginx-proxy.svg?&style=shield&circle-token=2da3ee844076a47371bd45da81cf27409ca7306a)](https://circleci.com/gh/jwilder/nginx-proxy) [![](https://badge.imagelayers.io/jwilder/nginx-proxy:latest.svg)](https://imagelayers.io/?images=jwilder/nginx-proxy:latest 'Get your own badge on imagelayers.io')
 
 nginx-proxy sets up a container running nginx and [docker-gen][1].  docker-gen generates reverse proxy configs for nginx and reloads nginx when containers are started and stopped.
 
@@ -120,6 +120,28 @@ Note that in the latter case, a browser may get an connection error as no certif
 to establish a connection.  A self-signed or generic cert named `default.crt` and `default.key`
 will allow a client browser to make a SSL connection (likely w/ a warning) and subsequently receive
 a 503.
+
+#### Let's Encrypt
+
+Use the Let's Encrypt service to automatically create a valid certificate for a virtual host.
+
+Set the following environment variables to enable Let's Encrypt support for a container being proxied.
+
+- `LETSENCRYPT_HOST`
+- `LETSENCRYPT_EMAIL`
+
+The `LETSENCRYPT_HOST` variable most likely needs to be the same as the `VIRTUAL_HOST` variable and must be publicly reachable domains. Specify multiple hosts with a comma delimiter.
+
+For example
+
+```
+$ docker run -d -p 80:80 -p 443:443 \
+    -e VIRTUAL_HOST="foo.bar.com,bar.com" \
+    -e LETSENCRYPT_HOST="foo.bar.com,bar.com" \
+    -e LETSENCRYPT_EMAIL="foo@bar.com" \
+    -v /var/run/docker.sock:/tmp/docker.sock:ro \
+    jwilder/nginx-proxy
+```
 
 ### Basic Authentication Support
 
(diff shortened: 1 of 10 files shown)


wader:
Nice. Maybe cert files update needs to be atomic (per domain?) so that nginx reload because docker event don't happen to see corrupt or missing crt/key file for domains?

JrCs:
@wader it is the case. Reload of nginx and creation/renewal of certs are optimise to only be done if necessary.
You can try my automatic build docker image: jrcs/nginx-proxy-letsencrypt:nginx-proxy

PunKeel:
:+1: for this, but why are you using "LETSENCRYPT_HOST" instead of pulling the list of hosts from the generated configuration ? (That's the whole point of this system :/)

Also, we can't specify one emails per "host" ... Maybe might it use one variable EMAIL per "guest" container ?

JrCs:
@PunKeel LETSENCRYPT_HOST and LETSENCRYPT_EMAIL are used to select the SSL certificates to create.
VIRTUAL_HOST is used by ngnix-proxy to redirect some domains to containers

PunKeel:
This `VIRTUAL_HOST` is useless, right ? (Sorry, forgot to include the link…)

https://github.com/JrCs/docker-nginx-proxy-letsencrypt/commit/670d175ee948a33bb284f2059a3040f6e099867e#diff-04c6e90faac2675aa89e2176d2eec7d8R139

JrCs:
@PunKeel NO the VIRTUAL_HOST is necessary for nginx to redirect domains !
VIRTUAL_XXX variables are use to configure nginx proxy.
LETSENCRYPT_XXX variables are use by letsencrypt service

README.md
```diff
@@ -121,6 +121,28 @@ to establish a connection.  A self-signed or generic cert named `default.crt` an
...
+
+For example
+
+```
+$ docker run -d -p 80:80 -p 443:443 \
+    -e VIRTUAL_HOST="foo.bar.com,bar.com" \
+    -e LETSENCRYPT_HOST="foo.bar.com,bar.com" \
+    -e LETSENCRYPT_EMAIL="foo@bar.com" \
+    -v /var/run/docker.sock:/tmp/docker.sock:ro \
+    jwilder/nginx-proxy
```
ArthurHoaro: Here it should be `...` or `whatever/image`, otherwise it's misleading.

atomaka:
@jcrs Using `LETSENCRYPT_HOST` requires us to configure hosts to use SSL.  Let's Encrypts goal is to have SSL always on. Wouldn't it make more sense to match this goal and have SSL enabled be default for all hosts?  Then, you could just use all hosts registered and create SSL certificates for each.  If for some reason SSL cannot be enabled, perhaps this should be configured per guest with something like `LETS_ENCRYPT_ENABLED=false`.

Awesome work though, I set it up and initial certificates come through. I'll let you know in 90 days if renewals work :)

PunKeel:
:+1: @atomaka ...

JrCs:
@atomaka: LETSENCRYPT_HOST is used to defined the domain for which we want to create certicates with letsencrypt service.
SSL is automatically configured by the nginx-proxy if a certificate exists in the certs directory (certs can be generated by letsencrypt or other service).
So if you have a container with environment variables like:
```pre
VIRTUAL_HOST=www.d1.com, www.d2.com, www.d3.com
LETSENCRYPT_HOST=www.d3.com
```
and you only have a certs for www.d2.com then
* SSL will NOT be enabled for www.d1.com (no certificate)
* SSL will be enabled for www.d2.com (certificate exists in certs directory)
* SSL will be enabled for www.d3.com (certificate will be generate/renew with letsencrypt)

PunKeel:
That still doesn't tell us why `LETSENCRYPT_HOST` is defined on nginx and not on each container.

atomaka:
@PunKeel it is defined per guest container and not on the nginx-proxy container.  The README is a little misleading.

@JrCs I hadn't considered the case where an external certificate was being used. Still, the most common use case seems like it would have `LETSENCRYPT_HOST` matching `VIRTUAL_HOST`` every time and this is a little cumbersome to setup.

README.md
```diff
@@ -121,6 +121,28 @@ to establish a connection.  A self-signed or generic cert named `default.crt` an
...
+```
```
atomaka: This `docker run` command is a misleading.  `VIRTUAL_HOST` and the `LETSENCRYPT_*` environmental variables are set on your guest containers and not on the `jwilder/nginx-proxy` image.

ArthurHoaro:
I've tested https://github.com/JrCs/docker-nginx-proxy-letsencrypt and it works fine - except the README problem.

Possible improvements from user point of view (but not really necessary):

  * define the email once on the proxy container and use it when not specified.
  * if `LETSENCRYPT` is set to `1` (or any variable name/value) on a webapp container, use just use `VIRTUAL_HOST` value instead of copy-pasting it into `LETSENCRYPT_HOST`.

Toobian:
Good job. it's working ! almost... =/
When i try to enable LETSENCRYPT option, i have lot of redirect's error. It does not happen when i use self-signed certificate.

Do you have any idea ?

JrCs:
Closing this PR.
Better to use a dedicated container: [letsencrypt-nginx-proxy-companion](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion)

Issue #304, opened 2015-11-29
nginx.conf does not get updated with overlay network IPs


schmunk42:
I created an overlay-network for my swarm and can ping and wget a hello-world webserver in the swarm from the reverseproxy container.

This is my `/etc/hosts`

```
10.0.1.7        mynetwork_hello_1
```

And this the relevant part of `nginx.conf`

```
upstream ~my.test.com {
                        # mynetwork_hello_1
                        server :80;
}
```

As you can see the IP is missing.

Looks like somewhere [around here](https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl#L6) the IP does not get written correctly.

Any idea how to fix this or is this an issue with `docker-gen`?

Tested on: `docker 1.9.1, docker-compose 1.5.1`

schmunk42:
Related issue: https://github.com/jwilder/nginx-proxy/issues/97

thaJeztah:
This could be related to the new networking in docker, where a container can be attached to multiple networks. Only the default "bridge" networking propagates the old fields for IP-address of the container, but custom networks will not, and the IP address is returned in a nested list, see https://github.com/jwilder/docker-gen/issues/129

schmunk42:
@thaJeztah Thank you for the feedback. I am not a go-programmer (yet), but I looked a bit through the code and played around with `docker-gen`.

So it looks to me, that we first need support of the new networking feature from `fsouza/go-dockerclient`. Not until then we could use the VXLan IPs in the nginx template. Right?

I thought about an additional ENV variable like `VIRTUAL_NETWORK` to define which IP should be used, any thoughts about this?

md5:
@schmunk42 See https://github.com/fsouza/go-dockerclient/issues/407

schmunk42:
@thaJeztah @jwilder @md5 Some updates from my side.

1. I updated the client and added a very basic struct for the networks. This is working as intended so far. These are my very first steps in Go, so don't be too harsh ;) https://github.com/jwilder/docker-gen/compare/master...schmunk42:feature/docker-network

2. With [this template](https://github.com/jwilder/docker-gen/compare/master...schmunk42:feature/docker-network#diff-6a1a65847c0c1f0a2604f21ecfdc771bR22) I am able to insert the correct IPs, but with some glitches ... I need to specify `VIRTUAL_NETWORK` and `VIRTUAL_PORT` at the moment. I think I could get rid of the port definition, but I think we need something which defines the network to use.

[...]

Issue #305, opened 2015-11-29
Doesn't seem to work with new docker networking


xoredg:
It doesn't seem to work with the new docker networking features. i.e. `--x-networking` for example on docker compose. I'm not really sure if it should to begin with, but it would be handy.
I'm running a SPA, API and DB under a new networked setup and nginx-proxy gives a 503 even when the LB has the `VIRTUAL_HOST` env var. Are you aware of this?

thaJeztah:
possible duplicate of https://github.com/jwilder/nginx-proxy/issues/304?

blackxored:
It might be. The only difference is that I'm talking local networking, but yes the lacking support for the new networking API definitely seems like the culprit.

modius:
:+1: 

Trying to move from links to networking and this definitely appears to be an issue.

> While links are still supported in this limited capacity, you should avoid them in preference of Docker networks. The link feature is expected to be deprecated and removed in a future release.

https://docs.docker.com/engine/userguide/networking/dockernetworks/

modius:
@schmunk42 did this commit make it into the Docker image? 

https://github.com/jwilder/docker-gen/commit/fdf292247a233d4775e258979b3fa5e6b1c187b9

md5:
@modius No. But see https://github.com/jwilder/docker-gen/pull/144

tkw1536:
Related issue is closed, closing.

PR #337, opened 2016-01-17
Support Overlay Network

baptistedonaux:
- Migrate docker-gen from 0.4.2 to 0.5.0 (add []Network supports)
- Nginx template use now current IP into network.
- Documentation updated. Proxy must run it with the option --net=host to access at all differents networks (overlay network support - tested with ```docker-compose --x-networking up -d```)

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -1,4 +1,4 @@
-{{ $CurrentContainer := where $ "ID" .CurrentContainerID | first }}
+{{ $CurrentContainer := where $ "ID" .Docker.CurrentContainerID | first }}
 
 {{ define "upstream" }}
 	{{ if .Address }}
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -1,3 +1,5 @@
+{{ $CurrentContainer := where $ "ID" .CurrentContainerID | first }}
+
 {{ define "upstream" }}
 	{{ if .Address }}
 		{{/* If we got the containers from swarm and this container's port is published to host, use host IP:PORT */}}
@@ -75,16 +77,24 @@ server {
 upstream {{ $host }} {
 {{ range $container := $containers }}
 	{{ $addrLen := len $container.Addresses }}
-	{{ $network := index $container.Networks 0 }}
-	{{/* If only 1 port exposed, use that */}}
-	{{ if eq $addrLen 1 }}
-		{{ $address := index $container.Addresses 0 }}
-		{{ template "upstream" (dict "Container" $container "Address" $address "Network" $network) }}
-	{{/* If more than one port exposed, use the one matching VIRTUAL_PORT env var, falling back to standard web port 80 */}}
-	{{ else }}
-		{{ $port := coalesce $container.Env.VIRTUAL_PORT "80" }}
-		{{ $address := where $container.Addresses "Port" $port | first }}
-		{{ template "upstream" (dict "Container" $container "Address" $address "Network" $network) }}
+
+	{{ range $knownNetwork := $CurrentContainer.Networks }}
+		{{ range $containerNetwork := $container.Networks }}
+			{{ if eq $knownNetwork.Name $containerNetwork.Name }}
+				## Can be connect with "{{ $containerNetwork.Name }}" network
+
+				{{/* If only 1 port exposed, use that */}}
+				{{ if eq $addrLen 1 }}
+					{{ $address := index $container.Addresses 0 }}
+					{{ template "upstream" (dict "Container" $container "Address" $address "Network" $containerNetwork) }}
+				{{/* If more than one port exposed, use the one matching VIRTUAL_PORT env var, falling back to standard web port 80 */}}
+				{{ else }}
+					{{ $port := coalesce $container.Env.VIRTUAL_PORT "80" }}
+					{{ $address := where $container.Addresses "Port" $port | first }}
+					{{ template "upstream" (dict "Container" $container "Address" $address "Network" $containerNetwork) }}
+				{{ end }}
+			{{ end }}
+		{{ end }}
 	{{ end }}
 {{ end }}
 }
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -5,13 +5,13 @@
 			# {{ .Container.Node.Name }}/{{ .Container.Name }}
 			server {{ .Container.Node.Address.IP }}:{{ .Address.HostPort }};
 		{{/* If there is no swarm node or the port is not published on host, use container's IP:PORT */}}
-		{{ else }}
+		{{ else if .Network }}
 			# {{ .Container.Name }}
-			server {{ .Address.IP }}:{{ .Address.Port }};
+			server {{ .Network.IP }}:{{ .Address.Port }};
 		{{ end }}
-	{{ else }}
+	{{ else if .Network }}
 		# {{ .Container.Name }}
-		server {{ .Container.IP }} down;
+		server {{ .Network.IP }} down;
 	{{ end }}
 {{ end }}
 
@@ -75,15 +75,16 @@ server {
 upstream {{ $host }} {
 {{ range $container := $containers }}
 	{{ $addrLen := len $container.Addresses }}
+	{{ $network := index $container.Networks 0 }}
 	{{/* If only 1 port exposed, use that */}}
 	{{ if eq $addrLen 1 }}
 		{{ $address := index $container.Addresses 0 }}
-		{{ template "upstream" (dict "Container" $container "Address" $address) }}
+		{{ template "upstream" (dict "Container" $container "Address" $address "Network" $network) }}
 	{{/* If more than one port exposed, use the one matching VIRTUAL_PORT env var, falling back to standard web port 80 */}}
 	{{ else }}
 		{{ $port := coalesce $container.Env.VIRTUAL_PORT "80" }}
 		{{ $address := where $container.Addresses "Port" $port | first }}
-		{{ template "upstream" (dict "Container" $container "Address" $address) }}
+		{{ template "upstream" (dict "Container" $container "Address" $address "Network" $network) }}
 	{{ end }}
 {{ end }}
 }
--- a/test/docker.bats
+++ b/test/docker.bats
@@ -73,6 +73,7 @@ load test_helpers
 		-v /var/run/docker.sock:/tmp/docker.sock:ro \
 		-v $BATS_TEST_DIRNAME/../nginx.tmpl:/etc/docker-gen/templates/nginx.tmpl:ro \
 		--volumes-from bats-nginx \
+		--expose 80 \
 		jwilder/docker-gen:0.7.0 \
 			-notify-sighup bats-nginx \
 			-watch \
--- a/test/docker.bats
+++ b/test/docker.bats
@@ -54,15 +54,21 @@ load test_helpers
 @test "[$TEST_FILE] separated containers (nginx + docker-gen + nginx.tmpl)" {
 	docker_clean bats-nginx
 	docker_clean bats-docker-gen
+	docker_network_clean bats-docker-gen-network
+
+	# MAKE network
+	run docker network create bats-docker-gen-network
+	assert_success
 	
 	# GIVEN a simple nginx container
 	run docker run -d \
 		--name bats-nginx \
+		--net=bats-docker-gen-network \
 		-v /etc/nginx/conf.d/ \
 		-v /etc/nginx/certs/ \
 		nginx:latest
 	assert_success
-	run retry 5 1s docker run appropriate/curl --silent --fail --head http://$(docker_ip bats-nginx)/
+	run retry 5 1s docker run --net=bats-docker-gen-network appropriate/curl --silent --fail --head http://$(docker_ips bats-nginx)/
 	assert_output -l 0 $'HTTP/1.1 200 OK\r'
 
 	# WHEN docker-gen runs on our docker host
--- a/test/docker.bats
+++ b/test/docker.bats
@@ -54,21 +54,15 @@ load test_helpers
 @test "[$TEST_FILE] separated containers (nginx + docker-gen + nginx.tmpl)" {
 	docker_clean bats-nginx
 	docker_clean bats-docker-gen
-	docker_network_clean bats-docker-gen-network
 
-	# MAKE network
-	run docker network create bats-docker-gen-network
-	assert_success
-	
 	# GIVEN a simple nginx container
 	run docker run -d \
 		--name bats-nginx \
-		--net=bats-docker-gen-network \
 		-v /etc/nginx/conf.d/ \
 		-v /etc/nginx/certs/ \
 		nginx:latest
 	assert_success
-	run retry 5 1s docker run --net=bats-docker-gen-network appropriate/curl --silent --fail --head http://$(docker_ips bats-nginx)/
+	run retry 5 1s docker run appropriate/curl --silent --fail --head http://$(docker_ip bats-nginx)/
 	assert_output -l 0 $'HTTP/1.1 200 OK\r'
 
 	# WHEN docker-gen runs on our docker host
@@ -95,7 +89,7 @@ load test_helpers
 		docker logs bats-docker-gen
 		false
 	} >&2
-	
+
 	# THEN
 	assert_nginxproxy_behaves bats-nginx
 }
@@ -120,4 +114,3 @@ function assert_nginxproxy_behaves {
 	run curl_container $container /data --header "Host: webFOO.bats" --head
 	assert_output -l 0 $'HTTP/1.1 503 Service Temporarily Unavailable\r'
 }
-
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN echo "daemon off;" >> /etc/nginx/nginx.conf \
 RUN wget -P /usr/local/bin https://godist.herokuapp.com/projects/ddollar/forego/releases/current/linux-amd64/forego \
  && chmod u+x /usr/local/bin/forego
 
-ENV DOCKER_GEN_VERSION 0.4.2
+ENV DOCKER_GEN_VERSION 0.5.0
 
 RUN wget https://github.com/jwilder/docker-gen/releases/download/$DOCKER_GEN_VERSION/docker-gen-linux-amd64-$DOCKER_GEN_VERSION.tar.gz \
  && tar -C /usr/local/bin -xvzf docker-gen-linux-amd64-$DOCKER_GEN_VERSION.tar.gz \
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN echo "daemon off;" >> /etc/nginx/nginx.conf \
 RUN wget -P /usr/local/bin https://godist.herokuapp.com/projects/ddollar/forego/releases/current/linux-amd64/forego \
  && chmod u+x /usr/local/bin/forego
 
-ENV DOCKER_GEN_VERSION 0.5.0
+ENV DOCKER_GEN_VERSION 0.7.0
 
 RUN wget https://github.com/jwilder/docker-gen/releases/download/$DOCKER_GEN_VERSION/docker-gen-linux-amd64-$DOCKER_GEN_VERSION.tar.gz \
  && tar -C /usr/local/bin -xvzf docker-gen-linux-amd64-$DOCKER_GEN_VERSION.tar.gz \
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ update-dependencies:
 	docker pull python:3
 	docker pull rancher/socat-docker:latest
 	docker pull appropriate/curl:latest
-	docker pull docker:1.9
+	docker pull docker:1.10
 
 test:
 	docker build -t jwilder/nginx-proxy:bats .
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ See [Automated Nginx Reverse Proxy for Docker][2] for why you might want to use 
 
 To run it:
 
-    $ docker run -d --net=host -p 80:80 -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy
+    $ docker run -d -p 80:80 -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy
 
 Then start any containers you want proxied with an env var `VIRTUAL_HOST=subdomain.youdomain.com`
 
@@ -33,6 +33,19 @@ If you need to support multiple virtual hosts for a container, you can separate 
 
 You can also use wildcards at the beginning and the end of host name, like `*.bar.com` or `foo.bar.*`. Or even a regular expression, which can be very useful in conjunction with a wildcard DNS service like [xip.io](http://xip.io), using `~^foo\.bar\..*\.xip\.io` will match `foo.bar.127.0.0.1.xip.io`, `foo.bar.10.0.2.2.xip.io` and all other given IPs. More information about this topic can be found in the nginx documentation about [`server_names`](http://nginx.org/en/docs/http/server_names.html).
 
+### Multiple Networks
+
+With the new overlay network, your proxy can be deal with many containers in many networks. Default, if you don't use ```--net``` flag, your proxy will be attached at ```bridge``` default network. You can define your container with ```--net=your_network``` option.
+
+If your proxy try to access at a container in an unattached network, the request is successful.
+
+#### Connect Another Network
+
+In current Docker release (1.9), you can create a container with only one network. To attach others networks, you can use this command.
+```
+docker network connect my_network my_container
+```
+
 ### SSL Backends
 
 If you would like to connect to your backend using HTTPS instead of HTTP, set `VIRTUAL_PROTO=https` on the backend container.
@@ -41,7 +54,7 @@ If you would like to connect to your backend using HTTPS instead of HTTP, set `V
 
 To set the default host for nginx use the env var `DEFAULT_HOST=foo.bar.com` for example
 
-    $ docker run -d --net=host -p 80:80 -e DEFAULT_HOST=foo.bar.com -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy
+    $ docker run -d -p 80:80 -e DEFAULT_HOST=foo.bar.com -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy
 
 
 ### Separate Containers
@@ -56,13 +69,13 @@ To run nginx proxy as a separate container you'll need to have [nginx.tmpl](http
 First start nginx with a volume:
 
 
-    $ docker run -d --net=host -p 80:80 --name nginx -v /tmp/nginx:/etc/nginx/conf.d -t nginx
+    $ docker run -d -p 80:80 --name nginx -v /tmp/nginx:/etc/nginx/conf.d -t nginx
 
 Then start the docker-gen container with the shared volume and template:
 
 ```
 $ docker run --volumes-from nginx \
-    --net=host \
+    \
     -v /var/run/docker.sock:/tmp/docker.sock:ro \
     -v $(pwd):/etc/docker-gen/templates \
     -t jwilder/docker-gen -notify-sighup nginx -watch -only-exposed /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
@@ -79,7 +92,7 @@ certificates or optionally specifying a cert name (for SNI) as an environment va
 
 To enable SSL:
 
-    $ docker run -d --net=host -p 80:80 -p 443:443 -v /path/to/certs:/etc/nginx/certs -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy
+    $ docker run -d -p 80:80 -p 443:443 -v /path/to/certs:/etc/nginx/certs -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy
 
 The contents of `/path/to/certs` should contain the certificates and private keys for any virtual
 hosts in use.  The certificate and keys should be named after the virtual host with a `.crt` and
@@ -128,7 +141,7 @@ In order to be able to secure your virtual host, you have to create a file named
 /etc/nginx/htpasswd/$VIRTUAL_HOST
 
 ```
-$ docker run -d --net=host -p 80:80 -p 443:443 \
+$ docker run -d -p 80:80 -p 443:443 \
     -v /path/to/htpasswd:/etc/nginx/htpasswd \
     -v /path/to/certs:/etc/nginx/certs \
     -v /var/run/docker.sock:/tmp/docker.sock:ro \
@@ -176,7 +189,7 @@ RUN { \
 
 Or it can be done by mounting in your custom configuration in your `docker run` command:
 
-    $ docker run -d --net=host -p 80:80 -p 443:443 -v /path/to/my_proxy.conf:/etc/nginx/conf.d/my_proxy.conf:ro -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy
+    $ docker run -d -p 80:80 -p 443:443 -v /path/to/my_proxy.conf:/etc/nginx/conf.d/my_proxy.conf:ro -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy
 
 #### Per-VIRTUAL_HOST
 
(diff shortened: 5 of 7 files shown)


baptistedonaux:
If PR merged, missing Circle CI container creation with flag --net=host to deal with network differents of ```bridge```.

md5:
Can you explain the --net=host requirement better? That shouldn't be
necessary for an overlay network since the proxy should also be on the
overlay network.

baptistedonaux:
```--net=host``` flag isn't required while proxy container is in the same network that the container called. So, if all containers called (with VIRTUAL_HOST) are in default network (currently ```bridge```), the flag isn't required.

With the new overlay network and the --x-networking docker-compose flag (currently beta - 1.5), every application are isolated in specific network. Containers into ```bridge``` network can't call containers in other network. With ```--net=host``` flag, proxy container is on the localhost network. All virtual networks are callables.

**Comment by md5** [comment]
Based on the [Getting started with multi-host networking](https://docs.docker.com/engine/userguide/networking/get-started-overlay/) document, it seems like they're recommending `--net=my-net` where `my-net` is the name of the overlay network.

I think this should be documented in its own section of the readme instead of adding `--net=` flags to every `docker run` in the existing readme examples. That new section should use `--net=my-net` instead of `--net=host` and should link to https://docs.docker.com/engine/userguide/networking/get-started-overlay/ or some other Docker documentation.

**Comment by baptistedonaux** [comment]
Okay, ```--net=host``` flag is probably too much and require few explications. ```host``` "network" is the ultimate solution for proxy in front-end.

@md5 Do you want propose a README text block to insert ?

**Comment by md5** [comment]
I haven't really played around with overlay enough to write the section myself and I'm not sure when I'll get a chance to do so.

Regarding `--net=host`, I'm really reluctant to ever use it since it means that the container shares *all* network namespaces with the host (including Unix domain sockets). I've seen weird, unexpected things happen when `--net=host` is used (eg https://github.com/docker/docker/issues/5899), so I consider it an absolute last resort.

**Comment by baptistedonaux** [comment]
I understand. The best solution could be an aggregation of namespace networks. I currently search a best solution.

**Comment by baptistedonaux** [comment]
Currently, we can associate a container at many networks with ```docker network connect my_network my_container``` but only one network at container creation (https://github.com/docker/docker/pull/17796).

Can be a best alternative at ```--net=host``` flag.

**Comment by md5** [comment]
I agree the connecting explicitly to all the relevant networks is the best approach. It looks like support for doing this in a `docker-compose.yml` was added in https://github.com/docker/compose/pull/2564

**Comment by baptistedonaux** [comment]
@md5 Please feedback for English language and content writing.

**Comment by md5** [comment]
@baptistedonaux I opened https://github.com/baptistedonaux/nginx-proxy/pull/1 with my suggestions.

One thing that made me wonder is whether some attempt should be made in `nginx.tmpl` to filter the containers to only include the ones in the same networks as the `nginx-proxy` instance. That would require the ability to identify the container ID in which `docker-gen` is currently executing. I think https://github.com/jwilder/docker-gen/pull/146 may be a step in that direction, but I'm not sure that's possible currently.

**Comment by md5** [comment]
Looking closer at https://github.com/jwilder/docker-gen/pull/146, I don't think it actually does anything to help us figure out the container ID of the `nginx-proxy` container in which `docker-gen` is currently executing.

**Comment by baptistedonaux** [comment]
@md5 I was thinking at this feature. I try implement it and I come back to you.

**Comment by md5** [comment]
It looks like `docker-gen` could determine this from `/proc/self/cgroup`. Here's an example with `awk`:

```console
# awk -F/ '/cpu:/ { print $NF }' /proc/self/cgroup 
7f72ffdf3089df475a9cfe459fd0ec3b0d214686e794887ca17bc1b32a6f995b
```

`docker-gen` could pretty easily read the `/proc/self/cgroup` file itself directly, assuming this is consistent across Docker hosts.

baptistedonaux:
@md5 To retrieve current container name, I propose ```awk``` command:
```bash
awk -F"-|\." '/1:/ {print $3}' /proc/self/cgroup
```
Then, get container's networks for current network requires to know current container name. So I propose two solutions:
* Add into ```Docker``` structure a new field with name of current container.
```diff
type Docker struct {
   Name            string
   NumContainers   int
   NumImages       int
   Version         string
   ApiVersion      string
   GoVersion       string
   OperatingSystem string
   Architecture    string
+  CurrentContainerName string
}
 ```
* Add new template's method into ```template.go``` to retrieve container name (with awk).
```go
func getCurrentContainerName() (string, error);
```

In two cases, docker-gen requires a new feature implementation (or I need help).

baptistedonaux:
@md5 What do you think of my propositions ?

md5:
@baptistedonaux I don't think that using `awk` will be necessary. The Go code should be able to deal with `/proc/self/cgroup` directly.

I think adding the current container id (or name) to the `Docker` struct makes sense. I started looking at the code to do so, but the code for populating `$.Docker` is spread around a bit and I didn't end up writing any code.

baptistedonaux:
@md5 I try too on a git branch to prepare the next implementation (matches between proxy container and networks).

baptistedonaux:
I have implemented a parser for ```/proc/self/cgroup```. I'm not satisfied because parse a file with split methods doesn't seem me "conventional". Default to find (to know) a best solution, this is my implementation.

https://github.com/baptistedonaux/docker-gen/commit/dc155e150814c35c80e8189e9efabb80704c06e9

I wait your implementation or your feedback.

md5:
I commented on the commit. Perhaps @jwilder can take a look when he has a minute.

jwilder:
Agree w/ all of @md5 comments.  Format of my `/proc/self/cgroup` matches @md5's as well.

baptistedonaux:
I propose this PR https://github.com/jwilder/docker-gen/pull/158

baptistedonaux:
> to filter the containers to only include the ones in the same networks as the nginx-proxy instance itself. @md5 

I have began the implementation.

Two questions:
* What's the best way to compare two networks (Gateway IP vs Name) ? https://github.com/baptistedonaux/nginx-proxy/blob/master/nginx.tmpl#L83
* Today, when I connect a container at a network, no events sent to re-generate template. I wait the debian repository update to test with the new 1.10 release.

md5:
@baptistedonaux I think doing it by network name should be fine. It looks like it might be worth adding some sort of `intersect` function to `docker-gen`, though.

baptistedonaux:
Add ```intersect``` function can be nice to limit fat code in ```nginx.tmpl```. But intersection can't just make a comparison between two Network structures because all items aren't similars.

Brief for two containers sharing the same network.

|   | Container 1     | Container 2     |
|:-:|:------------------:|:------------------:|
| IP | 10.0.0.2 | 10.0.0.3 |
| Name | foo | foo |
|Gateway | 10.0.0.1 | 10.0.0.1 |
EndpointID | x | y |

The prototype could be this :
```golang
/**
 * This function takes two arrays Networks.
 * It returns an array with n1 arrays Networks found in n2 variable.
 */
func ([]Network) intersect(n1 []Network, n2 []Network);
```

Currently, this function name already exists, so I propose ```intersectNetworks```. @md5 What's your opinion ?

baptistedonaux:
I try implement a solution. https://github.com/baptistedonaux/docker-gen/commit/5164d88caf41b0db333bfb671d68bae4f0c9894b

md5:
@baptistedonaux Thanks for your work on this. After seeing that function, I feel like the potential improvement in `nginx.tmpl` isn't worth adding yet another function to `docker-gen`. If it is added, I think the function name needs to make it clearer that it's matching based on `Network.Name`.

baptistedonaux:
So, I'm not sure to understand you.

- Should we keep the function and rename it ?
- Or simply don't add it and keep to match networks in template fully ?

md5:
Let's see what @jwilder says, but I'd probably leave the logic in the template as you have it now.

jwilder:
Template is a bit gnarly, but I'm fine with it like that for now.

baptistedonaux:
@md5 @jwilder So I keep the code without adding new function.

baptistedonaux:
@jwilder I think that my PR is ready. Unit tests are unsuccessful because linking at jwilder/docker-gen#158 to identify current container and its networks.

If none networks isn't shared between current container (nginx-proxy) and destination container (with ```VIRTUAL_HOST``` variable), second container isn't linked.

shouze:
:+1: will fix #304, #305 (and already closed #358, #353)

md5:
@baptistedonaux I think it would be great if there were some tests around this functionality. @thomasleveil did some great work a few months back to implement an initial test suite.

nginx.tmpl
```diff
@@ -1,3 +1,5 @@
+{{ $CurrentContainer := where $ "ID" .Docker.CurrentContainerID | first }}
```
jwilder: `docker-gen` 0.7.0 should have this field available now.

md5: I just realized this will break in the case that docker-gen is in a
different container from nginx and the two are not on the same networks.
Definitely an edge case, but it should at least be covered in the docs if
it isn't handled correctly.

baptistedonaux: @jwilder  Thanks, I'm currently working to run the tests.

@md5 Yes of course, but it's a problem. The best use case stays to use docker-gen and nginx in the same container. Currently, the only test which fails is docker.bats (nginx + dcoker-gen + nginx.tpml). I works on and I'm come back make a checkpoint.

baptistedonaux:
@md5 I have few fails with unit test. Can you build jwilder/nginx-proxy with docker-gen 0.7.0 and try run unit test ?

pitkley:
I've analyzed those issues quite a bit, and from what I can tell, they were introduced with [jwilder/docker-gen@`5043565`](https://github.com/jwilder/docker-gen/commit/50435652b16b3874a5907765d33fa31150876d92). Furthermore, they are independent of this PR, the tests fail on master too with any version of docker-gen since the mentioned commit.

From looking at the logs of the proxy-containers in `bats` runs, the events of containers starting, stopping and dying come in with a high delay, somewhere between 8 to 15 seconds (on a pretty low-power machine I tested it on). The `curl`-containers try to connect before the configuration has been regenerated, which causes the tests to fail.

As far as I can tell, the [locking in `SetDockerEnv`](https://github.com/jwilder/docker-gen/commit/50435652b16b3874a5907765d33fa31150876d92#diff-05a3aa0ab9e5687ff167e0ef3ae4c5a6R43) causes this delay. Removing it improved the test-results, but it is still intermittent if they fail or not.

I have [a patch](https://gist.github.com/pitkley/363b51414ff4bfe001db) which adds delays between starting the proxies and the first `curl` command, and removes a proxy after it is no longer needed. While this improves the situation too, I don't think it is guaranteed to work, and not clean at all.

Since I am neither comfortable with Go nor with docker-gen in general, I haven't posted this as an issue over at the docker-gen repo, since I don't know if it is an issue on docker-gen side.

@jwilder Do you think this is an issue with docker-gen? If not, how should we go about fixing the `bats` tests?

baptistedonaux:
ping @jwilder @pitkley

pitkley:
As a general follow-up on this pull request: I have successfully updated all my proxied version 1 compose-files to version 2, joining the containers into a single, user-defined network `nginxproxy`, so :+1: on the efforts.

What I have come across: if a container that defines `VIRTUAL_HOST` does not share a network with the nginx-proxy, the `upstream` block will be empty (and the `server` blocks exist although they probably shouldn't).

I guess that this requires checking for intersecting networks before creating the `upstream` or `server` blocks.

Other than that I had no issues what-so-ever, very cool stuff! :+1:

hasufell:
> joining the containers into a single, user-defined network nginxproxy

That breaks isolation. Now I have different web applications from different customers in one network. That's not really good.

I tried to do it the other way around: create networks for the web applications and then make the front proxy connect to each one of them. However, that doesn't seem to work, requests time out, although the generated nginx config looks ok.

pitkley:
You are right that isolated networks wasn't my use case, but I tried those too and they worked for me.

This [bash script](https://gist.github.com/pitkley/12164e1782147dbf3207) sets up a simple multi-network, multi-container scenario. (**Note!** The script assumes that this PR is available as the `nginx-proxy:overlay` image.) Can you try that out? Or were you talking about a different setup?

jeanpralo:
I have successfully migrated my compose from v1 to v2 and I did not see any problem so far.

Did a few test on multi network even though I do not have any use at the moment but it seems to be working fine.

Thanks for the work, hoping it gets merged soon :+1:

jwilder:
If anyone can help get the CI tests passing, I'd love to merge this.

baptistedonaux:
@jwilder Of course ! In my case, I have not the same fail. Locally, I have only one or two tests which abort. I retry tomorrow and I paste the test result to debug together.

jeanpralo:
@baptistedonaux it looks like this PR is not up to date as well, I am trying to have a look locally from your  fork it seems $CurrentContainer ends up being empty in the test

pitkley:
@jwilder @baptistedonaux 
As far as I can tell, the only test that is failing because of this PR is the `separated containers (nginx + docker-gen + nginx.tmpl)` in `docker.bats`. This is because of the reason @jeanpralo mentions: the `$CurrentContainer` variable doesn't get populated when run inside the docker-gen container -- unfortunately I have no idea why.

All the other tests are intermittent failures, presumably due to the locking introduced in [jwilder/docker-gen@`5043565`](https://github.com/jwilder/docker-gen/commit/50435652b16b3874a5907765d33fa31150876d92). They now surface because of the docker-gen 0.7.0 dependency, but are independent of this PR.

I am currently in the process of trying to fix the unit tests to work with docker-gen 0.7.0 independent of this PR, and I think I might have a solution. My proposal: I will open a separate PR lifting nginx-proxy to docker-gen 0.7.0 and fixing the unit tests, and if that PR is merged, @baptistedonaux can rebase this PR onto master, so he only has to worry about the one unit test.

If everyone is fine with that, I'll open a PR sometime (hopefully...) in the next 24 hours :+1:

jwilder:
@pitkley @baptistedonaux Sounds good. Thanks!  

@pitkley I don't see how the locking is causing failures off-hand.  `SetDockerEnv` is called once when `docker-gen` starts.  `SetServerInfo` is called before generating the templates.  It called more frequently and does a little more work while holding the write lock, but that shouldn't take much time at all.  If you have a way to reproduce the issue w/ `docker-gen`, could you log an issue in the `docker-gen` repo?

baptistedonaux:
```$CurrentContainerID``` is empty in the test because they are executed outside Docker. So no container ID found.

**Comment by jeanpralo** [comment]
@baptistedonaux  I have seen it empty when trying to execute the tests with containers on a docker-machine. 

What seems odd is that `$` does not contain the docker-gen container. Do not know if this is a bug from docker-gen as I have seen there is some special code for swarm. I just figured I'll say it here even if it's not really related to this PR directly.

**Comment by baptistedonaux** [comment]
I executed test in local just after have built ```jwilder/nginx-proxy:bats```.

```bash
$ bats test
 ✓ [default-host] DEFAULT_HOST=web1.bats
 ✓ [docker] start 2 web containers
 ✓ [docker] -v /var/run/docker.sock:/tmp/docker.sock:ro
 ✓ [docker] -v /var/run/docker.sock:/f00.sock:ro -e DOCKER_HOST=unix:///f00.sock
 ✓ [docker] -e DOCKER_HOST=tcp://...
 ✗ [docker] separated containers (nginx + docker-gen + nginx.tmpl)
   (from function `assert_output' in file test/lib/bats/batslib.bash, line 328,
    from function `assert_nginxproxy_behaves' in file test/docker.bats, line 104,
    in test file test/docker.bats, line 94)
     `assert_nginxproxy_behaves bats-nginx' failed
   
   
   -- line differs --
   index    : 0
   expected : HTTP/1.1 503 Service Temporarily Unavailable
   actual   : HTTP/1.1 200 OK
   --
   
 ✓ [multiple-hosts] start a nginx-proxy container
 ✓ [multiple-hosts] nginx-proxy forwards requests for 2 hosts
 ✓ [multiple-ports] start a nginx-proxy container
 ✓ [multiple-ports] nginx-proxy defaults to the service running on port 80
 ✓ [multiple-ports] VIRTUAL_PORT=90 while port 80 is also exposed
 ✓ [multiple-ports] single exposed port != 80
 ✓ [wildcard-hosts] start a nginx-proxy container
 ✓ [wildcard-hosts] VIRTUAL_HOST=*.wildcard.bats
 ✓ [wildcard-hosts] VIRTUAL_HOST=wildcard.bats.*
 ✓ [wildcard-hosts] VIRTUAL_HOST=~^foo\.bar\..*\.bats

16 tests, 1 failure
```

Only one fail for me. Maybe a network latency problem ?

pitkley:
Just a heads up: I have added PR #396 which fixes (at least as far as I can tell...) the other unit-tests on Circle-CI. It might have something to do with the btrfs storage-driver used by Circle-CI, which maybe increases wait times to the point of failures due to timeouts.

pitkley:
@baptistedonaux I found the issue for the `separated containers (nginx + docker-gen + nginx.tmpl)` test failing:

```diff
diff --git a/test/docker.bats b/test/docker.bats
index b7a52c0..232d584 100644
--- a/test/docker.bats
+++ b/test/docker.bats
@@ -73,10 +73,9 @@ load test_helpers
                -v /var/run/docker.sock:/tmp/docker.sock:ro \
                -v $BATS_TEST_DIRNAME/../nginx.tmpl:/etc/docker-gen/templates/nginx.tmpl:ro \
                --volumes-from bats-nginx \
                jwilder/docker-gen:0.7.0 \
                        -notify-sighup bats-nginx \
                        -watch \
-                       -only-exposed \
                        /etc/docker-gen/templates/nginx.tmpl \
                        /etc/nginx/conf.d/default.conf
```

Since docker-gen is instructed to filter for containers exposing a port, the docker-gen container itself will be excluded from the available containers in the template. So we either have to remove the `-only-exposed` line, or include a dummy port:

```diff
diff --git a/test/docker.bats b/test/docker.bats
index b7a52c0..232d584 100644
--- a/test/docker.bats
+++ b/test/docker.bats
@@ -73,10 +73,9 @@ load test_helpers
                -v /var/run/docker.sock:/tmp/docker.sock:ro \
                -v $BATS_TEST_DIRNAME/../nginx.tmpl:/etc/docker-gen/templates/nginx.tmpl:ro \
                --volumes-from bats-nginx \
+               --expose 1024 \
                jwilder/docker-gen:0.7.0 \
                        -notify-sighup bats-nginx \
                        -watch \
                        -only-exposed \
                        /etc/docker-gen/templates/nginx.tmpl \
                        /etc/nginx/conf.d/default.conf
```

While this fixes this test, this does not fix the issue @md5 [mentioned](https://github.com/jwilder/nginx-proxy/pull/337#discussion_r53999225) when the nginx container is in a different network to the docker-gen container. I don't know if that edge-case is per se fixable, but at least some docs should exist.

baptistedonaux:
@pitkley Thank you for your research and explication. I'm currently trying.

baptistedonaux:
Good for me. The last epic test successful thanks @pitkley !

```bash
bats test
 ✓ [default-host] DEFAULT_HOST=web1.bats
 ✓ [default-host] stop all bats containers
 ✓ [docker] start 2 web containers
 ✓ [docker] -v /var/run/docker.sock:/tmp/docker.sock:ro
 ✓ [docker] -v /var/run/docker.sock:/f00.sock:ro -e DOCKER_HOST=unix:///f00.sock
 ✓ [docker] -e DOCKER_HOST=tcp://...
 ✓ [docker] separated containers (nginx + docker-gen + nginx.tmpl)
 ✓ [docker] stop all bats containers
 ✓ [multiple-hosts] start a nginx-proxy container
 ✓ [multiple-hosts] nginx-proxy forwards requests for 2 hosts
 ✓ [multiple-hosts] stop all bats containers
 ✓ [multiple-ports] start a nginx-proxy container
 ✓ [multiple-ports] nginx-proxy defaults to the service running on port 80
 ✓ [multiple-ports] VIRTUAL_PORT=90 while port 80 is also exposed
 ✓ [multiple-ports] single exposed port != 80
 ✓ [multiple-ports] stop all bats containers
 ✓ [wildcard-hosts] start a nginx-proxy container
 ✓ [wildcard-hosts] VIRTUAL_HOST=*.wildcard.bats
 ✓ [wildcard-hosts] VIRTUAL_HOST=wildcard.bats.*
 ✓ [wildcard-hosts] VIRTUAL_HOST=~^foo\.bar\..*\.bats
 ✓ [wildcard-hosts] stop all bats containers

21 tests, 0 failures
```

pitkley:
The fails now probably stem from the fact that the `circle.yml` uses Docker 1.9.1 and not 1.10. I'm not sure why it fails under Docker 1.9, but I tried it on a 1.9.1 install and got similar errors to your last run.

Maybe try and change it to use [Docker 1.10 (beta)](https://discuss.circleci.com/t/docker-1-10-0-is-available-beta/2100):

```diff
diff --git a/Makefile b/Makefile
index 213a441..3a10ad4 100644
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ update-dependencies:
        docker pull python:3
        docker pull rancher/socat-docker:latest
        docker pull appropriate/curl:latest
-       docker pull docker:1.9
+       docker pull docker:1.10
 
 test:
        docker build -t jwilder/nginx-proxy:bats .
diff --git a/circle.yml b/circle.yml
index e0ecf4d..761684f 100644
--- a/circle.yml
+++ b/circle.yml
@@ -1,7 +1,6 @@
 machine:
   pre:
-    - sudo curl -L -o /usr/bin/docker 'https://s3-external-1.amazonaws.com/circle-downloads/docker-1.9.1-circleci'
-    - sudo chmod 0755 /usr/bin/docker
+    - curl -sSL https://s3.amazonaws.com/circle-downloads/install-circleci-docker.sh | bash -s -- 1.10.0
   services:
     - docker
 
diff --git a/test/lib/docker_helpers.bash b/test/lib/docker_helpers.bash
index 1735506..221234e 100644
--- a/test/lib/docker_helpers.bash
+++ b/test/lib/docker_helpers.bash
@@ -62,5 +62,5 @@ function docker_tcp {
                --expose 2375 \
                -v /var/run/docker.sock:/var/run/docker.sock \
                rancher/socat-docker
-       docker run --label bats-type="docker" --link "$container_name:docker" docker:1.9 version
+       docker run --label bats-type="docker" --link "$container_name:docker" docker:1.10 version
 }
```

baptistedonaux:
@pitkley Can be true. Instead of install manually docker, why not use the simplest way ? And simplified the installation process ?

```yml
# circle.yml
machine:
  services:
    - docker
```

In the [Circle CI documentation](https://circleci.com/docs/docker#deployment-to-a-docker-registry), no manual installation describes to use Docker.

pitkley:
Because the default version on Circle-CI [is 1.8.2](https://circleci.com/gh/pitkley/circle-ci-docker-test/1) (see the `docker version` section under "Dependencies"):

```console
$ docker version
Client:
 Version:      1.8.2-circleci-cp-workaround
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   4008b9c-dirty
 Built:        Wed Dec 16 10:05:40 UTC 2015
 OS/Arch:      linux/amd64

Server:
 Version:      1.8.2-circleci-cp-workaround
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   4008b9c-dirty
 Built:        Wed Dec 16 10:05:40 UTC 2015
OS/Arch: linux/amd64
```

baptistedonaux:
@pitkley Correct. Circle Ci conf updated

pitkley:
@baptistedonaux I don't see why the build still fails, so for completeness I tested the following scenarios:

|                | **Docker 1.9.1** | **Docker 1.10.0** | **Docker 1.10.3** |
|----------------|------------------|-------------------|-------------------|
| **Overlay FS** | fails            | works             | works             |
| **btrfs**      | fails            | works             | works             |

So we should be able to rule out btrfs as a potential source for the issues.

-----

As an aside inspired by those fails: If this PR gets merged, people won't be able to use nginx-proxy if they have a Docker version that doesn't support overlay networks. Can the `nginx.tmpl` be adapted to use the old technique of getting the upstream address if `$CurrentContainer.Networks` or `$container.Networks` has no value?

schmunk42:
@pitkley Many networking features were updated in 1.10.

I.e. when I try to run compose `v2` configs with `network` on a `1.9` swarm it fails, saying it needs `1.10` - so I don't think 1.9 needs to be tested for these networking features (I'd say it's not supported).

baptistedonaux:
@pitkley In my case, I runs Docker with AUFS driver storage.

pitkley:
@schmunk42 I agree that we shouldn't have to support 1.9 in regards to the overlay networking features. What I'm talking about is that if somebody uses 1.8 or lower or don't want to use the overlay networking at all, they still should be able to use nginx-proxy.

-----

@baptistedonaux I tested it under AUFS as well, only 1.10.3 though, but also had no issues there. It seems to be something different. I'm still troubleshooting Circle-CI if I can find the issue there...

pitkley:
@baptistedonaux I got good and bad news, I found [the issue](https://circleci.com/gh/pitkley/circle-ci-docker-test/19):

```console
$ docker run nginx cat /proc/self/cgroup
11:name=systemd:/user/1002.user/4.session/lxc/box185
10:hugetlb:/user/1002.user/4.session/lxc/box185
9:perf_event:/user/1002.user/4.session/lxc/box185
8:blkio:/user/1002.user/4.session/lxc/box185
7:freezer:/user/1002.user/4.session/lxc/box185
6:devices:/user/1002.user/4.session/lxc/box185
5:memory:/user/1002.user/4.session/lxc/box185
4:cpuacct:/user/1002.user/4.session/lxc/box185
3:cpu:/user/1002.user/4.session/lxc/box185
2:cpuset:/user/1002.user/4.session/lxc/box185
```

The container ID doesn't get exposed by Circle-CI through `/proc/self/cgroup`, thus the generation of the config fails. One alternative to using cgroups is the [`$HOSTNAME` environment variable](http://stackoverflow.com/a/21000924/758165) in conjunction with `/proc/self/mountinfo`:

```console
# echo $HOSTNAME
991c24d485ec

# cat /proc/self/mountinfo
664 652 0:36 /box1259/rootfs/var/lib/docker/btrfs/subvolumes/c9e1b36a7b2c52348f03888a6f05599db02a301f2f6a59b65fe6afeb189c3a02 / rw,nodev,noatime - btrfs /dev/xvdh rw,nodatasum,nodatacow,ssd,space_cache
665 664 0:271 / /proc rw,nodev,relatime - proc proc rw
666 664 0:272 / /dev rw,nosuid,nodev - tmpfs tmpfs rw,mode=755,uid=231072,gid=231072
667 666 0:273 / /dev/pts rw,relatime - devpts devpts rw,gid=231077,mode=620,ptmxmode=666
668 664 0:274 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime - tmpfs tmpfs rw,mode=755,uid=231072,gid=231072
669 664 0:36 /box1259/rootfs/var/lib/docker/containers/991c24d485ec3a13198b101a98152228194305243f10266fd0215d0689e4bb22/resolv.conf /etc/resolv.conf rw,nodev,noatime - btrfs /dev/xvdh rw,nodatasum,nodatacow,ssd,space_cache
670 664 0:36 /box1259/rootfs/var/lib/docker/containers/991c24d485ec3a13198b101a98152228194305243f10266fd0215d0689e4bb22/hostname /etc/hostname rw,nodev,noatime - btrfs /dev/xvdh rw,nodatasum,nodatacow,ssd,space_cache
671 664 0:36 /box1259/rootfs/var/lib/docker/containers/991c24d485ec3a13198b101a98152228194305243f10266fd0215d0689e4bb22/hosts /etc/hosts rw,nodev,noatime - btrfs /dev/xvdh rw,nodatasum,nodatacow,ssd,space_cache
672 666 0:269 / /dev/shm rw,nosuid,nodev,noexec,relatime - tmpfs shm rw,size=65536k,uid=231072,gid=231072
673 666 0:46 / /dev/mqueue rw,nosuid,nodev,noexec,relatime - mqueue mqueue rw
674 666 0:5 /null /dev/null rw,relatime - devtmpfs udev rw,size=30904560k,nr_inodes=7726140,mode=755
675 666 0:5 /zero /dev/zero rw,relatime - devtmpfs udev rw,size=30904560k,nr_inodes=7726140,mode=755
676 666 0:5 /full /dev/full rw,relatime - devtmpfs udev rw,size=30904560k,nr_inodes=7726140,mode=755
677 666 0:5 /tty /dev/tty rw,relatime - devtmpfs udev rw,size=30904560k,nr_inodes=7726140,mode=755
678 666 0:5 /urandom /dev/urandom rw,relatime - devtmpfs udev rw,size=30904560k,nr_inodes=7726140,mode=755
679 666 0:5 /random /dev/random rw,relatime - devtmpfs udev rw,size=30904560k,nr_inodes=7726140,mode=755
681 665 0:5 /null /proc/kcore rw,relatime - devtmpfs udev rw,size=30904560k,nr_inodes=7726140,mode=755
682 665 0:5 /null /proc/latency_stats rw,relatime - devtmpfs udev rw,size=30904560k,nr_inodes=7726140,mode=755
683 665 0:5 /null /proc/timer_stats rw,relatime - devtmpfs udev rw,size=30904560k,nr_inodes=7726140,mode=755

# grep "$HOSTNAME" /proc/self/mountinfo | head -n1 | perl -n -e '/([0-9a-f]{64})/ && print "$1\n"'
991c24d485ec3a13198b101a98152228194305243f10266fd0215d0689e4bb22
```

I would suggest adapting docker-gen to try to find the ID in `cgroup`, and if it doesn't find it there, use `/proc/self/mountinfo` similar to how I did it in pure shell commands above.

baptistedonaux:
@pitkley I think ```$HOSTNAME``` isn't good reference because it is overridable.

On my host, I have ```/proc/self/mountinfo``` too. So I think that this file can be a sure source. I implement it.

pitkley:
@baptistedonaux `/proc/self/mountinfo` can contain more than one 64-byte hexadecimal ID, see my example. At least on Circle-CI, you will have to use the `$HOSTNAME` variable. There might be another alternative, but I haven't found one yet...

thomasleveil:
Keep in mind that CircleCI compile their own custom docker engine binary. Make sure that issue isn't specific to CircleCI

thomasleveil:
See https://github.com/thomasleveil/nginx-proxy/commit/c85839bd71836e83c56df50264d8ed493e7a31dc for switching-eds-with-python back to TravisCI (if it helps)

baptistedonaux:
@thomasleveil What result with Travis CI ?

pitkley:
@thomasleveil Thanks for the input on that! I guess that this is rather specific, because I haven't heard of `/proc/self/cgroup` not providing the ID. I wouldn't "blame" it on the Circle-CI build without further analysis though, because it might have to do with the LXC setup they use rather than with the build itself.

Anyway, this would probably be a "custom-tailored" solution to get CI happy, with the side-effect of supporting the low percentage of other users who might run a special build of Docker that doesn't expose the ID through the cgroups directly.

While switching to Travis is an option (and a quick test shows that the [cgroups are available](https://travis-ci.org/pitkley/circle-ci-docker-test/builds/118301546#L205)), I'm not sure if @jwilder would be too happy about switching.

I think that this is @jwilder's call: "monkeypatch" docker-gen such that it tries to find the container ID elsewhere (like `/proc/self/mountinfo`), or switch CI to e.g. Travis?

baptistedonaux:
@pitkley True, only @jwilder can switch to Travis CI.

Neverheless, @thomasleveil was right because the last Travis built successful. https://travis-ci.org/thomasleveil/nginx-proxy/builds/95239874

pitkley:
@baptistedonaux Furthermore, I have tried to build and test this PR, [which worked too](https://travis-ci.org/pitkley/circle-ci-docker-test/builds/118311889). So Travis-CI definitely seems like a viable option! :+1:

jwilder:
I'm fine switching to Travis CI.  This project used to use it as well, but switched to Circle CI when the test suite was added.  See: #246 for why.

baptistedonaux:
@jwilder Great, Travais fail seems solved.

@thomasleveil @pitkley Somebody to propose a PR (switch Circle to Travais) or I add this changement in current PR.

baptistedonaux:
@jwilder Travis test works

test/docker.bats
```diff
@@ -73,6 +73,7 @@ load test_helpers
 		-v /var/run/docker.sock:/tmp/docker.sock:ro \
 		-v $BATS_TEST_DIRNAME/../nginx.tmpl:/etc/docker-gen/templates/nginx.tmpl:ro \
 		--volumes-from bats-nginx \
+		--expose 80 \
```
md5: Forgive me if this is covered in the previous comments, but why is this change needed?

nikashitsa:
@baptistedonaux any updates? Can't wait for it 😊

baptistedonaux:
I wait a feedback of @jwilder to advance this PR.

/cc @md5 @pitkley @nikashitsa

test/docker.bats
pitkley: @md5 See [this comment](https://github.com/jwilder/nginx-proxy/pull/337#issuecomment-200768674).

baptistedonaux: @pitkley @md5 Unit test are fixed and Travis executes with success.

jwilder:
Thanks @baptistedonaux!

lessless:
does that mean that it's possible to use with v2 syntax?

baptistedonaux:
It's a feature only available with the v2 syntax. In v1, it's impossible to manage networks.

PR #339, opened 2016-01-19
Updated intermediate ciphers and added optional alternatives

jacobtomlinson:
It looks like Mozilla have moved their specifications for ciphers and updated their intermediate spec.

https://wiki.mozilla.org/Security/Server_Side_TLS

They also list a "Modern" and an "Old" set of ciphers. I've added the option to set the env var `DEFAULT_CIPHER_SUITE` to select either of these, not setting this will keep the default of "Intermediate".


buchdag:
This was superseded by #957

PR #343, opened 2016-01-21
Add an ability to handle a group isolation.

### LINKED ISSUE #340: Is it possible to let a certain proxy server only serve a certain group of containers based on the port of proxy server?
As titled, I an gonna start several proxy servers with different ports on my host, and I want to start several containers which are the same that can be forwarded to depending ports of proxy servers. 

For instance, I have two proxy servers listening on port 80 and 81, and I start two same containers with same VIRTUAL HOST. Is it possible that I can access each container depenging the port that proxy server listened. I thought maybe adding an another env variable and modifing some template code might do the work? Or is there any better solution?

Hope that you can understand the situation I described ;(

qazbnm456:
As titled, this fixes #340 
Mean while, I update version and readme as well.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM nginx:1.9.6
+FROM nginx:1.9.9
 MAINTAINER Jason Wilder mail@jasonwilder.com
 
 # Install wget and install/updates certificates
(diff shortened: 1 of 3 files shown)


jwilder:
This PR is fairly old and not something I want to support.  Thanks!

roba91:
Too bad. I'm also interested in this feature (having multiple ngingx-proxies running on different ports handling different groups of containers).

PR #359, opened 2016-02-11
Set appropriate X-Forwarded-Ssl header

sw-double:
Many projects look for X-Forwarded-Ssl header, so it can simplify things for some users.
On the other hand I can see no downside in merging this.


jwilder:
@sw-double Can you rebase?

sw-double:
Rebased.

jwilder:
Thanks @sw-double!

PR #361, opened 2016-02-13
Host an generated list of proxyfied sites

marema31:
Hello,

I wanted to have a dynamic list of the sites proxyfied by this container. And try to mockup it.

It will generate a index.html listing all the VIRTUAL_HOST from the container. It will use the content of env variable TITLE for each container to replace the link by something more useful.

I don't know either go and go templating and this may be not sufficient for all the cases adressed by origin, but I think it could be a nice-to-have feature to be enhanced by more specialized people.


jwilder:
Thanks @marema31.  I think this would be better suited in a custom image derived from this image.

Issue #268, opened 2015-10-15
[bug] Using certificates missing apps redirect to wrong apps


rokcarl:
All seems to work as intended until I start using certificates.
### Short description

I do:
- run nginx-proxy, app1 and app2 both on https,
- I kill app2,
- visit app2 in my browser, but get app1 certificates and results.
### Long description

This is how I run nginx-proxy and both apps (app1 and app2).

```
docker run --name nginx-proxy -d -p 80:80 -p 443:443 -v /home/ubuntu/_certificates:/etc/nginx/certs -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy 
docker run -d --name app1.app.com -e VIRTUAL_PROTO=https -e VIRTUAL_PORT=443 -p 8003:443 -e VIRTUAL_HOST=app1.app.com app 
docker run -d --name app2.app.com -e VIRTUAL_PROTO=https -e VIRTUAL_PORT=443 -p 8005:443 -e VIRTUAL_HOST=app2.app.com app
```

I kill app2 with `docker kill app2.app.com`.
Visit app2.app.com, but the certificate I get is from app1 and the contents as well. If I also kill app1, then I get the server not found, which is of course correct. If I then run app1 again, app2 again redirects to the wrong app.

rokcarl:
Any ideas on how to solve this?

rokcarl:
This issue is very problematic for our use case as it's serving the wrong app. Imagine getting Facebook when you browse to Google...

Do you have any tips on where to look, maybe I can help solve this issue.

jwilder:
That's odd.  When you access the app, are you doing it over HTTP and then getting redirected to HTTPS?  Also, how are your certs named on disk?  Are they named after each virtual host or are they sharing a single wildcard `app.com` cert?

rokcarl:
Accessing all the apps through HTTPS. The certs are named the same as the domains, so `app1.app.com.crt` and `app1.app.com.key` for app1 and similarly `app2.app.com.crt` and `app2.app.com.key` for app2.

davidsyoung:
Having same issue here @md5 @rokcarl. Any luck fixing it so far @rokcarl?

md5:
I wonder whether a lack of [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) support in the browser could be to blame.

What browser/OS combo are you seeing these problems with @davidsyoung and @rokcarl?

[...]

PR #367, opened 2016-02-18
Modifiy nginx.tmpl to allow both 443 and 80.

tdemalliard:
Allow http when https is activated. This parameter avoid infinite redirect when the reverse proxy nginx-proxy is already behind another reverse proxy.

This parameter is not recommended for production. Use it for test / transition.

To use it, launch your container with the environnement variable ALLOW_80=1 :

    $ docker run -e VIRTUAL_HOST=foo.bar.com  ...

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -113,7 +113,28 @@ server {
 	server_name {{ $host }};
 	listen 80 {{ $default_server }};
 	access_log /var/log/nginx/access.log vhost;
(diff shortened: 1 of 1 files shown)


genezys:
This looks like a duplicate of #298

tdemalliard:
indeed it is a duplicate of #268.

I cancel my pull as we do not need duplicates.

PR #408, opened 2016-04-05
Add alpine image

### LINKED ISSUE #292: Using Alpine as a base?
Before I attempt to port this to a smaller image using Alpine, is it possible, and have you guys thought about it?

I have an alpine image that has nginx installed and it's 6MB, so I don't think that forego and docker-gen make it 185MB.

chuangbo:
Alpine linux is a tiny linux distribution which ideal for building
docker images.

for more commit details, this commit:
- Build from nginx:alpine (nginx:alpine does not have versions)
- Add full nginx.conf as nginx:alpine has a different nginx.conf then debian
- Download forego in Dockerfile for better building caching
- Change badge version ;-)

Fixed #292

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,18 @@
-FROM nginx:1.9.12
+FROM nginx:alpine
 MAINTAINER Jason Wilder mail@jasonwilder.com
 
 # Install wget and install/updates certificates
-RUN apt-get update \
- && apt-get install -y -q --no-install-recommends \
+RUN apk add --no-cache --virtual .run-deps \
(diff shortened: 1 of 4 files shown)


hasufell:
Alpine runs musl libc and often with downstream patches on specific packages which may or may not be reviewed. It's unclear what the side effects of running musl system-wide are. But there are side effects.

chuangbo:
Yes I agree. For the downstream compatibility we should release alpine to tag `alpine` but not `latest`.

jwilder:
Fixed via #558

PR #409, opened 2016-04-05
Add SKIP_HSTS support

wader:
Hi, is this useful to others? could it be done in a better way? if so i could fix it up, add tests, doc etc.
I would like to skip HSTS for a domain as it might be moved to another owner in the future.

Maybe better to make HSTS max-age configurable and if set to 0 just skip it?


wader:
Fixed by #969

PR #425, opened 2016-04-21
Fix for docker-compose v2 syntax

nikashitsa:
This PR solves #304 
Based on @duck1123 code

Notice: containers should be in same network. So, you may need to add `network_mode: 'bridge'` to every service in docker-compose.yml.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -5,9 +5,12 @@
 			# {{ .Container.Node.Name }}/{{ .Container.Name }}
 			server {{ .Container.Node.Address.IP }}:{{ .Address.HostPort }};
 		{{/* If there is no swarm node or the port is not published on host, use container's IP:PORT */}}
-		{{ else }}
+		{{ else if .Address.IP }}
 			# {{ .Container.Name }}
 			server {{ .Address.IP }}:{{ .Address.Port }};
+		{{ else }}
+			# {{ .Container.Name }}
+			server {{ .Container.Name }}:{{ .Address.Port }};
 		{{ end }}
 	{{ else }}
 		# {{ .Container.Name }}


tonivdv:
@nikashitsa it works only if I run a container after the proxy is started ... the other way around it doesn't work. It's not that big of deal, but in v1 mode that wasn't an issue.

pascalandy:
@tonivdv Is it a challenge? My for it's the logic sequence to launch things :)

tonivdv:
@pascalandy It happens I start a project and notice that my proxy was not running (I run the proxy for multiple projects on the same machine), and thus I need to restart them. As I said not a big issue, but I share it so others know about it, as I though it didn't work at first.

djmaze:
@pascalandy And what happens after a reboot, when containers are restarted (with `--restart=always`)? They would need to be restarted in order, which is only possible when using the (deprecated) linking feature.

nikashitsa:
@tonivdv not sure how to fix it correct. Perhaps, this PR #337 fixed it already. Any ideas?

pascalandy:
In a dev context, good points @tonivdv @djmaze. I didn't test that's much the restart scenario or a hard shutdown scenario at this point. 

I also miss the linking feature. I felt safer with it :)

Do you knows guys if an orchestrator like Rancher or DC/OS would have rules regarding this?

duck1123:
If you run the proxy with restart: always then it works fine. I haven't looked into the sequence of things yet, but I was thinking that if you ran the generator template once before nginx was started then it should leave you with a valid config that won't cause nginx to crash on startup.

iBobik:
Btw, will be good to include working example of docker-compose.yml in v2 syntax.

pascalandy:
Check this out - https://github.com/fatk/docker-letsencrypt-nginx-proxy-companion-examples

iBobik:
There are examples just for wiring nginx and jwilder/docker-gen images, not how to use nginx-proxy.

nikashitsa:
@tonivdv
> it works only if I run a container after the proxy is started ...

Just checked this example:

```
# docker-compose.yml
version: '2'
services:
  nginx:
    build: .
    network_mode: 'bridge'
    environment:
      VIRTUAL_HOST: 'mysite.loc' 
```  

```
# docker-compose.yml - nginx-proxy file in another folder
version: '2'
services:
  nginx-proxy:
    image: nikashitsa/proxy
    ports:
      - '80:80'
      - '443:443'
    network_mode: 'bridge'
    volumes:
```

Tested with different launch orders. All is working.

tonivdv:
@nikashitsa I will test again, but note that I tested this in a custom network and not bridge. Maybe that's the difference?!

nikashitsa:
@tonivdv maybe. Without `network_mode: 'bridge'` it doesn't work for me.

md5:
It seems to me that a fix that requires `network_mode: bridge` is not the right fix. I could see this working when all containers are in a single custom network (as would be the case when all the containers are defined in a single compose file), but that seems to limit the usefulness of the proxy.

The better fix for this issue is in https://github.com/jwilder/nginx-proxy/pull/337

nikashitsa:
@md5 agree with you. `network_mode: bridge` is not the best solution. #337 has no action for last 17 days, perhaps they need some help.

tonivdv:
@md5 @nikashitsa I tested with this fork https://github.com/rckrdstrgrd/nginx-proxy , but AFAI see it's the same change and I can use it in the following way:

    docker run -d -p 80:80 --name nginx-proxy --net adsdaq_proxy-tier -v /var/run/docker.sock:/tmp/docker.sock:ro rckrdstrgrd/nginx-proxy 

The above instruction I launch separately as I have different projects

The in my compose

```yaml

services:
  nginx:
      build: nginx
      ports:
          - 80
      environment:
          VIRTUAL_HOST: adsdaq.dev
      networks:
          - proxy-tier

networks:
  proxy-tier:
    driver: bridge
```

We then execute 

    docker-compose -p adsdaq up -d

As you can see no bridge mode ...

md5:
@tonivdv It looks like you're using bridge mode to me:

```yaml
networks:
  proxy-tier:
    driver: bridge
```

Doesn't the use of `-p adsdaq` in the `docker-compose` command and `--net adsdaq_proxy-tier` when you start `nginx-proxy` mean that you're using the same bridged network defined in the compose file? I would think it also means you need to `docker-compose up ...` your containers before your proxy.

md5:
@nikashitsa It took a while to get things sorted out for #337. It looks like it's working correctly, so I think it's just a matter of @jwilder having time to review and merge the changes.

tonivdv:
Hey @md5 

> Doesn't the use of -p adsdaq in the docker-compose command and --net adsdaq_proxy-tier when you start nginx-proxy mean that you're using the same bridged network defined in the compose file?

Yes indeed. I thought when talking about bridge it was the default bridge network and not the "custom" bridge network. But yes, I use the network with bridge driver.

> I would think it also means you need to docker-compose up ... your containers before your proxy.

yes. That's what I was explaining to @nikashitsa . It seems obvious though as the network doesn't exist yet. But for me/us it's not that big of an issue. We are going to right a little bash script so we can rapidly (re)start the proxy with the necessary networks.

But if #337 fixes this too then perfect :)

sdurrheimer:
#337  has been merged, this PR can be closed :).

nikashitsa:
Cool!

PR #442, opened 2016-05-04
Updated docker-gen to version 0.7.1 (fixes CurrentContainerID empty issue)

cjdell:
Tested that this fixes the problem that I and many others have had due to an updated version of CoreOS. The underlying problem was a change in the way cgroups are represented in the file `/proc/self/cgroup`. This uses an updated version of `docker-gen` which can deal with new and old formats of this file.


jwilder:
Fixed via #441

PR #455, opened 2016-05-12
mv nginx.tmpl to templates dir

ricardolee:
I move the nginx.tmpl to /app/templates.
It couse some problem when docker mount a file by volume


jwilder:
Thanks, I would prefer to keep the template in it's current location.

Issue #438, opened 2016-05-02
Error: no servers are inside upstream in


mrvini:
I've updated my proxy image today, and tried to restart all my other containers behind the proxy, however all of them failed, am I doing something wrong? ( i did follow explanations in issue #64, however that didn't help)

proxy

```
docker run -d --name nginx-proxy \
    -p 80:80 -p 443:443 \
    --restart=always \
    -v /opt/my-certs:/etc/nginx/certs \
    -v /var/run/docker.sock:/tmp/docker.sock:ro \
    jwilder/nginx-proxy
```

my dev container (nodejs)  built locally and it exposes port 8181

```
docker run -d --name www.dev1 \
    --restart=always \
    --link db --link redis \
    -e VIRTUAL_PORT=8181 \
    -e VIRTUAL_PROTO=https \
    -e VIRTUAL_HOST=dev1.mysite.com \
    -v /opt/my-volume/web/dev1/:/opt/my-volume/web/ \
    -v /opt/my-certs:/opt/my-certs:ro \
    -w /opt/my-volume/web/ localhost:5000/www \
    bash -c 'npm start server.js'
```

Right before I run dev container, i can see output of nginx -t

```
root@fba41f832f35:/app# nginx -t  
nginx: the configuration file /etc/nginx/nginx.conf syntax is ok
nginx: configuration file /etc/nginx/nginx.conf test is successful
```

After I start dev container, i see the following

```
root@fba41f832f35:/app# nginx -t        
2016/05/02 07:15:49 [emerg] 69#69: no servers are inside upstream in /etc/nginx/conf.d/default.conf:34
nginx: [emerg] no servers are inside upstream in /etc/nginx/conf.d/default.conf:34
nginx: configuration file /etc/nginx/nginx.conf test failed
```

when I check  /etc/nginx/conf.d/default.conf i see empty upstream

```
upstream dev1.mysite.com {
}
```

Is there anything I am doing wrong? I've been using same startup script for a good 6 month and it used to work right before I pulled the new image, did anything changed? Please help

dehy:
Same problem here.
docker 1.9.1cs2 on docker cloud

dehy:
I had to revert back to a72c7e6e20df3738ca365bf6c14598f6a8017500

wader:
@mrvini can you paste `docker network inspect $(docker network ls -q)`?

klaszlo:
Same error here. 
[...]

PR #459, opened 2016-05-16
Updates Readme on using Separate Containers

Giymo11:
- makes first command more readable
- adds name to docker-gen container
- adds volume for /etc/nginx/certs, which is needed by the latest .tmpl
- removes -only-exposed flag, it causes problems as discussed in issue #438

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,18 +73,26 @@ You may want to do this to prevent having the docker socket bound to a publicly 
 
 To run nginx proxy as a separate container you'll need to have [nginx.tmpl](https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl) on your host system.
 
-First start nginx with a volume:
+First start nginx (or nginx:alpine, etc) with a volume:
 
-
-    $ docker run -d -p 80:80 --name nginx -v /tmp/nginx:/etc/nginx/conf.d -t nginx
+```
+$ docker run -d -p 80:80 \
+    --name nginx-proxy \
+    -v /tmp/nginx:/etc/nginx/conf.d \
+    -t nginx
+```
 
 Then start the docker-gen container with the shared volume and template:
 
 ```
-$ docker run --volumes-from nginx \
+$ docker run -d \
+    --name nginx-proxy-gen \
+    --volumes-from nginx-proxy \
     -v /var/run/docker.sock:/tmp/docker.sock:ro \
-    -v $(pwd):/etc/docker-gen/templates \
-    -t jwilder/docker-gen -notify-sighup nginx -watch -only-exposed /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
+    -v $(pwd)/certs:/etc/nginx/certs:ro \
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,18 +73,26 @@ You may want to do this to prevent having the docker socket bound to a publicly
...
+$ docker run -d \
+    --name nginx-proxy-gen \
+    --volumes-from nginx-proxy \
     -v /var/run/docker.sock:/tmp/docker.sock:ro \
-    -v $(pwd):/etc/docker-gen/templates \
-    -t jwilder/docker-gen -notify-sighup nginx -watch -only-exposed /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
+    -v $(pwd)/certs:/etc/nginx/certs:ro \
+    -v $(pwd)/nginx.tmpl:/etc/docker-gen/templates/nginx.tmpl:ro \
+    -t jwilder/docker-gen \
+        -notify-sighup nginx-proxy -wait 5s:30s -watch /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
```
jwilder: This should have `-only-exposed` and no `-wait 5s:30s` to match the `nginx-proxy` image.

README.md
```diff
@@ -73,18 +73,26 @@ You may want to do this to prevent having the docker socket bound to a publicly
...
 
 ```
-$ docker run --volumes-from nginx \
+$ docker run -d \
+    --name nginx-proxy-gen \
+    --volumes-from nginx-proxy \
-    -v $(pwd):/etc/docker-gen/templates \
-    -t jwilder/docker-gen -notify-sighup nginx -watch -only-exposed /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
+    -v $(pwd)/certs:/etc/nginx/certs:ro \
```
jwilder: Can you remove this as it's not required.

README.md
wader: The problem with using `-only-exposed` in a separate containers setup with the new `nginx.tmpl` is that the container running docker-gen won't be included in the template context as it does not expose any ports. That in turn will cause `{{ $CurrentContainer := where $ "ID" .Docker.CurrentContainerID | first }}` to evaluate to `nil`.

jwilder: Ah.  Good point.

README.md
Giymo11: @jwilder Ah, my bad. It was giving me an error while testing, but I can't seem to reproduce the issue. Probably some kind of mistake on my side.

Giymo11: @jwilder I was able to reproduce the issue: When adding a container with `docker run -e VIRTUAL_HOST=foo.bar` I get the error `template error: template: nginx.tmpl:115:27: executing "nginx.tmpl" at <dir "/etc/nginx/cert...>: error calling dir: open /etc/nginx/certs: no such file or directory`, this is why I added the volume, as I am not comfortable fixing the nginx.tmpl.

jwilder:
This was fixed via #480 and https://github.com/jwilder/nginx-proxy/commit/a0dee5c833fc651cf77fc15a63069f0e34229cb7

Issue #453, opened 2016-05-11
Don't enable 'Strict-Transport-Security' header when HTTPS_METHOD=noredirect


palamccc:
When HTTPS_METHOD=noredirect is enabled, objective is to serve both 'http' and 'https' without  any conflict/redirect.

So in this case, if we enable 'Strict-Transport-Security' header, browser won't load http, after visiting https page. So  'Strict-Transport-Security' header should be disabled when HTTPS_METHOD=noredirect is used.

`add_header Strict-Transport-Security "max-age=31536000";`

related to PR #298 @kamermans

kamermans:
Good point, I agree.  My main purpose was for machine-to-machine traffic, so it did not occur to me that HSTS should be disabled.  @jwilder what do you think?  @palamccc makes a good point that there's really no point in explicitly saying you don't want to auto-redirect to HTTPS just to have HSTS force your browser to auto-redirect to HTTPS :)

kamermans:
@jwilder - this PR can be closed - I implemented the fix last year and now users can also use `HSTS=off`.

Issue #446, opened 2016-05-05
Improve Documentation for "env var to disable SSL redirect".


tobiashochguertel:
I'm missing where to set this new environment variables.. It would be create to add this information to the `readme.md` file. It was for me not clear that this env-variable has to be defined in web-application container (exposes port 80 or port 443, or both).

kamermans:
I've improved the docs on PR #462 - can you take a look and suggest improved wording if it is still not clear?

PR #462, opened 2016-05-20
Disable HSTS when HTTPS_METHOD=noredirect

kamermans:
First, thanks to @palamccc for reporting this in #453.

When using `HTTP_METHOD=noredirect`, the user is essentially saying "don't automatically redirect my users from HTTP to HTTPS, however, users that end up at the HTTPS site are still being sent the HSTS header, which forces the client to redirect all HTTP requests to HTTPS.  My feeling is that HSTS should be removed when `HTTPS_METHOD=noredirect`.

I have implemented the change in this PR and added tests to cover the normal HSTS behavior and the `noredirect` behavior (after sifting through the bats docs for an hour before stumbling on `refute_output` :) ).

I've also expanded the documentation on `HTTPS_METHOD` as suggested in #446

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,8 +143,12 @@ a 503.
 
 To serve traffic in both SSL and non-SSL modes without redirecting to SSL, you can include the
 environment variable `HTTPS_METHOD=noredirect` (the default is `HTTPS_METHOD=redirect`).  You can also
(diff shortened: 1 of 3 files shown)


jwilder:
Thanks @kamermans!

PR #472, opened 2016-06-03
Add env vars for custom template and nonstandard ports

edgimar:
The LISTEN_PORT_HTTP and LISTEN_PORT_HTTPS env variables can now be specified
on web-service containers in order to cause the proxy to listen on the
specified ports.  If you only want to allow listening on a port with SSL, these
variables can be set to the value "None" which causes them to not be used.  So,
for example, you could set LISTEN_PORT_HTTP=None in order to only proxy traffic
over a secure link.  If one or both of these env vars are omitted, the default
values of 80 and 443 will be used where omitted.  NOTE: these ports are
internal ports on the container.  If you set LISTEN_PORT_HTTPS=1234, and want
to make this accessible on the WAN at port 5555, then you will do something
like

```
docker run --expose 1234 -p 5555:1234 ...
```

Additionally, on the proxy container, a CUSTOM_TEMPLATE env variable can be
specified, which will instruct docker-gen to use the specified template instead
of the default template.  For example, if CUSTOM_TEMPLATE=/app/mytemplate, then
you will want to either copy your template to the container with 'docker cp',
or mount it from the host filesystem with something like

```
docker run -v /host/path:/app/mytemplate ...
```

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -19,4 +19,19 @@ if [ "$socketMissing" = 1 -a "$1" = forego -a "$2" = start -a "$3" = '-r' ]; the
 	exit 1
 fi
 
+# if a custom template has not been specified
+if [ -z "$CUSTOM_TEMPLATE" ]; then
+	# try to copy default config to another file if the other file doesn't yet exist
+	cp -n nginx.tmpl nginx_default.tmpl
+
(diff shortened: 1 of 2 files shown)


buchdag:
Custom ports part superseded by #1353

PR #480, opened 2016-06-12
Fix template error when /etc/nginx/certs does not exist

jwilder:


nginx.tmpl
```diff
@@ -112,11 +112,13 @@ upstream {{ $host }} {
 {{ $certName := (first (groupByKeys $containers "Env.CERT_NAME")) }}
 
 {{/* Get the best matching cert  by name for the vhost. */}}
+{{ $vhostCert := "" }}
+{{ if exists "/dir/nginx/certs" }}
```
wader: s/dir/etc/ ?

jwilder: Fixed.

README.md
```diff
@@ -19,6 +19,32 @@ The containers being proxied must [expose](https://docs.docker.com/reference/run
...
+    environment:
+      - VIRTUAL_HOST=whoami.local
+```
+
+```shell
+$ docker-compose up
+$ curl -H "Host: whoami.local" localhost
+I''m 5b129ab83266
+```
+
```
wader: I wonder if it would be good to add some more information about how to use nginx-proxy with multiple compose projects, using external network, bridge mode etc

jwilder: Definitely!  This is all I could add in the time I had to look at it today.

jonas-wilhelm:
Is it possible, that this change broke SSL support completely unless Env.CERT_NAME is set? As far as I understand, this does not work:

```
	{{ var $vhostCert := "" }}
	{{ if exists "/etc/nginx/certs" }}
		{{ $vhostCert := (closest (dir "/etc/nginx/certs") (printf "%s.crt" $host)) }}
		{{/* vhostCert is actually a filename so remove any suffixes since they are added later */}}
		{{ $vhostCert := replace $vhostCert ".crt" "" -1 }}
		{{ $vhostCert := replace $vhostCert ".key" "" -1 }}
	{{ end }}
```
I think `$vhostCert` is always empty after the `if` block since `$vhostCert`, my local tests seem to verify that. Or am I totally wrong?

jonas-wilhelm:
golang playground example regarding variable scope in templates: https://play.golang.org/p/4rFhZkpqvF

PR #482, opened 2016-06-13
Revert 9c93efa

jwilder:
Updating a variable inside an `if` does update the outer scoped variable.


ryneeverett:
So the error #480 addressed is still not fixed, right?

ryneeverett:
I'm guessing this is fixed by https://github.com/jwilder/docker-gen/pull/200 now.

PR #484, opened 2016-06-14
Add APP_KEY env variable to allow multiple nginx-proxy containers on one host

pysysops:
With the changes in this PR you can run multiple nginx-proxy containers on a single Docker host and pin selected backend containers to specific proxies.

This would be useful in cases where you might deploy 1 nginx and 4 api containers as a "deployable unit" where you might use a build number, a git short commit, git tag or other "APP_KEY" to assign backends to different proxies for example, in blue-green style deployment scenarios. 

If there is an APP_KEY, the nginx-proxy will add an "X-App-Key" header that can be utilised in deployment scripts or for building simple dashboards showing what version or what "side" (blue or green) of an app you are on.

I don't think there is a pull request to offer this or any issues regarding it.

Thanks,

Tim

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,32 @@ $ docker run -d -p 80:80 -p 443:443 \
 
 You'll need apache2-utils on the machine where you plan to create the htpasswd file. Follow these [instructions](http://httpd.apache.org/docs/2.2/programs/htpasswd.html)
 
+### Multiple nginx-proxy Containers
+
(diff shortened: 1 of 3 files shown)


jwilder:
Thanks @pysysops.  This is an interesting idea, but I don't think it's a good fit for this project.  This kind of functionality is supported by allowing custom templates to be used.

PR #493, opened 2016-06-27
Add docker-compose file for separate containers.

ryneeverett:
Demonstrate that this pattern works.

This is based on the example at
<https://github.com/fatk/docker-letsencrypt-nginx-proxy-companion-examples/blob/master/docker-compose/v2/simple-site/docker-compose.yml>.

### DIFF
--- a/docker-compose-separate-containers.yml
+++ b/docker-compose-separate-containers.yml
@@ -7,8 +7,6 @@ services:
       - "80:80"
     volumes:
       - /etc/nginx/conf.d
-    networks:
-      - proxy-tier
 
   dockergen:
     image: jwilder/docker-gen
@@ -18,16 +16,8 @@ services:
     volumes:
       - /var/run/docker.sock:/tmp/docker.sock:ro
       - ./nginx.tmpl:/etc/docker-gen/templates/nginx.tmpl
-    networks:
-      - proxy-tier
 
   whoami:
     image: jwilder/whoami
     environment:
       - VIRTUAL_HOST=whoami.local
-    networks:
-      - proxy-tier
-
-networks:
-  proxy-tier:
-    driver: bridge
--- a/docker-compose-separate-containers.yml
+++ b/docker-compose-separate-containers.yml
@@ -1,0 +1,33 @@
+version: '2'
+services:
+  nginx:
+    image: nginx
+    container_name: nginx
+    ports:
+      - "80:80"
+    volumes:
+      - /etc/nginx/conf.d
+    networks:
+      - proxy-tier
+
+  dockergen:
+    image: jwilder/docker-gen
+    command: -notify-sighup nginx -watch /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
+    volumes_from:
+      - nginx
+    volumes:
+      - /var/run/docker.sock:/tmp/docker.sock:ro
+      - ./nginx.tmpl:/etc/docker-gen/templates/nginx.tmpl
+    networks:
+      - proxy-tier
+
+  whoami:
+    image: jwilder/whoami
+    environment:
+      - VIRTUAL_HOST=whoami.local
+    networks:
+      - proxy-tier
+
+networks:
+  proxy-tier:
+    driver: bridge
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ services:
 ```shell
 $ docker-compose up
 $ curl -H "Host: whoami.local" localhost
-I''m 5b129ab83266
+I'm 5b129ab83266
 ```
 
 ### Multiple Ports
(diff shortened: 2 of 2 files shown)


jwilder:
Thanks @ryneeverett!

docker-compose-separate-containers.yml
```diff
@@ -0,0 +1,34 @@
...
+      - proxy-tier
+
+  dockergen:
+    image: jwilder/docker-gen
+    command: -notify-sighup nginx -watch /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
+    volumes_from:
+      - nginx
+    volumes:
+      - /var/run/docker.sock:/tmp/docker.sock:ro
+      - ./:/etc/docker-gen/templates
```
wader: should this volume be included?

docker-compose-separate-containers.yml
```diff
@@ -0,0 +1,34 @@
+version: '2'
+services:
+  nginx:
+    image: nginx
+    container_name: nginx
+    ports:
+      - "80:80"
```
wader: add 443:443?

docker-compose-separate-containers.yml
```diff
@@ -0,0 +1,34 @@
...
+    volumes:
+      - /tmp/nginx:/etc/nginx/conf.d
```
wader: should this volume be included?

docker-compose-separate-containers.yml
ryneeverett: A volume is needed to get the nginx.tmpl, but it might as well be a single-file volume. (I didn't realize that that was allowed, but it works.)

docker-compose-separate-containers.yml
ryneeverett: This is necessary for nginx to have access to the generated conf file. I suspect I'm missing your point.

docker-compose-separate-containers.yml
wader: ah yeah sorry, when running with separate i guess the point is to have your own mounted in somehow

docker-compose-separate-containers.yml
wader: just for debug or any other reason?

docker-compose-separate-containers.yml
ryneeverett: I'm not sure I see the point given that there aren't any certs. But I can add it if you like.

docker-compose-separate-containers.yml
ryneeverett: There must be a volume because dockergen mounts it with `--volumes-from nginx` below. It does not have to be mounted on /tmp but that is how the README documents doing it so I did it that way here as well.

docker-compose-separate-containers.yml
wader: ah yeah skip i guess

docker-compose-separate-containers.yml
wader: ok that make sense

ryneeverett: I'm removing mounting /tmp on the host. Suggesting to mount a static path on the host when not necessary seems like a recipe for a confusing situation where two docker-gen's are writing to the same volume.

huiwang:
i don't understand why the given example creates two networks :(

ryneeverett:
@huiwang I only get one -- "nginxproxy_proxy-tier".

wader:
@huiwang by example you mean https://github.com/fatk/docker-letsencrypt-nginx-proxy-companion-examples/blob/master/docker-compose/v2/simple-site/docker-compose.yml? that example involves two networks, the external network "nginx-proxy" (has the name "proxy-tier" in the config file) and the default network created by docker-compose named based on project name  ...in this case i would guess  "simple-site_default"

@ryneeverett you sure there is a "nginxproxy_proxy-tier" network? not "nginx-proxy"?

ryneeverett:
> @ryneeverett you sure there is a "nginxproxy_proxy-tier" network? not "nginx-proxy"?

@wader In this PR, yes. But if @huiwang was referencing the single-container example in the README that would clear up the confusion.

huiwang:
@wader and @ryneeverett I thanks for your reply. I'm referring to the example at https://github.com/fatk/docker-letsencrypt-nginx-proxy-companion-examples/blob/master/docker-compose/v2/simple-site/docker-compose.yml 

I still don't see the advantage to create two networks.

ryneeverett:
@huiwang Are you asking about "proxy-tier" vs. "nginx-proxy" in the example? They are the same network but nginx-proxy is the name it was created with prior to running compose (see that project's README). If you look in the diff for this PR I've eliminated the external part so you don't need to create the network manually.

huiwang:
@ryneeverett Indeed, I was talking about proxy-tier vs nginx-proxy. Thx for the clarification, there is indeed only one network. 

Regarding the compose file in the PR, I don't find it necessary to specify networks because we are just using the default one. Refer to the section Specifying custom networks in https://docs.docker.com/compose/networking/

ryneeverett:
> Regarding the compose file in the PR, I don't find it necessary to specify networks because we are just using the default one. Refer to the section Specifying custom networks in https://docs.docker.com/compose/networking/

You're right. I just pushed an update.

PR #511, opened 2016-07-18
Mitigate httpoxy attack by suppressing `Proxy` request header

kamermans:
The so-called `httpoxy` attack [see httpoxy.org, CVE-2016-(5385-5388,1000109-1000110)] occurs when an errant client sends an HTTP `Proxy` request header.  This is non-standard, and no well-behaved client will send this header.  Many languages (PHP, Python, Golang), when used in a CGI scenario expose this header as an environment variable `HTTP_PROXY` in the CGI environment (as per the CGI spec).  Unfortunately, many server-side HTTP clients use that environment variable as the address of an outbound HTTP proxy, meaning an attacker can easily coerce a backend server to send private, internal HTTP traffic out to the public internet, to a destination of their choice.

This PR tells Nginx to suppress that header, and not forward it on to the backend servers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,10 +205,14 @@ proxy_set_header Connection $proxy_connection;
 proxy_set_header X-Real-IP $remote_addr;
 proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 proxy_set_header X-Forwarded-Proto $proxy_x_forwarded_proto;
+
+# Mitigate httpoxy attack (see README for details)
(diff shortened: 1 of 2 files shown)


wader:
LGTM, maybe add comment in template and README.md why it's done? CVE ref etc

kamermans:
Good idea @wader, done!

jwilder:
Thanks @kamermans!

PR #517, opened 2016-07-29
FIX pass http basic authentication to upstream proxy

ctroncoso:
needed by nginx to process basic http authentication requests.


ctroncoso:
disregard

PR #518, opened 2016-07-29
upgrade nginx to 1.11.3

huiwang:
There is a known bug as described here https://stackoverflow.com/questions/36907767/nsurlerrordomain-code-1004-for-few-seconds-after-app-start-up/37444271#37444271

One solution is to upgrade nginx to 1.11.0 http://stackoverflow.com/a/37444404/341307


bartoszhernas:
+1 for immediate merge. I am expecting the same problem right now.

joe11051105:
+1

auchri:
Why not update to [1.11.3](https://nginx.org/en/CHANGES)?

SeerUK:
It looks like [1.11.1](https://hub.docker.com/r/library/nginx/tags/) is the latest officially released Docker image to base this off of, @auchri.

bernex:
+1

tve:
There is an official nginx 1.11.3 docker image. I would also appreciate an upgrade to this version. I had to fork and do it myself in order to fix http/2 issues with nginx pre-1.11.0. Thanks!

SeerUK:
The image has since been updated. Let's just assume that whenever this happens it'll be to whatever the latest mainline version of Nginx is.

jwilder:
@huiwang Can you bump this to `1.11.3`?

auchri:
I think it should be changed in https://github.com/jwilder/nginx-proxy/blob/master/Makefile too

huiwang:
@auchri thx for your advice, @jwilder done !

jwilder:
Thanks @huiwang!

PR #555, opened 2016-09-01
Update ciphers and HTST settings to get A+ rating

freakinruben:
The default config gets you an 'A' rating. Cipher settings are copied from [Mozilla SSL Configartion Generator](https://mozilla.github.io/server-side-tls/ssl-config-generator/?server=nginx-1.10.1&openssl=1.0.1t&hsts=yes&profile=intermediate)


jwilder:
Thanks @freakinruben!

mplx:
This patch introduces HSTS enabled for subdomains too - a behaviour which nginx-proxy didn't have before and might not be desired by everyone. In a scenario where you use https://domain.tld with nginx-proxy you're suddenly required to make all existing subdomains (even those not managed by nginx-proxy) to be available by HTTPS or they won't be accessable anymore for quite a long time...

PR #558, opened 2016-09-02
Add alpine base image

matt-hh:
Hi,

this PR adds a new alpine linux based version. 
Background: Nginx alpine ships with a newer openssl version and supports ALPN.

```
bash-4.3# nginx -V
nginx version: nginx/1.11.3
built by gcc 5.3.0 (Alpine 5.3.0)
built with OpenSSL 1.0.2h  3 May 2016
TLS SNI support enabled
```

I think it's better to keep the current nginx-proxy with debian and provide a second variant.

Try it: https://hub.docker.com/r/matthh/nginx-proxy/tags/

I simply created another automated build with "/Dockerfile.alpine" for "Dockerfile Location".
BTW: Different file sizes are amazing...

TODO:
- Create a new automated build with "alpine-latest" tag
- Add info to README

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -1,0 +1,31 @@
+FROM nginx:1.11.3-alpine
+MAINTAINER Jason Wilder mail@jasonwilder.com
+
+# Install wget and install/updates certificates
+RUN apk add --no-cache --virtual .run-deps \
+    ca-certificates bash wget \
+    && update-ca-certificates
+
+# Configure Nginx and apply fix for very long server names
+RUN echo "daemon off;" >> /etc/nginx/nginx.conf \
+ && sed -i 's/^http {/&\n    server_names_hash_bucket_size 128;/g' /etc/nginx/nginx.conf
+
+# Install Forego
+ADD https://github.com/jwilder/forego/releases/download/v0.16.1/forego /usr/local/bin/forego
+RUN chmod u+x /usr/local/bin/forego
+
+ENV DOCKER_GEN_VERSION 0.7.3
+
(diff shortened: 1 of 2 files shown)


ihor-sviziev:
It would be very good to have such images. @jwilder any reason to do not merge this PR?

matt-hh:
Hey alpine fans, i have updated to the latest available versions.

```
bash-4.3# nginx -V
nginx version: nginx/1.11.8
built by gcc 5.3.0 (Alpine 5.3.0)
built with OpenSSL 1.0.2j  26 Sep 2016
TLS SNI support enabled
```

Dockerfile.alpine
```diff
@@ -0,0 +1,31 @@
...
+RUN apk add --no-cache --virtual .run-deps \
+    ca-certificates bash wget \
+    && update-ca-certificates
+
+# Configure Nginx and apply fix for very long server names
+RUN echo "daemon off;" >> /etc/nginx/nginx.conf \
+ && sed -i 's/^http {/&\n    server_names_hash_bucket_size 128;/g' /etc/nginx/nginx.conf
+
+# Install Forego
+ADD https://github.com/jwilder/forego/releases/download/v0.16.1/forego /usr/local/bin/forego
```
jwilder: I'm surprised this actually works in alpine. I usually need to compile alpine binaries statically.

jwilder:
Thanks @matt-hh!

jwilder:
I've setup an automated build w/ an `alpine` tag that will build using this dockerfile for HEAD on master.  Would you mind submitting a PR to update the `README`?

ihor-sviziev:
@jwilder it's really great to see that nginx proxy has alpine based image officially. Thank you so much

Dockerfile.alpine
matt-hh: I guess the nginx:alpine image installs some necessary libs. libc-dev? https://github.com/nginxinc/docker-nginx/blob/master/stable/alpine/Dockerfile#L55
I don't know.

PR #561, opened 2016-09-06
Dedicated nginx configuration file

rgerovski:
I moved the nginx configuration part into a dedicated file to make the configuration editing simple. The nginx.conf file already includes the directives previously set from the Dockerfile.
I find this approach of configuration more clean and useful instead of manipulating the config with sed commands.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,9 +9,8 @@ RUN apt-get update \
  && apt-get clean \
  && rm -r /var/lib/apt/lists/*
 
-# Configure Nginx and apply fix for very long server names
-RUN echo "daemon off;" >> /etc/nginx/nginx.conf \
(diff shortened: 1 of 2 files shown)


wader:
Is this different from doing:
```
FROM jwilder/nginx-proxy
COPY nginx.conf /etc/nginx/nginx.conf
```

rgerovski:
The solution you mentioned works just fine, but IMO could be very error-prone for inexperienced users as it actually replaces the already-configured nginx.conf with a custom config (which might not contain the necessary directives for the reverse proxy setup).

kamermans:
I think the idea was to track the upstream `nginx.conf`, so as the nginx team updates it, we immediately get those changes (for better or for worse).  It should be relatively safe to track the upstream config since we pin the nginx version in the `Dockerfile`, but I could certainly see the upside to having a static config in the repo.  I just added an option to the default config in my PR #589 and it was a big precarious using sed.

Issue #502, opened 2016-07-08
Unable to profixy a container started with docker-compose version 2 


JulienD:
Hi,

I wanted to implement the nginx-proxy to serve several nginx servers so I followed the example on the homepage and it works fine. However, thing start being more complex when I try to implement it with my real servers. In short the nginx-proxy is up but web servers are not proxified and I can only see a 503 status code from the proxy.

Nginx-proxy is catching events and is generating the Nginx default configuration file every time I start or stop a container.

```
dockergen.1 | 2016/07/08 13:38:53 Generated '/etc/nginx/conf.d/default.conf' from 1 containers
dockergen.1 | 2016/07/08 13:38:53 Running 'nginx -s reload'
dockergen.1 | 2016/07/08 13:38:53 Watching docker events
dockergen.1 | 2016/07/08 13:38:53 Contents of /etc/nginx/conf.d/default.conf did not change. Skipping notification 'nginx -s reload'
```

But when I start on of my container I can see the following error message.

```
dockergen.1 | 2016/07/08 13:39:27 Received event start for container 69b904a872f0
dockergen.1 | 2016/07/08 13:39:28 Received event start for container 57e7ee6e4496
dockergen.1 | 2016/07/08 13:39:28 Generated '/etc/nginx/conf.d/default.conf' from 3 containers
dockergen.1 | 2016/07/08 13:39:28 Running 'nginx -s reload'
dockergen.1 | 2016/07/08 13:39:28 Error running notify command: nginx -s reload, exit status 1
dockergen.1 | 2016/07/08 13:39:28 Contents of /etc/nginx/conf.d/default.conf did not change. Skipping notification 'nginx -s reload'
```

I had a look at the Nginx-proxy Default configuration and found that the generated file had no information in it about the web server I wanted to proxify.

```
upstream baz.bar.local {
}
server {
    server_name baz.bar.local;
    listen 80 ;
    access_log /var/log/nginx/access.log vhost;
    location / {
        proxy_pass http://baz.bar.local;
    }
}
```

Has you can see there is nothing in the upstream section 

```
upstream baz.bar.local {
}
```

I have done a lot of tests to understand what going on and finally found that problem only happens when I'm starting a container using docker-compose in version 2 because in other cases it works fine.

Here is a chunk of the Default Nginx file auto-generated

```
# If we receive X-Forwarded-Proto, pass it through; otherwise, pass along the
# scheme used to connect to this server
[...]

PR #569, opened 2016-09-14
test config before reloading [untested]

schmunk42:
⚠️ This is an untested change for now ⚠️ 
### Question/discussion:

In general, wouldn't it be better to test the config before reloading nginx.
The expected outcome would be to have nginx still running, with your old config - instead of loosing connection through reverse-proxy to all your sites.
### Related to
- https://github.com/jwilder/nginx-proxy/issues/502
- https://github.com/jwilder/nginx-proxy/issues/438

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
 nginx: nginx
-dockergen: docker-gen -watch -notify "nginx -s reload" /app/nginx.tmpl /etc/nginx/conf.d/default.conf
+dockergen: docker-gen -watch -notify "nginx -t && nginx -s reload" /app/nginx.tmpl /etc/nginx/conf.d/default.conf


schmunk42:
CC: @handcode

thomasleveil:
even with this, if you restart the nginx-container, nginx will fail to start. This just delays the problem IMHO.

I'm afraid users will then wonder why their new container is not considered by the nginx-proxy and this will end up in a support request here.

schmunk42:
@thomasleveil Yes that's correct.

But the issue was more like, if there's a broken config somehow and you try to reload, nginx will fail and **all** your virtual-hosts will be offline, i.e. it's related to https://github.com/jwilder/nginx-proxy/pull/585

I'd say to original behavior might be more desirable in development, while the patch is definitely the way to go in production.

What's about using an ENV variable for this?

schmunk42:
Any news here? Without a fix for this we end up loosing connection to all sites, just because of one broken config/setup.

PR #572, opened 2016-09-17
Replace "replace" to "trimSuffix"

pvlg:
I have a domain key-mydomain.com. When I add domain www.key-mydomain.com with ssl cert I did not get the desired result. Function replace cut name ssl cert "www.key-mydomain.com.key" to "www-mydomain.com".


jwilder:
Thanks @pvlg!

PR #574, opened 2016-09-19
Enable OCSP stapling if certificate trust chain is provided

teohhanhui:
Previously disabled in 080a5157e6b1de3637e339a7cf54105f0316cfa7

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -18,6 +18,12 @@ fi
 # Note: if $DHPARAM_BITS is not defined, generate-dhparam.sh will use 2048 as a default
 /app/generate-dhparam.sh $DHPARAM_BITS
 
+# Compute the DNS resolvers for use in the templates
+export RESOLVERS=$(awk '$1 == "nameserver" {print $2}' ORS=' ' /etc/resolv.conf | sed 's/ *$//g')
+if [ "x$RESOLVERS" = "x" ]; then
+    echo "Warning: unable to determine DNS resolvers for nginx" >&2
+fi
+
 # If the user has run the default command and the socket doesn't exist, fail
 if [ "$socketMissing" = 1 -a "$1" = forego -a "$2" = start -a "$3" = '-r' ]; then
 	exit 1
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -22,6 +22,7 @@ fi
 export RESOLVERS=$(awk '$1 == "nameserver" {print $2}' ORS=' ' /etc/resolv.conf | sed 's/ *$//g')
 if [ "x$RESOLVERS" = "x" ]; then
     echo "Warning: unable to determine DNS resolvers for nginx" >&2
+    unset RESOLVERS
 fi
 
 # If the user has run the default command and the socket doesn't exist, fail
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ and `CERT_NAME=shared` will then use this shared cert.
 The SSL cipher configuration is based on the [Mozilla nginx intermediate profile](https://wiki.mozilla.org/Security/Server_Side_TLS#Nginx) which
 should provide compatibility with clients back to Firefox 1, Chrome 1, IE 7, Opera 5, Safari 1,
 Windows XP IE8, Android 2.3, Java 7.  Note that the DES-based TLS ciphers were removed for security.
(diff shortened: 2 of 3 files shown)


teohhanhui:
Possible issue:

> For a resolution of the OCSP responder hostname, the resolver directive should also be specified.

http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_stapling

But it's quite dangerous for us to set `resolver` because it affects all domain name resolving.

kamermans:
This logic should be safe since it uses the system's resolvers if they are specified in `/etc/resolv.conf`.  If there are no `nameserver` lines, there is no DNS at all on the system, so Google's DNS servers are used.  In practice this can't normally happen in a Docker container since Docker itself falls back to Google's DNS if the host's `resolv.conf` doesn't contain accessible nameservers.

```
#!/bin/sh -e

RESOLVERS=$(awk '$1 == "nameserver" {print $2}' /etc/resolv.conf)

if [ "x$RESOLVERS" = "x" ]; then
    echo "Warning: unable to determine private DNS resolvers, using public DNS instead" >&2

    # Fallback to Google DNS servers
    RESOLVERS="8.8.4.4 8.8.8.8"
fi

echo resolve $RESOLVERS
```

To integrate this, we would need to export the `$RESOLVERS` variable to the system's environment -before- so it can be used for replacement in the `nginx.tmpl`.

teohhanhui:
@kamermans In practice name resolution works just fine, but I'm not sure if it's static - that would be quite problematic.

kamermans:
I'm not sure what you mean by static, but the resolver caches DNS lookups for the TTL of the record unless a different TTL is specified in the resolver line.

Perhaps you're concerned that without the resolver directive nginx will use the system resolver, so changes to the system DNS would be immediately active, but with resolver, it will use the specified DNS until a reload.   I can check the nginx source to see how it does name resolution by default.

Also, I see some documentation saying resolver is optional, can you confirm that it is required?

kamermans:
The current Nginx-proxy config does not have the resolver directive, and according to the 2016 documentation, without resolver, Nginx uses the name servers from resolv.conf and permenantly caches the DNS responses:
> NGINX caches the DNS records until the next restart or configuration reload, ignoring the records’ TTL values.

https://www.nginx.com/blog/dns-service-discovery-nginx-plus/

This means that everything is basically already static, so using a method like mine above to populate the resolver directive is better than the default since it only caches DNS for the TTL.

teohhanhui:
ping @jwilder

This should be ready for review.

teohhanhui:
@kamermans The build is failing :(

teohhanhui:
ping @jwilder

Can you have a look at this? Build should pass when jwilder/docker-gen#228 is merged.

teohhanhui:
Build fixed. :smile:

docker-entrypoint.sh
```diff
@@ -14,6 +14,13 @@ if [[ $DOCKER_HOST == unix://* ]]; then
 	fi
 fi
 
+# Compute the DNS resolvers for use in the templates
+export RESOLVERS=$(awk '$1 == "nameserver" {print $2}' ORS=' ' /etc/resolv.conf | sed 's/ *$//g')
+if [ "x$RESOLVERS" = "x" ]; then
+    echo "Warning: unable to determine DNS resolvers for nginx" >&2
+    unset RESOLVERS
```
teohhanhui: @kamermans This is okay, right?

kamermans: It is for now, yes, but at some point Docker said they are going to switch to a different resolver system, and that could cause issues, although I imagine they will maintain backwards compatibility.

teohhanhui:
ping @jwilder

teohhanhui:
@jwilder Is there anything blocking this? Any issues?

teohhanhui:
Rebased.

Hope someone can help add some tests...

jwilder:
Thanks @teohhanhui!

buchdag:
Why the use of the `.crt` extension for the chain file ?

Isn't `.pem` supposed to be the standard extension for concatenated certificate container files ?

This is what jrcs/letsencrypt-nginx-proxy-companion [expect nginx-proxy to use](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/blob/master/app/letsencrypt_service#L37), meaning that currently OCSP stapling doesn't work out of the box with the LE companion.

This could also be fixed on `letsencrypt-nginx-proxy-companion` side but I'd prefer the template to expect a `.pem` for the chain file instead of a `.crt`.

Should I submit a PR ?

PR #577, opened 2016-09-21
Fix README.md proxy-wide sample using wrong filename

denny0223:
Custom proxy-wide configuration file should be `/etc/nginx/proxy.conf`
https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl#L42

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,12 +239,12 @@ FROM jwilder/nginx-proxy
 RUN { \
       echo 'server_tokens off;'; \
       echo 'client_max_body_size 100m;'; \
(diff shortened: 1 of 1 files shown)


wader:
Not sure i follow. I think `/etc/nginx/conf.d/*.conf` is for adding additional config inside `http { ... }` and `/etc/nginx/proxy.conf` is for replacing the default proxy configuration in the generated `/etc/nginx/conf.d/default.conf` file.

denny0223:
Oh, my fault.

@wader you're right! :smiley:

PR #581, opened 2016-09-24
add ssl_session_tickets to default site

### LINKED ISSUE #580: ssl_session_tickets off results in an SSL handshake error
A change in commit 87879c1 `ssl_session_tickets off` causes a handshake error in latest versions of Chrome 53 and Firefox 50. I was able to establish a ssl connection in Internet Explorer, Safari and curl. I'm no expert in this area, but I did a little wireshark sniffing and found that both Chrome and Firefox closed the connection after an SSL Alert sent by the browsers

Firefox and Chrome
`192.168.1.112	   XXX.XXX.XXX.XX  TLSv1.2  Client Key Exchange, Change Cipher Spec, Hello Request, Hello Request`
`XXX.XXX.XXX.XX	   192.168.1.112	TLSv1.2	New Session Ticket, Change Cipher Spec, Encrypted Handshake Message`
`192.168.1.112	   XXX.XXX.XXX.XX	TLSv1.2   Encrypted Alert`
`TCP FIN`

Removing `ssl_session_tickets off` or setting it on works, but curiously, my other servers are functioning just fine with `ssl_session_tickets off`

Moreover `ssl_session_tickets off` is not required to get an A+ rating

chulkilee:
Fixes #580

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -70,6 +70,7 @@ server {
 	access_log /var/log/nginx/access.log vhost;
 	return 503;
 
+	ssl_session_tickets off;
 	ssl_certificate /etc/nginx/certs/default.crt;
 	ssl_certificate_key /etc/nginx/certs/default.key;
 }


chulkilee:
@jwilder although I can override the template via mount, I think this should be fixed ASAP since it breaks all HTTPS sites when using default site.

jwilder:
Thanks @chulkilee!

PR #587, opened 2016-09-29
Added X-Forwarded-Port and proxy headers tests

kamermans:
This PR adds the `X-Forwarded-Port` HTTP request header.  Just as `X-Forwarded-Proto` reflects the protocol that was used to connect to the proxy (`http` or `https`), this header reflects the port that was used to connect to the proxy (`80`, `443`, `8080`, etc).  This is important because without it there is no way for a downstream application to know what port was used by the client to connect, and therefore, it is impossible to rebuild the original hostname / URL.  Like `X-Forwarded-Proto`, if a request is received by `nginx-proxy` with an already-defined `X-Forwarded-Port`, that value is passed unaltered, but if it is not present, the header is generated.  This allows it to operate behind another instance of `nginx-proxy` or other servers that use this defacto standard, like the AWS Elastic Load Balancer, [which also uses `X-Forwarded-Port`](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html).

In order to test this new feature, I upgraded the `test_helpers::prepare_web_container()` from the stock `python -m http.server` to a custom server (which also uses `http.server`), but includes three routes:
1. `/port` Outputs "answer from port <port>" as was previously handled by a file called `data`
2. `/headers` Outputs the HTTP request headers sent by `nginx-proxy` so they can be inspected
3. `<anything else>` Outputs `No route for this path!`

Note that all methods return `HTTP 200 OK`.

I added the BATS test suite `headers` to cover the new header in this PR as well as tests for many other headers that were previously untested.

I do not expect this PR to cause any issues with existing applications as it is non-destructive to the existing http request headers.  Note, however, that other PRs that have added new tests using the `/data` endpoint to get the port will need to be updated to use `/port`.  I can change the route to `/data` and everything will "just work", but it since there is now a routing system in place, I felt a more semantic route name was in order. 

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ proxy_set_header Connection $proxy_connection;
 proxy_set_header X-Real-IP $remote_addr;
 proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 proxy_set_header X-Forwarded-Proto $proxy_x_forwarded_proto;
-proxy_set_header X-Forwarded-Port $server_port;
+proxy_set_header X-Forwarded-Port $proxy_x_forwarded_port;
 
 # Mitigate httpoxy attack (see README for details)
 proxy_set_header Proxy "";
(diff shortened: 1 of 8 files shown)


sfudeus:
Exactly the addition I stumbled upon my self today when forwarding was broken because my application received did not know the right port to use. Anything missing to be mergeable?

bjoernHeneka:
Had the same problem. Would be great if you can merge this @jwilder

jwilder:
Thanks @kamermans!

kamermans:
Sure thing, thank you for merging!

Issue #586, opened 2016-09-27
DES ciphers should be removed


hasufell:
They are weak and https://tls.imirhil.fr/https gives a very bad grade for servers providing that cipher suite.

kamermans:
I'm working on this in my `feature_ssl_improvements` branch off my fork.  I've removed the ciphers and pushed it to Docker hub as `kamermans/nginx-proxy:ssl_upgrade`.  Note that it also includes a 2048-bit `dhparam.pem` by default now.  I'll try to give it a spin with a real cert later tonight, but your confirmation would be helpful as well.

tkw1536:
DES Ciphers have been removed since this issue was filed, closing,

See [the README](https://github.com/nginx-proxy/nginx-proxy#how-ssl-support-works) on how SSL profiles work and https://github.com/nginx-proxy/nginx-proxy/blob/c4ad18fecce35da83adcca334fffd00afd75dced/nginx.tmpl#L11-L50 for the concrete set of ciphers.

PR #575, opened 2016-09-20
Remove weak DH 1024 bits FS Cipher Suites

deinspanjer:
Try testing a site hosted through nginx-proxy with the following URL for more info:
https://www.ssllabs.com/ssltest/analyze.html

Issue #562, opened 2016-09-08
Documentation update: VIRTUAL_PROTO=https != VIRTUAL_PORT=443


danbo:
Please update the documentation to state that if you select your virtual_proto to be https, the proxy doesn't automatically go to 443 but rather that it has to be specified explicitly.

jwilder:
Would you mind submitting a PR to update it?

kamermans:
How is it possible that this sat for 21 days, and both me and @r4j4h fixed it just seconds apart? :)

italomaia:
I have VIRTUAL_PROTO set to https and VIRTUAL_PORT to 443, yet, default.conf still points proxies to port 80:

```
# bin.italomaia.com
upstream bin.italomaia.com {
}
server {
	server_name app.italomaia.com;
	listen 80 ;
	access_log /var/log/nginx/access.log vhost;
	include /etc/nginx/vhost.d/default;
	location / {
		proxy_pass https://app.italomaia.com;
	}
}
```

Is that really all there is to do?

kamermans:
That's all that is needed, but please make sure you are setting these variables on your application's webserver container, not the nginx-proxy container.

tkw1536:
Seems resolved, closing.

Issue #535, opened 2016-08-14
https://www.ssllabs.com/ssltest/analyze.html - B rating per default


waffel:
It would be nice if the default nginx.conf can contain the line 

`ssl_dhparam /etc/nginx/certs/dhparam.pem;`

inside the `http {` tag

You can create the default dhparam file with 

`openssl dhparam -out dhparam.pem 2048`

The problem is, that the default length ist 1024 which does the B rating. With a 2048 pem file, a A+ rating can be obtained.

No need to do this for every proxied container

kamermans:
This sounds like a good idea, and I started to implement a pull request for it, but then it occurred to me that if we generate `dhparam.pem` in the `Dockerfile`, it will be the same for everyone on that version, which is not great for security, but if we generate it on startup, the container will take *minutes* to start, which is a show stopper.  I am not an expert on SSL, but I assume that my first observation is already true for `nginx-proxy` today - that `openssl` has already generated some random-ish data which is used by `nginx`, but it's actually the same for everyone using that version since it was computed at the time of the Docker build.

The question I have is, "is it any less secure to generate a `dhparam.pem` in the Dockerfile at build time, than what we have without a `dhparam.pem`?"

kamermans:
I've added a branch `feature_ssl_improvement` on my fork off master that contains these updates.  Note that I put the default cert in `/etc/nginx/dhparam.pem` so it doesn't get removed if a user volume-mounts `/etc/nginx/certs`.  If this was not the case, it would only be possible to volume mount that dir if you also included your own `dhparam.pem`.  If this makes sense and solves your problem I can start a PR.

waffel:
hmm .. I'am also not that SSL expert. But the idea to have a "default" pre-genrated file and then use your own from `/etc/nginx/certs` sounds like a bullet proofe concept. 

There should be a place where this is documented for the end-user (like, please replace the default generated dhparam.pem file with your own. You can use this command: ....).

I was only stumbled over this problem after checking my new https site based on this nice solution. And I found it easy to get an A+ rating with this simple "replacement". But I'am with you, that a generation or startup of minutes because of generating such file is not a good solution.

rakshazi:
Some addition problems:

```
Session resumption (caching) | No (IDs assigned but not accepted)
Strict Transport Security (HSTS) | No
```

And on handshake simulation:

```
IE 8 / XP   No FS 1	  No SNI 2 | Server sent fatal alert: handshake_failure
```

[...]

PR #589, opened 2016-09-30
SSL security enhancement

kamermans:
This PR removes the DES-based TLS ciphers from the Mozilla intermediate TLS configuration, which mitigates a number of threats and increases the SSL score on [Qualys SSL Test](https://www.ssllabs.com/ssltest/) to _A+_ and [CryptCheck](https://tls.imirhil.fr/) from an _F_ to a _B_.  Some SSL/TLS scanners, like [Symantec CryptoReport](https://cryptoreport.websecurity.symantec.com/checker/) downgrade the score because `nginx-proxy` allows TLSv1.0, but in my research I've found that the following browsers would not work if we remove TLSv1.0:
- Chrome < 22
- Firefox < 27
- IE < 11
- Safari < 7
- iOS < 5
- Android Browser < 5

These represent a non-insignificant market share as of September 2016 (according to my company's [industry-wide device usage trends report](https://www.scientiamobile.com/page/movr-mobile-overview-report)) so I recommend we leave TLSv1.0 in place at least until the end of 2017, or perhaps as late as the official end of life for PCI compliance, June 30, 2018.

Also, this PR adds a default 2048-bit Diffie-Hellman groups file, `/etc/nginx/dhparam.pem`, which is generated at build time.  This allows Perfect Forward Secrecy via DH, which is an advantageous security improvement for those clients that support it.  No browsers will be turned away because of this setting and individual vhosts can still use their own DH file as before.

[edit 2017-04-04]
The following systems do not support HTTPS connections to servers with a DH param over 1024-bits and would be unable to connect with the default 2048-DHE proposed by this PR:
- Java < 6
- Java 6 < update 105
- Java 7 < update 91
NOTE: neither Java 6u105 nor Java 7u91 was a public release, so only Java users with a paid subscription to Java SE Advanced or Java SE Support would have access to this update!

[end edit 2017-04-04]

I see some talk about OCSP Stapling as well, and as tempting as it is to add it here, I need to understand the ramifications of the additional resolver and external certificate fetching before I can endorse it.

This PR is currently available from docker hub as `kamermans/nginx-proxy:ssl_upgrade` for testing, built off today's `master` branch.

As far as I see, this PR solves the following issues and PRs:
- #586 
- #535 
- #562 
- PR #588
- PR #575


cridenour:
@kamermans Should the dhparams be built in the Dockerfile? That would mean anyone pulling this from the docker registry would get the same params - while not the end of the world, certainly isn't the ideal setup. 

Maybe something like this in docker-entrypoint.sh?

```bash
if [ ! -f /etc/nginx/dhparam.pem ]; then
    openssl dhparam -out /etc/nginx/dhparam.pem 2048
  fi
fi
```

jwilder:
@kamermans What do you think about @cridenour suggestion to generate a default DH key at startup?  I can't think of any downsides to it.

kamermans:
If I recall correctly, I did it in the Dockerfile because it took a very long time (tens of seconds or minutes) to generate.  It's not ideal to have it in the Dockerfile, but I'm not sure it's any less secure than not having it at all.  I'm not a security expert, but I would guess that knowing the dhparams in advance would only allow a malicious user to break perfect forward secrecy, but they still need to break through TLS after that.  If someone knows more than me about it, I'm all ears!

Off the top of my head, alternate ways of securing dhparams would be:
 1. Suggest the user rebuild the image.
 2. Suggest the user mount their own dhparams at `/etc/nginx/dhparam.pem`, and emit warnings on startup.
 3. Generate a less secure, but faster dhparams on startup using less than 2048 bits (not sure if this will pass the SSL/TLS checks).
 4. Generate the dhparams in the background and once it's finished, reload nginx (might be an interesting option).

dehy:
As the dhparams need to be generated once, I think the build time at startup is acceptable, if in a volume for surviving future redeployment of service. But I like the #4 as well!

kamermans:
In my environment, I could not handle a potentially 1-minute startup as I almost always `--force-recreate` my images to get rid of transient data.  Let me think about option 4 tonight...

dehy:
So in this thread : https://github.com/jchaney/owncloud/issues/42#issuecomment-182110124

> I choose option 2. This is what most people seem to be doing, and what most ssl configuring guides and stackexchange answers point to. And because it takes away the fear of an attack like logjam, where a part of the impact came from the fact that lots of sites where using the same primes.

I agree with him. But he also note that generating the primes in the build could also be a viable option. So that's your call :D

kamermans:
Ok, so here are my findings.  I've generated the `2048` bit prime `dhparams` on a variety of machines to see how long they take, there does seem to be a non-linear time component to it:

```
time openssl dhparam -out /tmp/dhparam.pem 2048

Bare-Metal: Intel(R) Core(TM) i7-5960X CPU @ 3.00GHz
0:13        

AWS EC2 t2.micro: Intel(R) Xeon(R) CPU E5-2670 v2 @ 2.50GHz
1:18

Bare-Metal: Intel(R) Xeon(R) CPU E5-2620 0 @ 2.00GHz
2:26

Rackspace Cloud: Quad-Core AMD Opteron(tm) Processor 2374 HE
6:28

Bare-Metal: Intel(R) Atom(TM) CPU  230   @ 1.60GHz
12:00
```

This literally means that if we add it to the `entrypoint` and start the container on my production Rackspace Cloud instance (which is an HTTP load balancer, btw), it will add *six and a half minutes* to the startup time.  Imagine you update the image and run `docker-compose up --force-recreate` to get the new features, and it takes nearly seven minutes to start :-1:

kamermans:
@jwilder: I spent a bunch of time thinking this through and testing things.  The `dhparam` generation time is not predictable.  On my fastest machine, I usually get times like 20 seconds, but sometimes it will take nearly two minutes.  It's really just a matter of chance since `openssl` is looking for large prime numbers arbitrarily.  I think it's conceivable that it could take over 10 minutes on a modern cloud instance (it took 12 min on my old Rackspace instance and 20 min on my Atom 1.6GHz CPU).

As annoying as this generation time is (it's a show-stopper for me), it is important for security reasons, so I went ahead and implemented my 4th idea .  This branch (`kamermans/nginx-proxy:feature_ssl_improvement`) now ships with `/app/dhparam.pem.default`, which I generated on my computer.  On startup, if `/etc/nginx/dhparam/dhparam.pem` is not present, or has the same `md5sum` as `/app/dhparam.pem.default`, a new `2048` bit dhparam is generated in the background with `nice -n +5`, which decreases its CPU priority, since without it, it will peg a CPU core the whole time.  After the generation is complete, nginx is reloaded via `nginx -s reload`.  The process is kicked-off by `docker-entrypoint.sh` (I tried `forego/Procfile`, but I don't see how to run a one-time script).  Since the entrypoint could be called again while the background generation is running, the system uses a lockfile to prevent multiple background generations from running in parallel (and pegging all CPU cores).

I think the whole system should be pretty safe - I've already run it in my environment without issue.

In case someone else wants to give it a spin, I've uploaded it to Docker hub as `kamermans/nginx-proxy:dhparam_bg_init`.  When you first start it, you will see these comments:
```
WARNING: /etc/nginx/dhparam/dhparam.pem was not found. A pregenerated dhparam.pem will be used for now while a new one
... normal nginx-proxy init stuff ...
Generating DH parameters, 2048 bit long safe prime, generator 2
This is going to take a long time
dhparam generation complete, reloading nginx
```

Note that there's really no reason we couldn't increase from `2048` bits to `4096` if there is some advantage there, but we could be literally seeing things like hour-long generation times for very slow machines :P

Also, I see that the alpine variant is failing the BATS tests, and I'm not sure why yet, but it's hard to imagine it's related to the dhparam stuff.  I'll take a look at it anyway.

kamermans:
I'm not sure why these BATS tests are failing in Travis-CI - it's like the `openssl` command is not running at all or something.  I'll get to the bottom of it.

kamermans:
Ok, the Travis-CI environment was not fast enough to generate a `2048` bit dhparam, so I've added an env var `DHPARAM_BITS` which defaults to `2048`, but I set it to `256` for the BATS tests and now it's passing.

kamermans:
@jwilder This PR is stable now and I've fixed the merge conflict with a BATS test in `master`.  I think it's ready to be merged if you want.

jwilder:
This will need a rebase now that #655 is merged.

kamermans:
@jwilder: Rebase complete.  I've put a small hack in place to workaround the Docker version issue, but it should merge fine with #757.

@thomasleveil: please note that I switched the python base image to `python-alpine` so I could get `openssl >= 1.0.2`.  This makes it possible for me to confirm that the `dhparams.pem` file is being used by `nginx` in a unit test.  All tests are passing, so I assume the alpine version is ok.  I'm not a python developer, so perhaps you could take a quick look at `test/test_ssl/test_dhparam*.py` to see if I am doing anything crazy :)

test/test_ssl/test_dhparam.py
```diff
@@ -0,0 +1,74 @@
...
+    containers = docker_client.containers.list(filters={"ancestor": "jwilder/nginx-proxy:test"})
+    if len(containers) != 1:
+        return
+    container = containers[0]
+    for line in container.logs(stream=True):
+        if "Watching docker events" in line:
+            break
+
+def test_dhparam_is_not_generated_if_present(docker_compose, nginxproxy):
+    wait_for_nginxproxy_to_be_ready()
```
thomasleveil: This isn't needed as the `docker_compose` fixture calls it already

test/test_ssl/test_dhparam.yml
```diff
@@ -0,0 +1,16 @@
+web5:
+  image: web
+  expose:
+    - "85"
+  environment:
+    WEB_PORTS: "85"
+    VIRTUAL_HOST: "web5.nginx-proxy.tld"
+    HTTPS_METHOD: nohttp
```
thomasleveil: `HTTPS_METHOD` does not seem to have any effect on the outcome of the test. I would remove it

thomasleveil:
@kamermans I've submitted to you kamermans/nginx-proxy/pull/2 with changes that IMHO helps understanding the intents of your tests

kamermans:
Very nice, thanks!

test/test_ssl/test_dhparam.py
kamermans: Ah, so the container is already ready before the tests are run?

test/test_ssl/test_dhparam.yml
kamermans: Good point, no need for it here - it was an artifact of me copying another test :)

test/test_ssl/test_dhparam.py
thomasleveil: Exactly

thomasleveil: Exactly

kamermans:
Please note that at my organization I have run into issues with 2048-bit DH primes.  Java 6 only supported up to 1024-bit DH primes for quite a while, and I have customers that require support for these older versions of Java, so at least in my organization, I have had to move back to 1024-bit.

jwilder:
@kamermans I'd lean towards more secure by default so leave the 2048 as default as is.  Did you have any other changes you needed to make before merging?  Looks good to me though.

kamermans:
Ok, sounds good to me @jwilder, I have no more changes.

kamermans:
@jwilder the merge from `master` caused the tests to fail because `server_names_hash_bucket_size` is now defined in `nginx.tmpl` instead of via `sed` in the `Dockerfile`.  I've fixed that issue and added a blurb to the `README.md` about the DH param compatibility with older clients, as well as a workaround that can be used with these clients (passing `DHPARAM_BITS=1024` to the `nginx-proxy` container).

jwilder:
Hmm.  Looks like something on travis or elsewhere is failing now.  Another PR failed for the same reason.

```
1.19s$ sudo apt-get -o Dpkg::Options::="--force-confnew" install -y --force-yes docker-engine=${DOCKER_VERSION}
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Version '1.13.1-0~ubuntu-trusty' for 'docker-engine' was not found
The command "sudo apt-get -o Dpkg::Options::="--force-confnew" install -y --force-yes docker-engine=${DOCKER_VERSION}" failed and exited with 100 during .
```

jwilder:
Build is fixed in master.  Thanks @kamermans!

Issue #573, opened 2016-09-17
Subdomain without ssl certificate when domain ssl certificate


pvlg:
I have domain mydomain.com with ssl certificate. I want to use the domain sub.mydomain.com without a certificate.

How can I do it?

I replaced the row

`
{{ $vhostCert := (closest (dir "/etc/nginx/certs") (printf "%s.crt" $host))}}
`

to

`
{{ $vhostCert := printf "%s.crt" $host}}
`

I think it would be a good idea to add a options to the configuration, but do not know how best to do it.

wader:
Hi, so you want SSL for `mydomain.com` but not for `sub.mydomain.com`? I think you can do kind of a hack by settings `CERT_NAME=` (empty string) for the `sub.mydomain.com ` container.

kamermans:
We have `HTTPS_METHOD=nohttp` to disable the HTTP site entirely, I could add `HTTPS_METHOD=nohttps` to disable HTTPS if this is a common need.

wader:
@kamermans yes i think that would be nice

tkw1536:
Seems resolved by the above, closing

PR #590, opened 2016-10-01
Added HTTPS_METHOD=nohttps to disable SSL site

kamermans:
This feature complements `HTTPS_METHOD=nohttp`, which disables the HTTP site.  For containers with `HTTPS_METHOD=nohttps` set, the template sets `is_https` to `false`, which treats everything as if there was no matching certificate.  

This solves issue #573, in which a user needs to disable the HTTPS site for a specific container with a subdomain that happens to match the SSL cert of another container.

A test has also been added to cover this directive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,12 +178,13 @@ a 503.
 
 To serve traffic in both SSL and non-SSL modes without redirecting to SSL, you can include the
 environment variable `HTTPS_METHOD=noredirect` (the default is `HTTPS_METHOD=redirect`).  You can also
-disable the non-SSL site entirely with `HTTPS_METHOD=nohttp`. `HTTPS_METHOD` must be specified
-on each container for which you want to override the default behavior.  If `HTTPS_METHOD=noredirect` is
-used, Strict Transport Security (HSTS) is disabled to prevent HTTPS users from being redirected by the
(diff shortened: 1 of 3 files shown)


Lange:
I have this use case as well, and would love to see this PR merged!

kachkaev:
@Lange for now you can use this trick: `CERT_NAME=` ([suggested by @wader](https://github.com/jwilder/nginx-proxy/issues/573#issuecomment-249412689)). I struggled with the same problem just now and this worked for me like a charm!

kamermans:
@jwilder This build is failing because Travis-CI is running a newer version of Docker than the one specified in your `.travis.yml`.  We will need to do one of the following:

1. Add `--force-yes` to the `apt-get` command to force Travis to downgrade Docker to the version specified (currently `1.12.1-0~trusty`)
2. Change the required Docker version to a version that is available on Travis-CI without downgrading
3. Be less specific and just accept whatever version of Docker is available in the testing environment

I think it would be best to do 1 and 2, where we figure out the latest version in Travis, pin that one in `.travis.yml` and add the `--force-yes` flag to prevent this from happening in the future.

kamermans:
Sure, I'll rebase and push today.

jwilder:
Thanks @kamermans!

PR #605, opened 2016-10-13
Add explanation in README on how to create a Basic Auth password file

bryant1410:
It's somehow similar to the explanation in [Native basic auth in Docker Registry deployment docs](https://docs.docker.com/registry/deploying/#/native-basic-auth).

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,11 @@ $ docker run -d -p 80:80 -p 443:443 \
     jwilder/nginx-proxy
 ```
 
-You'll need apache2-utils on the machine where you plan to create the htpasswd file. Follow these [instructions](http://httpd.apache.org/docs/2.2/programs/htpasswd.html)
+You'll need to create a password file for each virtualhost:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -198,7 +198,11 @@ $ docker run -d -p 80:80 -p 443:443 \
     jwilder/nginx-proxy
 ```
 
-You'll need apache2-utils on the machine where you plan to create the htpasswd file. Follow these [instructions](http://httpd.apache.org/docs/2.2/programs/htpasswd.html)
+You'll need to create a password file for each virtualhost:
+
+```shell
+docker run --rm -ti m31271n/htpasswd <username> <password> > htpasswd
```
jwilder: Can you change `htpasswd` to `$VIRTUAL_HOST` so it's a little more clear that the password file needs to be named as the hostname.

buchdag:
The Docker image used in the instruction no longer exist.

PR #607, opened 2016-10-15
Add options to listen on custom port

nickygerritsen:
This PR is an improved version of (part of) #472 .

It adds the variables `VIRTUAL_LISTEN_PORT_HTTP` and `VIRTUAL_LISTEN_PORT_HTTPS` to containers. When a container is running with such a variable, the `nginx-proxy` container will create the virtual host for that given port instead of the default 80/443.

This is useful if your Docker host has multiple external network interfaces and you want some sites to be available on one IP, while others are available on the other. You can then map one port to one IP and another port to the other IP.

Some notes:
- The `_` host will always be created for port 80 (443) AND all `VIRTUAL_LISTEN_PORT_HTTP` (`VIRTUAL_LISTEN_PORT_HTTPS`) defined in any container.
- The default server (`DEFAULT_HOST`) will also listen on all ports. If this is something that is not desired, I can remove this part again, making `_` the default server again for all ports except the one the `DEFAULT_HOST` listens on.
- I changed / added quite some tests to make sure this all works.

If this is something you are willing to merge, I can add an extra commit to update the readme with this information. I haven't done this yet because I'm not sure how you think about this feature. Just let me know! We are going to use this version anyway, but I think it might be useful for other people as well.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -56,9 +56,17 @@ proxy_set_header X-Forwarded-Proto $proxy_x_forwarded_proto;
 proxy_set_header Proxy "";
 {{ end }}
 
+{{ $http_ports := (groupByKeys $ "Env.VIRTUAL_LISTEN_PORT_HTTP") }}
+{{ $https_ports := (groupByKeys $ "Env.VIRTUAL_LISTEN_PORT_HTTPS") }}
+
 server {
 	server_name _; # This is just an invalid value which will never trigger on a real hostname.
 	listen 80;
+	{{ range $port := $http_ports }}
+		{{ if ne $port "80" }}
+			listen {{ $port }};
+		{{ end }}
+	{{ end }}
 	access_log /var/log/nginx/access.log vhost;
 	return 503;
 }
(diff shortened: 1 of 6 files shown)


JonathanTab:
Can we make this happen, or is there some alternative route for doing this?

edemaine:
This PR would also resolve #921. Is there interest in adding this feature to upstream? Anything that needs to be done (beyond resolving conflicts)?

rakshazi:
Hello!
Any update on this? Really must-have feature!

nginx.tmpl
```diff
@@ -132,15 +151,33 @@ upstream {{ $host }} {
...
 	return 301 https://$host$request_uri;
 }
 {{ end }}
 
 server {
 	server_name {{ $host }};
-	listen 443 ssl http2 {{ $default_server }};
+	{{ if eq $default_server 1 }}
+		listen 443 ssl http2 default_server;
+		{{ range $port := $http_ports }}
```
alextov: Shouldn't this read `$https_ports`?

mdeboer:
Can this be merged please?

PR #608, opened 2016-10-17
set ssl_session_tickets to off by default

MelwinKfr:
Necessary when other server{} blocks are defined in `/etc/nginx/conf.d/*.conf` without ssl_session_tickets set to off.
See https://community.letsencrypt.org/t/errors-from-browsers-with-ssl-session-tickets-off-nginx/18124/5?u=melwinkfr
Related to #580

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -63,6 +63,9 @@ server {
 	return 503;
 }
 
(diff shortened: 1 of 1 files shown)


jwilder:
Already implemented via #581

PR #614, opened 2016-10-26
Do not reply to SSL requests with invalid hostnames

colthreepv:
If a client requests a domain with an invalid (or not present) hostname, nginx-proxy replies with a `503`, when in HTTP.
But when in HTTPS, it will just give you the first hostname in the configuration.

I find this unexpected for the client; it that might have mistyped the url and receives another different website as response. Also as a security concern against a specific type of attack, [Stackexchange discussion](http://security.stackexchange.com/q/55790/69147)

This PR is at protototype stage, there are no tests, and the response is 444 (that in nginx language means 'close the connection')

It's an acceptable PR? I would gladly learn how to make tests in bats, and update the PR

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -109,6 +109,10 @@ upstream {{ $host }} {
 {{/* Get the VIRTUAL_PROTO defined by containers w/ the same vhost, falling back to "http" */}}
 {{ $proto := or (first (groupByKeys $containers "Env.VIRTUAL_PROTO")) "http" }}
 
+{{/* Prepare hostnames to be converted in regex */}}
+{{ $http_host := replace $host "." "\\." -1 }}
+{{ $http_host := replace $http_host "," "|" -1 }}
+
 {{/* Get the HTTPS_METHOD defined by containers w/ the same vhost, falling back to "redirect" */}}
 {{ $https_method := or (first (groupByKeys $containers "Env.HTTPS_METHOD")) "redirect" }}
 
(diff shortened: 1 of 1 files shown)


toussa:
isn't it better to use the same mechanism for non-SSL handling in nginx.tmpl ? Meaning without if (see: [if is evil](https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/)  ) and returning 503 to be consistent ?

colthreepv:
returning 503 means still replying to the response, "replying 444" means forcefully closing the connection - I think the latter is less time consuming for the server, but I am not an expert to such extent.

Regarding the "if is evil" I think it has to be taken into perspective, there are cases in which if(s) in nginx are perfectly safe, and there are no other ways to do things

There are 2 kinds of "wrong hostname queries":
1 - legit users typing wrong hostname, in that case I don't want the server to reply, as this hostname is not-existant! If I reply 503 the user might think there's a problem with the server (temporary), and the hostname is valid.
2 - malicious users that try to exploit this hostname incongruency to force the server to reply

In both cases I see "drop the connection ASAP" a better approach - but this is not the point of the PR

The point of the PR is to have a consistent behaviour for connections established with HTTP and HTTPS

maritz:
This is still an issue as far as I can tell. It should be noted, that the TravisCI failure looks like a temporary apt issue. Maybe someone could retrigger that?

matheuscmpm:
Nothing new about this PR/issue? I still have this issue, and it is quite troublesome.

nginx.tmpl
```diff
@@ -109,6 +109,10 @@ upstream {{ $host }} {
 {{/* Get the VIRTUAL_PROTO defined by containers w/ the same vhost, falling back to "http" */}}
 {{ $proto := or (first (groupByKeys $containers "Env.VIRTUAL_PROTO")) "http" }}
 
+{{/* Prepare hostnames to be converted in regex */}}
+{{ $http_host := replace $host "." "\\." -1 }}
+{{ $http_host := replace $http_host "," "|" -1 }}
```
MartijnHols: This breaks wildcard virtual hosts. Adding this line seems to fix it:

```
{{ $http_host := replace $http_host "*" ".+" -1 }}
```

buchdag:
This was implemented in #2186

PR #618, opened 2016-10-28
Support for upstream template config

alirezaDavid:
for upstream config we needed add keepalive directive to template.
I make it dynamically from template.
If you add default file to upstream folder you can add upstream directives.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -47,7 +47,7 @@ proxy_http_version 1.1;
 proxy_buffering off;
 proxy_set_header Host $http_host;
 proxy_set_header Upgrade $http_upgrade;
-proxy_set_header Connection $proxy_connection;
+proxy_set_header Connection "";
 proxy_set_header X-Real-IP $remote_addr;
 proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 proxy_set_header X-Forwarded-Proto $proxy_x_forwarded_proto;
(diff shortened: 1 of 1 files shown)


wader:
Hi, can you give some more details on why to remove the connection header? I think that will break proxying of websockets.

alirezaDavid:
Hi, Ya you'r right. My bad :)
If need be we can add it in server default.
I update it.

thomasleveil:
The current convention for customizing the [per-VIRTUAL_HOST default configuration](https://github.com/jwilder/nginx-proxy#per-virtual_host-default-configuration) is to provide the `/etc/nginx/vhost.d/default` file. Wouldn't it be more coherent to have the _per-VIRTUAL_HOST default upstream_ file be `/etc/nginx/vhost.d/default_upstream`?

Also, some might want the upstream customization to apply to one VHOST only, following the same conventions as the [per-VIRTUAL_HOST location configuration](https://github.com/jwilder/nginx-proxy#per-virtual_host-location-configuration), files in `/etc/nginx/vhost.d/` named after the VIRTUAL_HOST with suffix `_upstream` could be used. How does this sound?

BirkhoffLee:
+1 for @thomasleveil, its a must-have

desimaniac:
@thomasleveil Any possibility of that being added in ? 😛

thomasleveil:
I do not have the time to contribute to this project anymore (and have not in years). This won't come from me sorry `¯\_(ツ)_/¯`

PR #649, opened 2016-12-05
Travis-CI - upgrade docker-engine and allow downgrades

kamermans:
This PR upgrades `docker-engine` in Travis-CI from `1.12.1` to `1.12.3` and adds the `--allow-downgrades` flag.  Without this flag, when `1.12.4` comes out, `apt-get` will fail with an error because it does not allow automated downgrades of packages without confirmation, so all builds will start to fail (this just happened in #590 ).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
   # list docker-engine versions
   - apt-cache madison docker-engine
   # upgrade docker-engine to specific version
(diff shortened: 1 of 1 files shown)


kamermans:
Debian deprecated `--force-yes` in favor of `--allow-*` (like `--allow-downgrades`), but in Ubuntu 14.04, apparently the new flags are not supported yet, so I am forced to use the deprecated `--force-yes` option to force the installation of a potentially older version of Docker.  This should work for at least a few years, until Travis is using Ubuntu > 16.04.

jwilder:
Thanks @kamermans!

PR #653, opened 2016-12-08
Updated nginx to 1.11.6

cramaker:
I ran into the following but that was fixed in 1.11.5 - "nginx returned the 400 response on requests with the "-" character in the HTTP method." (https://forum.nginx.org/read.php?27,270178)

I've bumped the version to the latest from the official Nginx container.


jwilder:
Thanks @cramaker!

PR #655, opened 2016-12-10
rewrite tests using docker-compose and pytest

thomasleveil:
Following up #197. 

I never was quite satisfied with Bats and more precisely having to write tests in Bash. While the current test suite serves a purpose it has not evolved as it could have. IMHO, Bash is hard to read and when a test case fails, it is time consuming to reproduce/understand/fix the issue. Those barriers might have dissuade potential contributors.

IIRC @jwilder originally expressed the idea of writing test cases with the help of docker-compose. That's what I started to work on.

I kept the actual test suite and added the new one under the `test2` directory. The Travis config file is setup so that it will run both tests suites, but only legacy tests can mark the build as broken. This is to allow a transition period that should prove that a failing test case under `test` matches a failing test case under `test2`.

To sum up the objectives:

- [x] make tests easier to read/write
- [x] make it easier to understand what's going on with a failing test case
- [x] no regression : a failing test case with legacy tests should match a failing test case with the new test suite
- [x] in the end, get rid of Bats completely
- [x] port all the actual tests
    - [x] default-host.bats
    - [x] docker.bats
    - [x] headers.bats
    - [x] multiple-hosts.bats
    - [x] multiple-ports.bats
    - [x] ssl.bats
    - [x] wildcard-hosts.bats


jwilder:
@thomasleveil This looks a lot easier to understand then the existing bats test. The compose files also provide some documentation of sample usage.

thomasleveil:
@jwilder this PR is now complete. I would appreciate  if you could try it on your machine and confirm there is no trouble to run this new test suite.

jwilder:
Is there a way to override the DNS resolver when running the tests to allow a different IP to to be used for the `nginx-proxy` container?  On a mac, I see:
```bash
$ pytest -s
...
INFO:patched DNS:resolving domain name 'nginx-proxy' as IP address is 172.17.0.2
```
but `172.17.0.2` isn't routeable from my host so the tests just hang.

thomasleveil:
are you using [docker for Mac](https://docs.docker.com/engine/installation/mac/#docker-for-mac) or [Docker Toolbox](https://docs.docker.com/engine/installation/mac/#docker-toolbox)?

I don't own a Mac myself so I can't provide a solution that would work around the network specificities of Docker on Mac.

Some pointers thought for workaround solutions:

Maybe we can find a way to run pytest within a docker container:

Dockerfile-nginx-proxy-tester:

    FROM python:2.7
    COPY requirements.txt /requirements.txt
    RUN pip install -r /requirements.txt
    WORKDIR /test
    ENTRYPOINT ["pytest"]

then build:

    docker build -t nginx-proxy-tester -f Dockerfile-nginx-proxy-tester .
    
and run:    

    docker run -v $(pwd):/test -v /var/run/docker.sock:/var/run/docker.sock --rm nginx-proxy-tester

The issue with this is that for every test we start by removing all containers (including that one)... you would need to find a way to prevent that in https://github.com/thomasleveil/nginx-proxy/blob/pytest/test2/conftest.py#L113. 

----

Or playing with dind (Docker in Docker): https://forums.docker.com/t/how-to-run-docker-inside-a-container-running-on-docker-for-mac/16268/3

thomasleveil:
I worked on the _running pytest inside a container_ idea and it almost work. It fails in cases where the docker-compose file uses volumes ; so when it does solve the network issue you have on mac, it does raises a new one --> NOGO

I will give a try to the _dind_ option.

@jwilder are you using docker for Mac or Docker Toolbox?

thomasleveil:
I managed to get the test suite running from a docker container. just call the `test2/nginx-proxy-tester.sh` script and let me know.

thomasleveil:
There is still an issue with the `./nginx-proxy-tester.sh` script when a test is using docker compose syntax v2. (for instance when running test_composev2.py). 

you can skip this particular test by running:

    ./nginx-proxy-tester.sh --ignore test_composev2.py

thomasleveil:
The `./nginx-proxy-tester.sh` script is now able to handle tests using docker compose files written syntax v2.

Mac users should now be able to run the full test suite with it.

thomasleveil:
@kamermans since you contributed to the nginx-proxy test suite in the recent past, I would appreciate your feedback on this PR ;)

The related branch to checkout is https://github.com/thomasleveil/nginx-proxy/tree/pytest

kamermans:
Hi @thomasleveil, I did look over the PR last month and it made sense to me.   That having been said, I've also written a number of tests since then, particularly on the SSL side, so I'd like to see how they are implemented in Python.   I do think that the testing will be easier in Python than BATS, despite my soft-spot for Bash :). 

I've got a red eye from Silicon Valley to Washington DC tonight. If I can't sleep, I'll try to give your PR a spin on the plane.

kamermans:
Hi @thomasleveil, I've tried it out and it works pretty well.  There were some bugs that I fixed, for example, you cannot create files with a `*` in the filename on Windows, so I was unable to clone the project at all.  As per the docs, for wildcard domains, like `*.nginx-proxy.tld`, `nginx-proxy` will load the cert `nginx-proxy.tld`, if present, so I renamed the `*.nginx-proxy.tld` files to `nginx-proxy.tld` and updated the `yml` files that referenced them.

I've fixed these issues in my fork of your fork (sounds like Inception :) ):
https://github.com/kamermans/nginx-proxy/tree/steve_pytest

To keep things in one place, I issued my PR against your fork (https://github.com/thomasleveil/nginx-proxy/pull/4), so if you pull it my commits will end up on this PR as well, and everything should be clean here.

thomasleveil:
@kamermans thank you for helping out on this PR, much appreciated 👍

kamermans:
Sure thing!  I'm pretty sure that's my first PR from 10,000 feet off the ground at 600 MPH :)

jwilder:
@thomasleveil I haven't forgotten about this.  Have not had a chance to try it out yet.

jwilder:
@thomasleveil This PR is awesome!  It works great locally for me.  If all the bats tests have been ported over, I think we can remove them as well.

jwilder:
@thomasleveil Let me know if you are ready for me to merge this.

kamermans:
I agree that we should remove the bats tests.  Most of the tests have been written or improved by the three of us anyway, and I don't mind chipping in if there are other open PRs with tests that need to be ported, assuming those PRs are approved.

thomasleveil:
@jwilder all bats tests have been ported and there are even additional tests covering even more features. Previous test suite is gone.

Note that on Travis-CI, IPv6 related tests are skipped since the Travis-ci boxes [don't support IPv6](https://docs.travis-ci.com/user/ci-environment/#Networking)

This PR is ready to be merged. If you which, I can squash all those commits together before you do. Let me know.

jwilder:
Thanks @thomasleveil!

PR #668, opened 2017-01-04
Increase limits for size of files that can be passed

dcornejo:
I am using nginx-proxy to front for a series of Wordpress containers. Some of the content and themes on those sites exceeds the default limits for nginx - this change increases the size of what can be uploaded.


jwilder:
You can do this already by mounting in a custom proxy config or using a derived images.  See: https://github.com/jwilder/nginx-proxy#proxy-wide

Issue #677, opened 2017-01-12
Gitlab crashed because of proxy_pass


RaphaelJenni:
I updated the nginx-proxy and suddenly the url of my gitlab was changed to 
https://6ad65ad95ad9f5ad9f5adfa131a65d16f1a

This is set in the conf.d/default.conf under each server in the location section.
Please make that optional or find a different way to solve that problem.

jwilder:
Likely #671 caused this. cc @thomasleveil 

Can you provide the conf file it's generating?  I assumed this change was basically an internal change and you would not see it in the backends or anywhere else.

RaphaelJenni:
This is my `default.conf` file. I hope it helps.


[default.conf.txt](https://github.com/jwilder/nginx-proxy/files/703720/default.conf.txt)

thomasleveil:
Hi could you share the contents of:
/etc/nginx/vhost.d/jenkins.domain1.tdl
And
/etc/nginx/vhost.d/default

Please?

Le ven. 13 janv. 2017 à 08:14, RaphaelJenni <notifications@github.com> a
écrit :

This is my default.conf file. I hope it helps.

default.conf.txt
<https://github.com/jwilder/nginx-proxy/files/703720/default.conf.txt>

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/jwilder/nginx-proxy/issues/677#issuecomment-272376879>,
or mute the thread

RaphaelJenni:
**/etc/nginx/vhost.d/default**
```
## Start of configuration add by letsencrypt container
location /.well-known/acme-challenge/ {
    allow all;
    root /usr/share/nginx/html;
    try_files $uri =404;
    break;
}
## End of configuration add by letsencrypt container
proxy_http_version 1.1;
proxy_set_header Connection "";

client_max_body_size 1g;
```


**/etc/nginx/vhost.d/jenkins.domain1.tdl**
```
## Start of configuration add by letsencrypt container
location /.well-known/acme-challenge/ {
    allow all;
    root /usr/share/nginx/html;
    try_files $uri =404;
    break;
}
## End of configuration add by letsencrypt container
proxy_set_header        Host $host;
proxy_set_header        X-Real-IP $remote_addr;
proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_set_header        X-Forwarded-Proto $scheme;

```

cncolder:
I have same issue.

[...]

PR #671, opened 2017-01-07
fix crash when using end-of-string symbol `$` in regex VIRTUAL_HOST

thomasleveil:
This PR adds tests to show a case where the nginx configuration file produced is not valid and makes nginx fail to reload.

----

`$ bats wildcard-hosts.bats`

    ✓ [wildcard-hosts] start a nginx-proxy container
    ✓ [wildcard-hosts] VIRTUAL_HOST=*.wildcard.bats
    ✓ [wildcard-hosts] VIRTUAL_HOST=wildcard.bats.*
    ✓ [wildcard-hosts] VIRTUAL_HOST=~^foo\.bar\..*\.bats
    ✗ [wildcard-hosts] VIRTUAL_HOST=~^foo\.bar\..*\.bats$
      (from function `assert_output' in file lib/bats/batslib.bash, line 328,
       from function `assert_200' in file wildcard-hosts.bats, line 83,
       in test file wildcard-hosts.bats, line 65)
        `assert_200 foo.bar.whatever.bats' failed
        HTTP/1.1 503 Service Temporarily Unavailable
      ab2531a8a276
      container_name: bats-wildcard-hosts-4
      ports: 80
      options: -e VIRTUAL_HOST=~^foo.bar..*.bats$
      expose_option: --expose=80

      -- line differs --
      index    : 0
      expected : HTTP/1.1 200 OK
      actual   : HTTP/1.1 503 Service Temporarily Unavailable
      --

    ✓ [wildcard-hosts] stop all bats containers

     tests, 1 failure

----

```
$ docker start bats-nginx-proxy-wildcard-hosts
$ docker start bats-wildcard-hosts-4
$ docker logs bats-nginx-proxy-wildcard-hosts
```

    dockergen.1 | 2017/01/07 22:48:10 Received event start for container 4aba463b7021
    dockergen.1 | 2017/01/07 22:48:11 Generated '/etc/nginx/conf.d/default.conf' from 2 containers
    dockergen.1 | 2017/01/07 22:48:11 Running 'nginx -s reload'
    dockergen.1 | 2017/01/07 22:48:11 Error running notify command: nginx -s reload, exit status 1

----

`$ docker exec bats-nginx-proxy-wildcard-hosts cat /etc/nginx/conf.d/default.conf`

    # If we receive X-Forwarded-Proto, pass it through; otherwise, pass along the
    # scheme used to connect to this server
    map $http_x_forwarded_proto $proxy_x_forwarded_proto {
      default $http_x_forwarded_proto;
[...]

PR #679, opened 2017-01-14
regexp: use sha1 for upstream only if regexp is used

thomasleveil:
Change a bit the new way to name upstream introduced in #671. With this PR, the new way to name upstreams is only triggered if matching hostname with a regular expression. 

The intent is to reduce changes to existing users who might have added custom conf to the vhost.d directory as discovered in #677


jwilder:
Thanks @thomasleveil!

PR #684, opened 2017-01-20
Support for VIRTUAL_PATH and some additional restructuring

DrMurx:
This PR contains a bigger change:

First of all, it adds support for virtual paths to allow backends being connected to a specific path (`VIRTUAL_PATH`) on their `VIRTUAL_HOST`.

The semantic of the `VIRTUAL_PORT` variable has been changed to be more consistent with `VIRTUAL_HOST` and `VIRTUAL_PATH`. Instead of specifying the backend container's port, it will now allow to set a non-default port on the frontend. Depending on `HTTPS_METHOD`, the configuration will properly support HTTPS also on a non-standard port.

To specify an alternative backend port, now use `VIRTUAL_BACKEND_PORT` instead of `VIRTUAL_PORT`. In the same way `VIRTUAL_BACKEND_PROTO` replaces `VIRTUAL_PROTO`.

The inclusion of additional files (such as `auth_basic_user_file` and the `location` include) has been extended to allow fine grained control, now either per virtual host+port+path, host+port, host or globally.

As this change requires two additional steps of aggregation and iteration on the container list, I've also completely restructured the template into a couple of `defines` which have a clearer separation of concerns. The indenting now reflects the structure of the template, not only the structure of the nginx configuration. I believe that this eases up understanding.

I've yet not written any tests - mainly because I was unsure whether you'd like to add such a huge rework of the template (with breaking changes) to the main line. Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,21 @@ $ curl -H "Host: whoami.local" localhost
 I'm 5b129ab83266
 ```
 
+### Virtual Route Support
+
+If you want to reach your container not at the URL root of VIRTUAL_HOST, but rather at a specific path, just set the environment variable `VIRTUAL_PATH=/path`.
+
+By default, `/path` is removed from the forwarded requests. However, if your application requires the full URL (for example because it creates absolute URLs for assets), you might also want to set `VIRTUAL_PATH_PASS_THROUGH=true`.
+
+### Virtual Port Support
+
+You can tell nginx-proxy to publish the backend container on a non-standard port, different from 80 and 443. Just specify the environment variable `VIRTUAL_PORT` on the backend container. Depending on the variable `HTTPS_METHOD` (see below), nginx-proxy will be configured to support HTTPS on the non-standard port.
+
+Currently, this feature requires to run the nginx-proxy container with these ports already published as Docker doesn't yet provide a method to publish and unpublish ports on the fly.
+
 ### Multiple Ports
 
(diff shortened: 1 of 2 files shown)


Vad1mo:
I really would like to see such a functionality becoming official. 

However to make life easier for @jwilder we need change the PR to make it compliant 

1. Backward compatibility. 
2. One new future per PR it will become easier to test and merge.
3. Make the test and build pass. 

But before doing all this effort wen need feedback from @jwilder  if he would like to see this feature in mainline

jwilder:
Agree w/ @Vad1mo.  I'm closing this PR because it changes too much and is not backwards compatible.  #254 is an older `VIRTUAL_PATH` PR that has a lot of discussion on the feature.  I'd rather resurrect that to get `VIRTUAL_PATH` support working.

DrMurx:
I completely agree on too many feature changes in a single PR, and I also share your point on the backwards compatibility. 

But @jwilder what do you think about the template refactoring? If you agree that this makes sense (in terms of improved readability and DRY) and you would consider it worth merging, then I'll invest some work into separating the additional features from the refactoring and create a new PR.

Vad1mo:
@DrMurx I was looking at your refactoring regarding the templates. Your refactoring makes it much easier understanding the template. I would even go further and split it in multiple files so a user of the library could replace parts of the of template the fit his needs without the need to submit PRs. 

@jwilder It would be very beneficial to refactor the template as the current stat is difficult to understand.

PR #709, opened 2017-02-09
Compose v3: Moved shared volume to top-level volumes entry

sjawhar:
For compatibility for Docker Compose v3, replaced the unnamed volume in the nginx service with a named nginx_conf volume that is shared by both services.

https://docs.docker.com/compose/compose-file/#/upgrading


endersonmaia:
I get this error :

    myservice_dockergen.1.07txurxupiwb@moby    | 2017/02/23 22:13:47 Cnable to create temp file: open /etc/nginx/conf.d/docker-gen327150582: no such file or directory

sjawhar:
@endersonmaia That's strange, running `docker-compose -f docker-compose-separate-containers.yml up -d` works flawlessly for me on both Windows and CentOS 7. Can anyone else reproduce the issue.

teohhanhui:
We have no real way of controlling which base image the files are copied from. So this may or may not work sometimes...

docker-compose-separate-containers.yml
```diff
@@ -1,23 +1,25 @@
...
 
   dockergen:
     image: jwilder/docker-gen
     command: -notify-sighup nginx -watch /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
-    volumes_from:
-      - nginx
     volumes:
       - /var/run/docker.sock:/tmp/docker.sock:ro
       - ./nginx.tmpl:/etc/docker-gen/templates/nginx.tmpl
+      - nginx_conf:/etc/nginx/conf.d
```
teohhanhui:
```yaml
- nginx-conf:/etc/nginx/conf.d:nocopy
```

docker-compose-separate-containers.yml
```diff
@@ -1,23 +1,25 @@
...
 
   whoami:
     image: jwilder/whoami
     environment:
       - VIRTUAL_HOST=whoami.local
+
+volumes:
+  nginx_conf: {}
```
teohhanhui: `nginx-conf`, because that's the convention used by Docker.

For example here: https://docs.docker.com/engine/tutorials/dockervolumes/#mount-a-shared-storage-volume-as-a-data-volume (`my-named-volume`)

docker-compose-separate-containers.yml
```diff
@@ -1,23 +1,25 @@
-version: '2'
+version: '3'
```
teohhanhui: No need to bump the version here, so that it can still work with older Docker and Docker Compose versions.

docker-compose-separate-containers.yml
sjawhar: Can you explain what you mean by "the wrong container"?

docker-compose-separate-containers.yml
sjawhar: You're right, no reason to rush to v3 until other Stack compatibility issues are fixed.

sjawhar: Thought version 3 is list as the recommended version
https://docs.docker.com/compose/compose-file/compose-versioning/

docker-compose-separate-containers.yml
sjawhar: Considering that Docker Compose names everything with an underscore (`$PROJECT_default` network, `$PROJECT_$SERVICE_1` container, etc.), I'm not so sure about "convention." Doing that results in a volume called `<project_name>__nginx-conf`, which is just... ugly.

teohhanhui: That's intended. The underscore is used for prefix (e.g. project name) and postfixs (e.g. `_default`, `_1`).

docker-compose-separate-containers.yml
teohhanhui: Docker copies files from the image into the volume when starting the container.

If files are copied from the `dockergen` container into the `nginx-conf` volume, it'd become empty. This would happen when the `dockergen` container is started before the `nginx` container. Adding `nocopy` here ensures that files are never copied from this container.

sjawhar: I just tested it, and the volume is not emptied if it has data.

docker-compose-separate-containers.yml
sjawhar: I don't agree. This seems like an issue of standards. Maybe the project maintainers have an opinion? @jwilder

docker-compose-separate-containers.yml
teohhanhui: 1. Build an `nginx` image with your own configuration files in `/etc/nginx/conf.d`.
2. Delete the containers and the volume.
3. Start the `dockergen` service.
4. Start the `nginx` service.

You'll see that the `nginx-conf` volume was initialized to empty from the `dockergen` image. Then it'll only have one `default.conf` after generation from template. Your own configuration files would be lost.

sjawhar: I finally got a chance to try this. Worked just fine, my test.conf file was still there.

teohhanhui: Did you make sure to start the `dockergen` service before the `nginx` service? Because startup order is not guaranteed, and that's why the problem could happen sometimes.

sjawhar: I was finally able to reproduce this issue, but adding `nocopy` doesn't fix it. If the problem is startup order, maybe the right solution is to add `depends_on` to the `dockergen` service?

teohhanhui: Are you sure adding `nocopy` doesn't fix it? Because it works for me. You'll need to make sure to remove the volumes before each run to be able to test it properly. `depends_on` has no effect when using `docker stack deploy`, so it's not a solution.

sjawhar: 1. If I just run `docker-compose up`, the config files are there regardless of whether `nocopy` is added to the `dockergen` service.
2. If I run `docker-compose up dockergen` then `docker-compose up nginx`, the config files are missing regardless of whether `nocopy` is added to the `dockergen` service.

Is there some other test I can run to verify the behavior you're claiming?

sjawhar: I'm starting to think this is not actually an issue. If you're using a derived `nginx` image, just place the config file somewhere other than `/etc/nginx/conf.d` and add it to `/etc/nginx/nginx.conf`. Problem solved.

PR #713, opened 2017-02-11
Add IPv6 listen address

### LINKED ISSUE #127: IPv6 Support
Hi.
Would it be possible to add an option, which adds a listen directive for IPv6?

schrieveslaach:
I want to resolve #127, adding support for IPv6


jwilder:
Thanks @schrieveslaach!

jwilder:
Reverting this since it appears to break hosts that don't support IPv6.

PR #723, opened 2017-02-16
display error message in logs

thomasleveil:
The issue tracker is bloated with support requests. The intent of this PR is to reduce such requests by helping users help themselves.

One issue users frequently encounter is when `nginx -s reload` fails. In such case, no useful info about the actual issue can be found in the docker log.

In this PR, `docker-gen` is run with the `-notify-output` option and will now notify the `reload_nginx.sh` script. This script will run the `nginx -s reload` command, <s>and in case of failure dump the nginx config files to the docker logs</s>.

----

<s>As the forego binary downloaded from https://github.com/jwilder/forego/releases/download/v0.16.1/forego is missing [this](https://github.com/ddollar/forego/issues/51) fix, this PR embed a forego binary statically build from commit [c69a0e4](https://github.com/jwilder/forego/commit/c69a0e4a4a03bd618db22b2e740605eb361de9d3) (the HEAD of jwilder/forego@master).

Ideally, @jwilder would build a new binary and make it available as before.</s>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,8 +14,9 @@ RUN echo "daemon off;" >> /etc/nginx/nginx.conf \
  && sed -i 's/^http {/&\n    server_names_hash_bucket_size 128;/g' /etc/nginx/nginx.conf
 
 # Install Forego
-ADD https://github.com/jwilder/forego/releases/download/v0.16.1/forego /usr/local/bin/forego
-RUN chmod u+x /usr/local/bin/forego
+RUN wget --quiet -O- https://bin.equinox.io/a/aHZqYk1TwYN/forego-20170111202937-linux-amd64.tar.gz \
+    | tar xz -C /usr/local/bin/ \
+    && chmod u+x /usr/local/bin/forego
 
 ENV DOCKER_GEN_VERSION 0.7.3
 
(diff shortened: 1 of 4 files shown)


jwilder:
I think the releases on the official forego site have been fixed now.

Can you try pulling in https://bin.equinox.io/a/aHZqYk1TwYN/forego-20170111202937-linux-amd64.tar.gz instead of including a binary directly?  That link comes from: https://dl.equinox.io/ddollar/forego/stable/archive

thomasleveil:
@jwilder done, enjoy :smile:

thomasleveil:
I'd like to give more thoughts to this feature. I'm afraid in its current state it would produce large amount of logs. Maybe dumping config files to the log could be activated only if the `DEBUG` environment variable is `true`?

thomasleveil:
In the end, I simplified the changes brought in this PR. There are now only 2 changes:
- upgrade to latest forego binary
- use `-notify-output` in the `docker-gen` command

and 1 addition:
- provide the `/app/dump.sh` script which can be used to dump all config files content to the console using:

        docker exec nginxproxy /app/dump.sh

thomasleveil:
closing as I discovered this change produce an infinite nginx restart loop in case you restart the nginx-proxy container and the nginx conf has an empty upstream block.

I'll investigate further and open a new PR later

PR #736, opened 2017-02-21
[WIP] get rid of forego

thomasleveil:
:warning: needs reviews

This work started when trying to remove the race condition between the nginx and docker-gen processes on container creation.

I ended up getting the whole thing work without forego. The workflow when the nginx-proxy container starts is now:

1. entrypoint is executed
1. entrypoint registers a trap for `SIGHUP` which will reload docker-gen
1. entrypoint registers a trap for `SIGUSR1` which will reload nginx
1. entrypoin removes the default.conf
1. entrypoint starts docker-gen
1. docker-gen initializes, create default.conf and sends `SIGHUP` to entrypoint
1. `SIGHUP` is caught and starts nginx

As a side effect, this fixes the crash on restart if files referenced in previous generated configuration are now missing (see [#734@7c27f3ee](https://github.com/jwilder/nginx-proxy/pull/734/commits/7c27f3eea29fe6cf77d13a626560283c46ef50ca))

Since this is a major change, I'll let this PR in [WIP] state so more contributors can review and discuss it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,10 +13,6 @@ RUN apt-get update \
 RUN echo "daemon off;" >> /etc/nginx/nginx.conf \
  && sed -i 's/^http {/&\n    server_names_hash_bucket_size 128;/g' /etc/nginx/nginx.conf
 
-# Install Forego
-ADD https://github.com/jwilder/forego/releases/download/v0.16.1/forego /usr/local/bin/forego
-RUN chmod u+x /usr/local/bin/forego
-
 ENV DOCKER_GEN_VERSION 0.7.3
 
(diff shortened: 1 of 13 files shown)


thomasleveil:
I'm thinking it might be better to let the entrypoint trap two signals:
- `SIGHUP` would be repeated to docker-gen
-  `SIGUSR1` would reload nginx

This way side projects like [letsencrypt-nginx-proxy-companion](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/blob/master/app/functions.sh#L71) will have an easy mean of triggering docker-gen. @JrCs: would that be fine?

jsmestad:
@thomasleveil any progress on this?

jsmestad:
I am using this locally and it does seem to resolve the crash on startup you mentioned. Any way to release this as a test or beta build somewhere on Docker hub?

thomasleveil:
@jsmestad I'm now using Traefik and do not use nginx-proxy anymore. I won't develop this PR any further

PR #834, opened 2017-05-26
Remove weak Cipher DES-CBC3-SHA Support

eimers:


jwilder:
Looks like this was removed via https://github.com/jwilder/nginx-proxy/pull/589

PR #837, opened 2017-05-28
Alpine build stages

dschaper:
Break Alpine build into stages, reduces file size by 10MB / 25%

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -1,25 +1,36 @@
-FROM nginx:1.13.0-alpine
-MAINTAINER Jason Wilder mail@jasonwilder.com
+# Build stages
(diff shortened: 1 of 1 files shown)


matt-hh:
https://docs.docker.com/engine/userguide/eng-image/multistage-build/
> Edge only: This topic is only applicable to Docker CE Edge versions. See Docker CE Edge.
> Multi-stage builds are a new feature in Docker 17.05...

Current Docker CE version is 17.03.1.

Very interesting feature, but we should wait for a stable release.

dschaper:
It won't pass Travis anyways until they update to be compatible with the new build structure. (But it does pass local pytest.)

buchdag:
Closed by #1470

PR #842, opened 2017-06-02
Allow containers to be restricted to internal network

kamermans:
After using `nginx-proxy` for a year or two, I've bumped into this scenario several times:

I have a `nginx-proxy` container running on a server, and use port-forwarding to send traffic from the public internet to that container.  I run many containers on that same server, but some of them are only for internal network usage (a local intranet site, etc).  Currently, external clients could access these internal sites from the public internet if they just forge the `Host` header (for example, by editing their `hosts` file).

This PR adds an environment variable `NETWORK_ACCESS`, which can be set to `external` (default) or `internal`.  If `internal` is used, the file `/etc/nginx/network_internal.conf` is included in each `server` block with the following contents:

```
# Only allow traffic from internal clients
allow 127.0.0.0/8;
allow 10.0.0.0/8;
allow 192.168.0.0/16;
allow 172.16.0.0/12;
deny all;
```

This will cause `nginx-proxy` to return an `HTTP 403 Forbidden` for requests originating from outside the network.

For users that have upstream proxies which hide the real client IP (ex: AWS ELB), there is documentation in the README on enabling the `realip` module, which is already included in `nginx-proxy` today.

The `/etc/nginx/network_internal.conf` file is externalized to easily allow users to customize their list of internal networks for their needs.

I am currently testing this PR in my production environment (~100 containers), and will include any improvements if I find any.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -147,6 +147,12 @@ upstream {{ $upstream_name }} {
 {{/* Get the VIRTUAL_PROTO defined by containers w/ the same vhost, falling back to "http" */}}
 {{ $proto := trim (or (first (groupByKeys $containers "Env.VIRTUAL_PROTO")) "http") }}
 
+{{/* Get the NETWORK_ACCESS defined by containers w/ the same vhost, falling back to "external" */}}
+{{ $network_tag := or (first (groupByKeys $containers "Env.NETWORK_ACCESS")) "external" }}
+
+{{/* Get the NETWORK_ACCESS defined by containers w/ the same vhost, falling back to "external" */}}
+{{ $network_tag := or (first (groupByKeys $containers "Env.NETWORK_ACCESS")) "external" }}
+
 {{/* Get the HTTPS_METHOD defined by containers w/ the same vhost, falling back to "redirect" */}}
 {{ $https_method := or (first (groupByKeys $containers "Env.HTTPS_METHOD")) "redirect" }}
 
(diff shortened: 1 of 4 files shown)


jwilder:
Seems reasonable. I can't say I've seen anyone ask for this feature though.

kamermans:
I agree, I tried to find other issues that would be solved with this, but a cursory search didn't yield any results.  I suspect most people that would have this problem haven't considered that all their internal sites are accessible from the public internet by simply forging the Host header.

kamermans:
Rebased on `origin/master`.  I'm still using this in my environment, no issues so far.

jwilder:
Thanks @kamermans!

kamermans:
Thanks for merging @jwilder, now I can stop using `kamermans/nginx-proxy` and rebasing every few weeks :)

nginx.tmpl
```diff
@@ -147,6 +147,12 @@ upstream {{ $upstream_name }} {
 {{/* Get the VIRTUAL_PROTO defined by containers w/ the same vhost, falling back to "http" */}}
 {{ $proto := trim (or (first (groupByKeys $containers "Env.VIRTUAL_PROTO")) "http") }}
 
+{{/* Get the NETWORK_ACCESS defined by containers w/ the same vhost, falling back to "external" */}}
+{{ $network_tag := or (first (groupByKeys $containers "Env.NETWORK_ACCESS")) "external" }}
+
+{{/* Get the NETWORK_ACCESS defined by containers w/ the same vhost, falling back to "external" */}}
+{{ $network_tag := or (first (groupByKeys $containers "Env.NETWORK_ACCESS")) "external" }}
```
josephholsten: Am I crazy, or is this block duplicated?

kamermans: You're absolutely right!  This PR was way behind master and was conflicting, so I rebased it and squashed the commit, and must have duplicated it in the process.  Thanks for reporting it, @josephholsten, I'll send a bugfix PR right away.

kamermans: Pushed fix in #977

PR #858, opened 2017-06-18
set client_max_body_size to 4G

Neilpang:
set client_max_body_size to 4G

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -17,6 +17,7 @@
 	{{ end }}
 {{ end }}
 
+client_max_body_size 4096M;
 # If we receive X-Forwarded-Proto, pass it through; otherwise, pass along the
 # scheme used to connect to this server
 map $http_x_forwarded_proto $proxy_x_forwarded_proto {


dhrp:
Why would you want to do that?

Neilpang:
The default size is about 2MB,   it's too small as a production reverse proxy server.  For example, I want to upload a photo to my Wordpress blog,  the images size is usually more than 2MB.

Neilpang:
@dhrp

dhrp:
Ok, so in your setup, for a Wordpress blog, the size is too small. Did you consider all other possible situations where the proxy is used? e.g. as a frontend for an api, hosting a static site, etc. etc. This proxy is used in a wide variety of situations. So changing it globally is not the place.

What you should do is add this line where it should be: in /etc/nginx/vhost.d/your.site.name it will then end up in the config for your site.

Neilpang:
@dhrp 
I knew where to change the conf.

But I think 2MB is a too small size for most current web use cases.

jeeftor:
We ran into this error yesterday.  Hope this gets pulled soon.

kamermans:
@dhrp is right, `4G` is way, way to high to run safely.  With such a high limit, it would be quite easy to DoS an `nginx-proxy` instance, possibly even leading to an OOM (which could crash `nginx`), or a full disk (which could crash everything else).  How many 4GB files would you need to slowly upload to fill up all the RAM and a 20GB data partition?  Not many!

@jeeftor: you don't need this pull request (unless @jwilder wants it), as this can already be done via configuration.

I think if the argument was about `16M` vs `2M`, or something like that, there would be some good points on both sides, but moving from `2M` to `4G` seem quite extreme to me.

On another note, I am also running into this problem, but I was hoping to find a way to solve it on a per-host basis.  For example, you could add an env var on the downstream containers like `CLIENT_MAX_BODY_SIZE=4G` that triggers `nginx.tmpl` to add that directive to just that one section.  This way you can have different values for each container.

ingwinlu:
this should not be necessary at all. the readme has information on how to config either globally or on a vhost basis (which i recommend).

kamermans:
Good point @ingwinlu, you can definitely add per-vhost config files.  My scenario is this:

I have several hosts that each run `nginx-proxy`.  I have dozens of Docker services that can be run on any/all of these hosts.  My issue is that in order to use the vhost-based config files, I need to change something on the host's `nginx-proxy` instance.  Currently I do all vhost-specific configuration in the `docker-compose.yml` files that I use to start my services, which works very nicely.

Ideally I could add something like this to my service's `docker-compose.yml`:
```
volumes:
    - ./nginx/vhost.d:/etc/nginx/vhost.d:ro
```

Then `nginx-proxy` could reach out to that container, pull out the per-vhost configs and add them to its config.  Unfortunately, the containers are isolated from each other, so this is probably impossible.

I think I'll need to come up with some sort of creative strategy to allow me to create these `vhost.d` files on the fly when/before my service containers start.

jwilder:
4G is too high for a default for this project.

Issue #880, opened 2017-07-21
Security updates for using nginx 1.13.0 in dockerfile


yexingok:
Hi, 

I notice the [Dockerfile](https://github.com/jwilder/nginx-proxy/blob/master/Dockerfile#1) is using nginx:1.13.0 on upstream,

Which could be vulnerable to [CVE-2017-7529](http://mailman.nginx.org/pipermail/nginx-announce/2017/000200.html) , 

To get auto update from upstream nginx, shall we update docker file to use nginx:1.13 which always get security fix automatically once upstream updates?

PR #881, opened 2017-07-21
Upgrade to nginx 1.13.3

matt-hh:
Important security fix in nginx 1.13.3
See http://nginx.org/en/CHANGES


matt-hh:
Closed in favour of #880 and #882

PR #882, opened 2017-07-21
Move to 1.13 base image to get auto updates

### LINKED ISSUE #880: Security updates for using nginx 1.13.0 in dockerfile

matt-hh:
Less maintenance and more security
Resolves #880


yexingok:
Thanks @matt-hh

PR #914, opened 2017-08-28
fix: fix 'create-server-certificate' by using a Dockerfile

matteocng:
`openssl` was removed from the `docker` `nginx` image (https://github.com/nginxinc/docker-nginx/issues/182); `nginx:1.13` is affected.
```
docker run -it nginx:1.13 openssl
docker: Error response from daemon: oci runtime error: container_linux.go:262: starting container process caused "exec: \"openssl\": executable file not found in $PATH".
```

Command: `./create_server_certificate.sh "*.example.com"`

Errors:
```
Create a host key: /Users/matteocng/web/nginx-proxy/test/certs/*.example.com.key
rpc error: code = 2 desc = oci runtime error: exec failed: container_linux.go:262: starting container process caused "exec: \"openssl\": executable file not found in $PATH"

Create a host certificate signing request
rpc error: code = 2 desc = oci runtime error: exec failed: container_linux.go:262: starting container process caused "exec: \"openssl\": executable file not found in $PATH"

ERROR: failed to generate server certificate signing request
```

This PR introduces a simple `Dockerfile` for an `nginx` image with `openssl`, and modifies `create_server_certificate.sh`to build it and use it.

### DIFF
--- a/test/certs/Dockerfile
+++ b/test/certs/Dockerfile
@@ -1,0 +1,8 @@
+FROM nginx:1.13
+LABEL maintainer="Jason Wilder mail@jasonwilder.com"
+
+# Install `openssl` and clean up the `apt` cache.
+RUN apt update \
+ && apt install openssl -y --no-install-recommends \
+ && apt-get clean \
+ && rm -r /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


mac2000:
In case the problem is with creating yet another docker image, may be we can just add `apt update && apt install openssl -y` at line 30? which also solve problem

Or if there is no problems with images maybe it will be better to create separate repo and image for this stuff so anyone can use it out of the box?

Issue #845, opened 2017-06-08
[emerg] 1#1: host not found in upstream "down" in /etc/nginx/conf.d/default.conf


m-a-v:
I have many containers. One of the containers didn't get an IP address (due to a misconfiguration).

With the current nginx.tmpl file this creates an incorrect nginx.conf file, since .Network.IP is empty. As a result ALL hosts are down.

	{{ else if .Network }}
		# {{ .Container.Name }}
		server {{ .Network.IP }} down;
	{{ end }}

Any idea how to solve this?

therealgambo:
Had this happen recently as well. Could we not just use `127.0.0.1` if the `.Network.IP` is empty? Nginx requires  an IP address or domain for any upstream server declaration.

m-a-v:
This seems to be a very critical issue. After a reboot one docker image did not get an IP address. If this is happens ALL the other services become unavailable.

I tested the solution from therealgambo and it works as expected.. Is there any disadvantage using the localhost (127.0.0.1) ip address?

mprokopov:
I can confirm, this is VERY annoying.

m-a-v:
Since 22th June I use the method described above without any problems.

ericdfields:
+1. Just ran into this.

Niondir:
Fix is working for me as well. Should just be added.

therealgambo:
Fix has been merged. This can now be closed.

m-a-v:
Duplicate of #926

PR #585, opened 2016-09-27
always add server to upstream (hot)fix #438 #565

schmunk42:
Prevents broken nginx configuration, might be implemented in a nicer way...

Credits: @deacix

PR #926, opened 2017-09-13
Fallback if container has no IP

Paike:
Sometimes containers will not be assigned an IP (after reboot or due to misconfiguration). This leads to an incorrect `server <missing ip here> down;` line in default.conf and crashes nginx. 
@therealgambo provided a fix for this: #845

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -13,8 +13,13 @@
 		{{ end }}
 	{{ else if .Network }}
 		# {{ .Container.Name }}
-		server {{ .Network.IP }} down;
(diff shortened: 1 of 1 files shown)


schmunk42:
Similar to https://github.com/jwilder/nginx-proxy/pull/585 

@jwilder Can we get one of the fixes for this? Would be really needed.

fzarifian:
@jwilder could you provide us information about this unmerged issue identified 6 months ago ?

jwilder:
Thanks @Paike!

Issue #938, opened 2017-09-26
Generated nginx config is invalid if namesevers are IPv6-only


pupssman:
If hosts' `/etc/resolv.conf` looks like
```
nameserver ::1
nameserver 2a02::1:1
search mydomain.com
options timeout:1 attempts:1
```
then nginx config in the container gets line
```
resolver 2a202::1:1
```
which, apparently, is illegal:
```
forego     | starting dockergen.1 on port 5000
forego     | starting nginx.1 on port 5100
nginx.1    | 2017/09/26 09:03:28 [emerg] 28#28: invalid port in resolver "2a02::1:1" in /etc/nginx/conf.d/default.conf:33
forego     | starting nginx.1 on port 5200
forego     | sending SIGTERM to nginx.1
forego     | sending SIGTERM to dockergen.1
```

Instead, it should be like `resolver [2a202::1:1]` as per nginx doc: http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver

kamermans:
I will try to fix this for you tonight.  It broke when we pulled in the code for OCSP stapling.

kamermans:
I've submitted the fix PR at #951 

You can test it out now with image `kamermans/nginx-proxy:bugfix_ipv6_resolvers`.

buchdag:
Fixed by #951

PR #951, opened 2017-10-20
Added support for IPv6 DNS servers in resolver generation (issue #938)

kamermans:
Bugfix for #938

Currently, if there are IPv6 nameservers in `/etc/resolv.conf`, they are not properly formatted when they are inserted into the Nginx template.  Nginx requires them to be surrounded by `[]` like `[2a02::1:1]`

This PR assumes all IPv6 addresses contain a colon character `:` (which I believe to be true), and surrounds them with square brackets before passing them to Nginx.

@pupssman - I've pushed this branch to my repo at `kamermans/nginx-proxy:bugfix_ipv6_resolvers` (built from `master`) - can you please give it a spin to make sure it works ok in your environment?  It's working fine here, but I don't have IPv6 support in my test environment, so I would feel better if it was tested properly.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -18,8 +18,8 @@ fi
 # Note: if $DHPARAM_BITS is not defined, generate-dhparam.sh will use 2048 as a default
 /app/generate-dhparam.sh $DHPARAM_BITS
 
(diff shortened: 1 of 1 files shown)


jwilder:
Thanks @kamermans!

PR #954, opened 2017-10-23
Fix dhparam.pem path

evertramos:
The dhparam.pem is located in the certs folder.


evertramos:
I believe what should be fix now is this lines:

https://github.com/jwilder/nginx-proxy/pull/964

PR #955, opened 2017-10-25
Fix default dhparam.pem when using separate containers

buchdag:
The existence of `/etc/nginx/dhparam/dhparam.pem` should be tested and not presumed, as the separate container setup described in the doc won't auto generate this file (neither `docker-gen` nor `nginx` has this capability), and SSL can work without it.

Presuming the file is here leads to failure when one attempt to use SSL on the separate container setup with self provided certificate (ie without using an automated solution like the recommended [JrCs/docker-letsencrypt-nginx-proxy-companion](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion) or when using self signed default cert).

This issue has been described by @returntrip there : https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/issues/264#issuecomment-338893690


jwilder:
Thanks @buchdag!

PR #869, opened 2017-07-02
Remove TLS 1.0 for PCI compliance

buchdag:
TLS 1.0 has finally been EOL'd by PCI on 30/04/2018: https://payment-services.ingenico.com/int/en/ogone/support/products/tls

PR #957, opened 2017-10-27
Allow optional use of Mozilla modern SSL configuration

buchdag:
This PR is a follow up to #869 and the discussion there with @kamermans.

It allows to use the Mozilla modern SSL configuration (TLS v1.2 only, no DHE) on a per proxyed container basis by setting the environment variable `MODERN_SSL` to `true`.

The template still default to the intermediate profile if `MODERN_SSL` isn't set.


buchdag:
@jwilder > any advice on this ?

kamermans:
I feel like it should be an enumerable value as opposed to a bool.  We could, for example, mirror the AWS ELB Security Policies: http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-security-policy-table.html

And/or we could mirror the Mozilla security policies (Modern, Intermediate, Old): http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-security-policy-table.html

I think an option like `SECURITY_POLICY` or `SSL_POLICY` would make sense with the following values:
- `Modern`
- `Intermediate` (default)
- `Old`

Or

 - `Mozilla-Modern`
 - `Mozilla-Intermediate` (default)
 - `Mozilla-Old`
 - `AWS-TLS-1-1-2017-01`
 - `AWS-TLS-1-2-2017-01`
 - `AWS-2016-08`
 - `AWS-2015-05`
 - `AWS-2015-03`
 - `AWS-2015-02`

Maybe I'm overthinking it, but when I see things like `MODERN=true` I can't help but wonder if it we will need `MODERN_NEW` and `MODERN_NEWER` at some point in the future :)

I guess you were planning to track the Mozilla Modern profile anyway, so that is future-proof from what I understand - it gets updated over time and we will need to update accordingly.

Another approach would be to expose `SSL_PROTOCOLS` and `SSL_CIPHERS`, but that can get pretty nasty and it doesn't really encompass OCSP stapling, HSTS, etc.

buchdag:
Well truth is I hesitated between the enumerable value with different settings and the simple bool before doing this, and I went for the simpler version just for the sake of keeping it simple and not adding too much stuff.

I agree that a more comprehensive `SSL_POLICY` option is the better approach and if it suits @jwilder I'll modify the PR in this way.

That get a little bit hairier if this option includes `Mozilla-Old` as this configuration calls for a 1024 bits DH parameter size, while both this container and letsencrypt-nginx-proxy-companion create a 2048 bits dhparam.

kamermans:
All great points!

jwilder:
I like idea @kamermans suggested (e.g. `SSL_POLICY=...`).  I think that would be a little more robust over time as `Modern` will probably change in the future and we add new values vs new env vars.

buchdag:
I updated the PR with the idea suggested by @kamermans , and `generate-dhparam.sh` was also modified to defaults to a 1024 bits dhparam when `SSL_POLICY` is set to `Mozilla-Old`.

buchdag:
On a side note I have no idea why `README.md` shows way more changes than I really did.

I'm using atom and not doing anything fancy, any idea where does that come from ?

kamermans:
The edits in `README.md` are due to Atom stripping trailing spaces off.  Although I also use Atom, I use *MarkdownPad* on Windows and *Remarkable* on Linux to edit Markdown files because they have a nice GitHub-Flavored-Markdown live-preview.

What if, instead of putting the SSL params in an if statement in the `nginx.tmpl`, we instead created a folder like `/etc/nginx/ssl_policies` and put inside it `Mozilla-Intermediate.conf`, `Mozilla-Modern.conf` and all the others, then in `nginx.conf`, we could just do this in place of the `if` statement:

```
{{ if (exists $ssl_policy) }}
include /etc/nginx/ssl_policy/{{ $ssl_policy }}.conf;
{{ end }}
```

That way if people want to include their own custom policy, they can just mount it into `/etc/nginx/ssl_policy/somepolicy.conf` and use `SSL_POLICY=somepolicy`.

The only problem I see with this method is that it doesn't account for the `ssl_dhparam` being related to the policy.

generate-dhparam.sh
```diff
@@ -1,7 +1,12 @@
 #!/bin/bash -e
 
-# The first argument is the bit depth of the dhparam, or 2048 if unspecified
-DHPARAM_BITS=${1:-2048}
+# The first argument is the bit depth of the dhparam, 1024 if unspecified and SSL_POLICY is set to Mozilla-Old,
+# or 2048 if unspecified and SSL_POLICY is set to any policy other than Mozilla-Old (including the default).
+if [[ $SSL_POLICY == "Mozilla-Old" ]]; then
```
kamermans: Technically `==` is not correct in Bash (or any other POSIX shell).  The equality operator is a single `=`, but `==` is widely tolerated.

buchdag: Will fix, thanks !

buchdag:
> What if, instead of putting the SSL params in an if statement in the nginx.tmpl, we instead created a folder like /etc/nginx/ssl_policies and put inside it Mozilla-Intermediate.conf, Mozilla-Modern.conf and all the others, then in nginx.conf, we could just do this in place of the if statement:

That would break a lot of existing and new setups with [letsencrypt-nginx-proxy-companion](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion), namely split nginx + docker-gen setup with the updated nginx.tmpl (and the split setup is recommended by letsencrypt-nginx-proxy-companion doc, so it's probably widely used).

Being the [recommended](https://github.com/jwilder/nginx-proxy#ssl-support-using-letsencrypt) way of getting LE certificate with nginx-proxy, I think backward compatibility with @JrCs's letsencrypt-nginx-proxy-companion should be more thorougly checked when doing SSL related changes (it wasn't in two cases already, dhparam.pem location and [chain file extension for OCSP stapling](https://github.com/jwilder/nginx-proxy/pull/574#issuecomment-339926552)).

I have no problem submitting PR to `letsencrypt-nginx-proxy-companion` to track changes like these in `nginx-proxy` if you think it should be the other way around, but truth is getting stuff merged there is increasingly slow as the maintainer seems to be out of time to spare for this project.

> [EDIT] isn't `ssl_dhparam` a good idea in all modes? In modern mode, plain `DHE` is already disallowed by being absent from `ssl_ciphers`, but Diffie-Hillman Exchanges are still used in the modern TLS 1.2 profiles (the `ECDHE` in` ECDHE-ECDSA-AES256-GCM-SHA384` is `Elliptical Curve DHE`). Any time you're using DH, you're better off with a unique `dhparam`. In short, I think you should leave it in for all profiles.
> 
> [EDIT 2] I've confirmed that you need `ssl_dhparam` with the modern profile or you cannot perform PFS (perfect forward secrecy), and therefore you cannot get an A+ SSL rating on most sites.

Ok, that's puzzling. The [Mozilla SSL Configuration Generator](https://mozilla.github.io/server-side-tls/ssl-config-generator/) explicitly remove the `ssl_dhparam` from nginx config with the modern profile. When I tested the PR on SSL Qualys before submitting it, I did get an A+ and I think PFS was ok, but that was three weeks ago. Do you want me to test again ?

I'd rather keep the `ssl_dhparam` in all modes as you suggest even if the configuration generator says otherwise.

kamermans:
Good points about `letsencrypt-nginx-proxy-companion`.   I'm the one that usually breaks compatibility with that project, so I will pay more attention to it in the future.  I'm OK with leaving the logic as you have it then, just trying to keep the template manageable :)

Regarding dh parameters, there is some weird information.  I now understand that dh parameters are only used in plain `DHE`.  For `ECDHE`, the server uses points on a mathematical curve instead of large safe primes. There is still a concept of dhparams, but it's not `ssl_dhparam`.   From what I can tell, the dh parameters are removed from the Modern profile because there is no support for `DHE`.

Regarding PFS, it looks like it will work fine (as you've confirmed) as long as `ECDHE` is used as the key exchange.

Can you confirm that you get A+ when you leave the dh parameters in?  It should be ignored.

buchdag:
I'm testing right now.

The 1024 bits dhparam modification I made is garbage and I'll revert it, I completely lost sight on the fact that you set `SSL_POLICY` on the proxyed containers, not on the nginx-proxy/docker-gen container.

What would you recommend regarding this ?

1. Do not include `Mozilla-Old` at all.
2. Include it but use the default 2048 bits dhparam.
3. Include it, warn in the doc that you have to create a 1024 dhparam at let say `/etc/nginx/dhparam/dhparam1024.pem` and have the template pick it up if it exists and the SSL policy is `Mozilla-Old`.
4. Include it, have `generate-dhparam.sh` create both a 1024 and a 2048 bits dhparam files and let the template use the correct one depending on the SSL policy as above.

kamermans:
I think I would keep it at 2048 and put a note in the readme that the old profile requires dh parameters <= 1024 bits, and that can be done using the `DHPARAM_BITS=1024` environment var (please double check that, I'm replying from my phone).

buchdag:
[Without `ssl_dhparam` (the current state of the PR).](https://www.ssllabs.com/ssltest/analyze.html?d=mozilla-modern-without-dhparam.duuu.ch)

[With `ssl_dhparam`.](https://www.ssllabs.com/ssltest/analyze.html?d=mozilla-modern-with-dhparam.duuu.ch)

No difference at all (aside from the unrelated SNI stuff)

kamermans:
Look great, thanks for the links!  I slightly prefer leaving the `ssl_dhparam` there and letting Nginx ignore it since it's also present in other places (it's still generated, etc).  I'd say this PR is good to go.  Thanks for all your work on this!
CC @jwilder

buchdag:
`ssl_param` handling has been reverted to the way it is now in `master`.

I didn't want to further rebase the commits as there is a merge in between, I'm pretty sure I'll end up breaking something if I try.

buchdag:
@jwilder > any advice on the modified PR ?

buchdag:
@jwilder up ?

jwilder:
Thanks @buchdag and @kamermans!

rgarcia89:
Has this been tested?
I have added the env.Variable to my build command, but still get a A ranking in Qualys SSL Labs...
Am I missing something?

```
docker run -d -p 80:80 -p 443:443 \
       --restart always \
       --name nginx-proxy \
       -e SSL_POLICY=Mozilla-Modern \
       -v /path/to/certs:/etc/nginx/certs:ro \
       -v /docker/config/nginx_proxy:/etc/nginx/vhost.d \
       -v /docker/config/nginx_proxy:/usr/share/nginx/html \
       --label com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_proxy \
       jwilder/nginx-proxy
```

jwilder:
@rgarcia89 I think you might need to set the env var on the containers and not the proxy container.

rgarcia89:
@jwilder thanks I didnt get it out of the comments above -sry.
Now I only need to figure out how to get fixed the below warning.

**Strict Transport Security (HSTS)**  Invalid   Server provided more than one HSTS header

kamermans:
That's a strange error, you should send a "curl -I" command to your server and see if there are two HSTS headers.

rgarcia89:
I am indeed seeing twice the Strict-Transport-Security entry O_o

```
HTTP/1.1 302 Found
Server: nginx/1.13.10
Date: Mon, 02 Apr 2018 18:22:34 GMT
Content-Type: text/html; charset=UTF-8
Connection: keep-alive
Set-Cookie: ocadc83b19e7=0dea07708e0e22a15f785710bed13e2e; path=/; secure; HttpOnly
Expires: Thu, 19 Nov 1981 08:52:00 GMT
Cache-Control: no-store, no-cache, must-revalidate
Pragma: no-cache
Set-Cookie: oc_sessionPassphrase=mq7xVitfrC2d1b41Vq7Ef6s8EYX3AFF00qmoUfu5bxNgYkXEDlwygBbfkthnn3r9r3dYZSKetxN2E6yowYbR1zSTDbHsTcU%2BB1DGwrCQz92c55RtjVzFagXy%2F1GpK1tl; path=/; secure; HttpOnly
Content-Security-Policy: default-src 'self'; script-src 'self' 'unsafe-eval' 'nonce-QXRPZG1kYUNRVnRFVVBINHdpcmZ5WXlKTGFwSmZWQVhsYXNPOXFIQXdKbz06VjdYWHo1N2xKUXN5ZjRDcnBRR2FndWF4WE1nOERpRmwrdWxtaHNpUGxQND0='; style-src 'self' 'unsafe-inline'; frame-src *; img-src * data: blob:; font-src 'self' data:; media-src *; connect-src *; object-src 'none'; base-uri 'self';
X-Frame-Options: SAMEORIGIN
Set-Cookie: __Host-nc_sameSiteCookielax=true; path=/; httponly;secure; expires=Fri, 31-Dec-2100 23:59:59 GMT; SameSite=lax
Set-Cookie: __Host-nc_sameSiteCookiestrict=true; path=/; httponly;secure; expires=Fri, 31-Dec-2100 23:59:59 GMT; SameSite=strict
Location: https://xx.xxx.de/login
Strict-Transport-Security: max-age=15768000; includeSubDomains; preload
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
X-Robots-Tag: none
X-Download-Options: noopen
X-Permitted-Cross-Domain-Policies: none
Strict-Transport-Security: max-age=31536000
```

rgarcia89:
I am basically trying to achieve what is mentioned here.
https://github.com/Wonderfall/docker-nextcloud#reverse-proxy
Seems like the nextcloud container is already creating the HSTS entry :/

Issue #953, opened 2017-10-23
More control over the Strict-Transport-Security header


Archcry:
It would be nice to have more control over the `Strict-Transport-Security` header so that we can specify which options we want to apply. To be compliant with the HSTS Preload List submission requirements we would have to add the `includeSubDomains` and `preload` attributes to the `Strict-Transport-Security` header like so:

`add_header Strict-Transport-Security "max-age=31536000; includeSubDomains; preload";`

It would also be nice if we could specify the max age, although this is not a requirement since 31536000 is already compliant with the submission requirements.

For more information on the requirements see: https://hstspreload.org/

kamermans:
I can add a PR for this if @jwilder is interested in pulling it.

Basically we just need to modify this line: https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl#L229

One way to do it would be to have `HSTS=off` or let the user put their own HSTS in, like `HSTS=max-age=31536000; includeSubDomains; preload`, with a default like we have today: `HSTS=max-age=31536000`.  This is the most attractive of the methods I've proposed here in my opinion because it makes the least assumptions.

Another method would be to parameterize it:
- `HSTS_ENABLE=true`
- `HSTS_MAX_AGE=31536000`
- `HSTS_SUBDOMAINS=true`
- `HSTS_PRELOAD=true`

A third option would make it work like `HTTPS_METHOD`:
- `HSTS_METHOD=off|subdomains|preloads|preload-subdomains`
- `HSTS_MAX_AGE=31536000`

Thoughts?

Archcry:
I think it would be nice to have the option to configure the header yourself and have a sensible default (which does not include subdomains for when the user wants to run a subdomain without HTTPS).
So the first option would be nice because then the user can define his or her own header without having a lot of environment variables dedicated to it.

though the last option also has nice flexibility and prevents user error.

jwilder:
@kamermans I'd prefer fewer environment variables so what you proposed here sounds good:

> One way to do it would be to have HSTS=off or let the user put their own HSTS in, like HSTS=max-age=31536000; includeSubDomains; preload, with a default like we have today: HSTS=max-age=31536000. This is the most attractive of the methods I've proposed here in my opinion because it makes the least assumptions.

kamermans:
Ok, cool, I'll send a PR.

Archcry:
Closing this issue because a custom environment variable has been added

PR #969, opened 2017-11-09
Added custom HSTS support (issue #953)

kamermans:
This PR adds custom HSTS support as discussed in #953.

It's available from my Docker repo as `kamermans/nginx-proxy:feature_custom_hsts`.

Note that the tests are passing, but there are two other tests failing on my build machine.  At least one of them is due to a change in output in `openssl 1.1.0` on my Ubuntu 17.10 test machine, so I'll probably send a PR for that issue later.


Archcry:
Looks good to me, defaults are set to the previous implementation so this should not affect existing users in any way.

PR #970, opened 2017-11-09
Fixed failing SSL tests

kamermans:
The following tests are failing on my build machine (Ubuntu 17.10):

1. `test_ssl/test_dhparam.py::test_web5_dhparam_is_used` - was failing because the `openssl s_client` was used with `-verify 0` to disable certificate verification.  `-verify 0` has actually been wrong forever, but only recently has `openssl` stopped tolerating it.  Removing the `-verify` statement entirely disables certificate verification properly.

2. `test_ssl/wildcard_cert_and_nohttps/test_wildcard_cert_nohttps.py::test_web3_https_is_500_and_SSL_validation_fails` - was failing because `requests` raises a `SSLError(CertificateError)`, but we're expecting the inner error, a `CertificateError`.  I don't know if this is related to my version of Python (`2.7.14`) or `requests` (`2.18.4)`, but on my machine I get an `SSLError`, not a `CertificateError`.

It would be nice to have another pair of eyes on test 2, since python is not my native language.  I wrote test 1 in the first place, so I'm pretty comfortable with that one :)

### DIFF
--- a/test/test_ssl/test_dhparam.py
+++ b/test/test_ssl/test_dhparam.py
@@ -89,5 +89,5 @@ def test_web5_dhparam_is_used(docker_compose):
 
     host = "%s:443" % sut_container.attrs["NetworkSettings"]["IPAddress"]
     r = subprocess.check_output(
-        "echo '' | openssl s_client -verify 0 -connect %s -cipher 'EDH' | grep 'Server Temp Key'" % host, shell=True)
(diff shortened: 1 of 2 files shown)


kamermans:
Ok, I saw from the Travis-CI output that older versions of `requests` raise `CertificateError` and new ones raise `SSLError`, so I've added them both.  Now it passes in both environments.

jwilder:
Thanks!

PR #977, opened 2017-11-13
Removed duplicate lines

kamermans:
Removed duplicated lines in `nginx.tmpl` as reported by @josephholsten.


kamermans:
This is failing due to a Travis-CI dependency issue in Alpine.  I haven't traced it back to the root yet, but I suspect that triggering it again will fix it.

PR #984, opened 2017-11-19
OCSP Stapling was not working

sydoveton:
The nginx.tmpl was looking for the chain certificate with the file extension of .crt when in fact the chain certificate has a .pem extension. Adjusted the template to ensure SSL stapling was correctly enabled by detecting the correct file.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -222,10 +222,10 @@ server {
 	ssl_dhparam {{ printf "/etc/nginx/certs/%s.dhparam.pem" $cert }};
 	{{ end }}
 
-	{{ if (exists (printf "/etc/nginx/certs/%s.chain.crt" $cert)) }}
+	{{ if (exists (printf "/etc/nginx/certs/%s.chain.pem" $cert)) }}
 	ssl_stapling on;
 	ssl_stapling_verify on;
-	ssl_trusted_certificate {{ printf "/etc/nginx/certs/%s.chain.crt" $cert }};
+	ssl_trusted_certificate {{ printf "/etc/nginx/certs/%s.chain.pem" $cert }};
 	{{ end }}
 
 	{{ if (and (ne $https_method "noredirect") (ne $hsts "off")) }}


sydoveton:
Confirmed with SSL labs before and after and confirmed it is now working ok.

image

buchdag:
I confirm that due to this (.crt instead of .pem) OCSP stapling does not work with projects like `letsencrypt-nginx-proxy-companion`.

buchdag:
@jwilder , @kamermans could you take a look at this ?

kamermans:
Thanks @sydoveton and @buchdag, looks good to me!  @buchdag, correct me if I'm wrong, but you're saying that if this PR is merged, the fact that the `.pem` extension is now used will make it easier to use OCSP stapling in the `letsencrypt-nginx-proxy-companion`?

kamermans:
Also, I could swear that there was some verbiage in the README about enabling OCSP stapling, but now I don't see it.  Maybe you guys recall seeing it?  In any case, we should add a section like this to the README after `SNI` if it is indeed missing:

```
#### OCSP Stapling
To enable OCSP Stapling for a domain, `nginx-proxy` looks for a PEM certificate containing the trusted
CA certificate chain at `/etc/nginx/certs/<domain>.chain.pem`, where `<domain>` is the domain name in
the `VIRTUAL_HOST` directive.  The format of this file is a concatenation of the public PEM CA
certificates starting with the intermediate CA most near the SSL certificate, down to the root CA.  This is
often referred to as the "SSL Certificate Chain".  If found, this filename is passed to the [NGINX
`ssl_trusted_certificate` directive](http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_trusted_certificate)
and OCSP Stapling is enabled.
```

buchdag:
@kamermans it will make OCSP work out of the box compared to not working at all `letsencrypt-nginx-proxy-companion` automatically generate chain files and [creates symlinks to them as `/etc/nginx/certs/<domain>.chain.pem`](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/blob/e6db45e24c06626a3aa58991b92db5e0c1273943/app/letsencrypt_service#L39)

It could be the other way around (modifying `letsencrypt-nginx-proxy-companion` to create `/etc/nginx/certs/<domain>.chain.crt` symlinks instead), but the de-facto extension for chain files seem to be `.pem` (it's what `cerbot` does).

buchdag:
ping @jwilder

buchdag:
@jwilder again, could you please take a look at this and consider it for merging ? I'm sorry to insist but this is just an incredibly small, six characters change that would just make something work out of the box for `letsencrypt-nginx-proxy-companion` users instead of not working at all.

It's actually the same person that contributed OCSP support to both [`nginx-proxy`](https://github.com/jwilder/nginx-proxy/pull/574) and [`letsencrypt-nginx-proxy-companion`](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/pull/108), but the owner of the later decided to change the `.crt` to `.pem` while `nginx-proxy` kept `.crt`

jwilder:
Thanks @buchdag @kamermans @sydoveton!

Issue #729, opened 2017-02-18
How to properly tell nginx-proxy to use bridge from docker-compose?


patakijv:
When I start nginx-proxy using

`docker run -d -p 80:80 -p 443:443 -v /var/run/docker.sock:/tmp/docker.sock:ro -v /etc/nginx/certs:/etc/nginx/certs:rw --name nginx-proxy jwilder/nginx-proxy`

I see the following network definition for the nginx-proxy container

```
            "Networks": {
                "bridge": {
                    "IPAMConfig": null,
                    "Links": null,
                    "Aliases": null,
                    "NetworkID": "c3d28dfbcfe6c8a4dd28f5d782c205f3cafdac990bb1d0685d4b62ba28e98a79",
                    "EndpointID": "6b1c06ecae27fd994d2a12b4d8818aa53b90b4153284af9e97e4396644e66445",
                    "Gateway": "172.24.26.1",
                    "IPAddress": "172.24.26.17",
                    "IPPrefixLen": 24,
                    "IPv6Gateway": "",
                    "GlobalIPv6Address": "",
                    "GlobalIPv6PrefixLen": 0,
                    "MacAddress": "02:42:ac:18:1a:11"
                }
            }
```

but if I start docker using docker-compose with the following docker-compose.yml file

```
version: '2'

services:
  nginx-proxy:
    image: jwilder/nginx-proxy
    container_name: nginx-proxy
    restart: always
    mem_limit: 4g
    `ports:`
      - 80:80
      - 443:443
    volumes:
      - /var/run/docker.sock:/tmp/docker.sock:ro
      - /etc/nginx/certs:/etc/nginx/certs
```

if get the following network definition

```
            "Networks": {
                "nginxproxy_default": {
                    "IPAMConfig": null,
                    "Links": null,
                    "Aliases": [
                        "nginx-proxy",
                        "c091a1b37d8e"
                    ],
                    "NetworkID": "e05cc93cbe94d2bfc2a545df41c490f84be62329736a3ca74fab575ed014379f",
                    "EndpointID": "369993e43d114f2b5763048fc0451671450d8c3d21e93cad2c5428c93a8b5935",
                    "Gateway": "172.17.0.1",
                    "IPAddress": "172.17.0.2",
                    "IPPrefixLen": 16,
                    "IPv6Gateway": "",
                    "GlobalIPv6Address": "",
                    "GlobalIPv6PrefixLen": 0,
[...]

PR #1029, opened 2018-01-15
Use bridge as network_mode in examples

mattlubner:
See: #729 

Connecting dependent projects' to the correct Docker network has a common gotcha when using nginx-proxy. Setting `network_mode: "bridge"` in the default `docker-compose.yml` should help to address this.


buchdag:
Using a docker network is the preferred way.

PR #1037, opened 2018-01-23
Adding the ability to serve static files

Mactory:
I have come across the problem, that I have one webserver where I want to mix my web-apps with some static websites. As I really like the easy, pluggable way of setting my web-apps up with nginx-proxy (+ let's encrypt companion), I wanted to use this infrastructure to also serve my static files.
I therefore added a new backend that configures nginx to directly serve static files from a mounted volume in the nginx-proxy container.
To allow for static containers to be stopped, I also added the possibility to include stopped containers via an environment variable.


dhm116:
I could have avoided having to spin up a dummy python web server to serve the assets if this had been available - looking forward to having this feature merged in!

Mactory:
I'm still interested in seeing this merged in. I updated the branch to be up-to-date with the upstream branch.

adeel41:
Why this is not pulled in yet? is this project dead?

alexkoval123:
any updates?

JonathonReinhart:
The biggest problem I see with this attempt is that it still requires that the volume with the static content be mounted into the `nginx-proxy` container, which must happen manually; `nginx-proxy` can't dynamically add new volume/bind mounts to itself.

The end result is that you have configuration split between two containers (or `docker-compose.yml` files):
- `nginx-proxy` configuration determines where the static data is mounted
- The source app container sets `VIRTUAL_ROOT` which must match the above path

I love the automatic configuration that `nginx-proxy` provides, but I can't use it for Dockerized web-apps which have static content I want to serve with NGINX. I landed here looking for a solution to that problem, but unfortunately due to the limitations listed above, I don't think this helps. Thanks though, @Mactory.

dhm116:
Thanks for the insights @JonathonReinhart and @buchdag, very much appreciated. My gut was telling me that it wasn't possible, but I was hoping for something within the SDK that I was unfamiliar with 😞 

I think I would tend to agree that having to explicitly mount volumes to the `nginx-proxy` configuration at runtime would not be very ideal or "clean".

Would an approach where `nginx-proxy` has standard, named volume(s) that could be utilized so that, rather than directly exposing the nginx `certs` or `dhparam` (or static assets) directories, these named volumes would become the interface used to mount certificates and, perhaps, static files? This should still enable the current pattern of being able to react to new containers starting with environment variables triggering template updates, with the named volumes being mapped to known destinations within the `nginx-proxy` container. Then anyone is free to create mount points to the named volumes however they choose, right? It was the first alternative that sprang to mind that might support the existing and static file hosting use-cases, so forgive me if I glossed over something or left anything important out 😊  Hopefully the underlying concept makes sense - but I can elaborate anything that may not.

PR #1100, opened 2018-03-18
Non-Docker Host

CWempe:
This PR adds support to generate the config (and certificates) for hosts that are not docker containers.

Useful if you want to proxy to another physical machine.

Just start a dummy container with the new `UPSTREAM_NAME` environment variable, because `nginx-proxy` and `docker-gen` cannot detect the correct server address if it is not a connected container.

I tried the local test before I created this PR, but it failed completely. 😞 

    2 passed, 12 xfailed, 113 error in 57.40 seconds

However I don't think this is because of my changes, since it faild with the master branch, too.


pho3nixf1re:
I currently need this functionality. Integrating a patch for now.

pho3nixf1re:
Any idea how this could be used in conjunction with the Let's Encrypt companion integration? Or would it JustWork™.

Taznc:
This works great. I manually patched my template file but this gets overwritten when updated. Is there any way to do this automatically or is this going to be merged into the build? Thank you!

pho3nixf1re:
I commented out the line that automatically pulls the latest template. I think it would be good to provide that url as an ENV variable so that way we can override it with our own. I've got a couple other customizations I've made to it as well.

bkraul:
This is much needed. Please merge!

bkraul:
In the meantime, [this project](https://github.com/mikesplain/nginx-proxy-pass-docker) does the job. However, I still think this should be a feature of jwilder/nginx-proxy.

Taznc:
Is there any way to get this merged or a feature like this added? I have set this manually multiple times but it gets overwritten. Thank you! @jwilder

jolly-jump:
Please, please, I'm all for merging this one

jolly-jump:
Hi @CWempe 

BTW, if the merge won't be attractive because it seems you needed a dummy-container, you could point out, that any container would do. e.g. I reuse a "python:3-slim" container that I use anyway and added your "command: tail -f /dev/null" to the docker-compose file. This seems to do the job just as well.
Thanks for your solution anyway.

BNolet:
Did this ever get merged? Will it?

hanscees:
please merge this one?

sjdrc:
Where is this PR currently at? This functionality is super useful and will allow us to just use this one proxy instead of having to maintain this and apache2

radoslav-stefanov:
I am looking forward this feature too. Currently I use a workaround by editing nginx.tmpl to be able to generate my list of upstreams.

bkraul:
Hey man is it possible you could share and example for your tmpl workaround?

radoslav-stefanov:
> Hey man is it possible you could share an example for your tmpl workaround? Also I doubt jwilder will be doing anything about this. Been waiting on this PR for ages

It is very ugly lol, but does the job for what I want.

Basically I used @CWempe's template and I just added another ```UPSTREAM_NAME``` variable and then parse it. Should be around line 20.

```
{{ if .Container.Env.UPSTREAM_NAME }}
                        {{ if .Container.Env.VIRTUAL_PORT }}
                                server {{ .Container.Env.UPSTREAM_NAME }}:{{ .Container.Env.VIRTUAL_PORT }};
                        {{ else }}
                                server {{ .Container.Env.UPSTREAM_NAME }};
                                server {{ .Container.Env.UPSTREAM_NAME2 }};
                        {{ end }}
{{ else if .Network.IP }}
```

I am too lazy to learn how Go templating works. Maybe someone can add a check for UPSTREAM_NAMES if it is defined (similar to what we have now) and loop over it. Separated by space or comma.

In bash that would be something like:

```
UPSTREAMS="192.168.10.2 192.168.10.3"

for upstream in $UPSTREAMS; do
  echo server $upstream\;
done
```

mfenniak:
This PR looks really handy for this situation.  But in the absence of it, here's a workaround... create a docker container that forwards the network traffic to your off-docker service host.  I've accomplished this work a socat container like this:

```
docker create \
  --name=someservicename-forward \
  -p 4455:4455 \
  -e LETSENCRYPT_EMAIL=... \
  -e VIRTUAL_HOST=... \
  -e VIRTUAL_PORT=... \
  -e LETSENCRYPT_HOST=... \
  -e LETSENCRYPT_ACCOUNT_ALIAS=... \
  alpine/socat tcp-listen:4455,fork,reuseaddr tcp-connect:10.1.1.156:8123
```

nginx-proxy then treats this container as an upstream, which actually sends all network traffic to `10.1.1.156:8123` by a socat command.

In the absence of a merge of this PR, or similar functionality, this seemed like a nice 'compatible' way to get everything working without much hacking.

spfuture:
@mfenniak Your solution pretty, but have overhead at CPU load, app (alpine/socat) container, and traffic load.
Your traffic way scheme: nginx-proxy -> alpine/socat -> real_host_with_app

Solution by @CWempe  nginx-proxy ->  real_host_with_app its much better.

gustavogmoraes:
> In the meantime, [this project](https://github.com/mikesplain/nginx-proxy-pass-docker) does the job. However, I still think this should be a feature of jwilder/nginx-proxy.
> 
> If it could be so that we don't have to use a dummy container, that would be great (by maybe using an array of domains=>upstreams in an ENV var, or even a file somewhere. Just a thought ;).

Tried using the mikesplain/nginx-proxy-pass-docker but when I try to connect to another physical machine on my local network I get "*378 no live upstreams while connecting to upstream", any ideas?

bkraul:
@gustavogmoraes Are you putting the appropriate port to listen to the physical machine?

i.e.
```
docker run -d -p 80:80 -e TARGET_SERVER=192.168.8.15:8080 mikesplain/nginx-proxy-pass
```

buchdag:
Could the questions about mikesplain/nginx-proxy-pass-docker could be asked on this project repository rather than here ?

bkraul:
@buchdag sure. Or this PR could be merged and we wouldn't need to speak about the other project repository 🙂 ...

buchdag:
@bkraul while I understand this is a useful and wanted feature, it won't be implemented using zombie containers.

jolly-jump:
> @bkraul Both docker-gen and nginx-proxy stated scope was always to generated config file from / proxy to Docker container.
> 
> While I understand this is a useful and wanted feature, it won't be implemented using zombie containers.

Hi @buchdag , 
I don't understand why you call it "zombie containers"? You can use any container, it runs, maybe uses up some CPU-cycles initially and then does nothing. So what?
The upside is: I can use acme-companion to have valid certificates in no time and still forward any traffic to the not-so-intelligent service upstream.

Yes, if I just wanted the traffic to work, I guess that what @amosipov said: https://github.com/nginx-proxy/nginx-proxy/issues/998#issuecomment-460618515 would suffice. But then I don't have the comfort of automatic certificates...

Please reconsider!

PR #1106, opened 2018-03-22
do not create an empty upstream entry for invisible containers

### LINKED ISSUE #1105: nginx-proxy chokes on containers from invisible networks
## Scenario

* `nginx-proxy` is connected to some Docker network `reverseproxy`, but not connected to the default `bridge` network.
* A Docker container is started with a `VIRTUAL_HOST=foobar.example.com` environment setting and connected to the `bridge` network or some other network not connected to `nginx-proxy`.

## Expected behaviour

`nginx-proxy` ignores this container and proxies all others.

## Actual behaviour

`nginx-proxy` generates an nginx config file with invalid syntax
```
upstream foobar.example.com {
}

```
Due to this error, the proxy may fail to (re)start successfully and will not map any new containers.

## Notes

This is similar to but different from issue #845, i.e. that fix does not help in this scenario.

hwellmann:
Fixes #1105.

CAUTION: This is a monkey patch.... I don't fully understand the template, but at least this patch solved my problem.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -142,6 +142,9 @@ upstream {{ $upstream_name }} {
 					{{ $address := where $container.Addresses "Port" $port | first }}
 					{{ template "upstream" (dict "Container" $container "Address" $address "Network" $containerNetwork) }}
 				{{ end }}
+			{{ else }}
+				# Cannot connect to network of this container
+				server 127.0.0.1 down;
 			{{ end }}
 		{{ end }}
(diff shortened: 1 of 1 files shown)


jwilder:
Thanks @hwellmann!

jpic:
I don't see how adding an else to the if will prevent anything from happening.
It's actually the opposite: for any container that's not in the if, it will add a configuration line that will break the service.

Issue #503, opened 2016-07-11
Disable Logs


stevezau:
Hi There,

Do you have any way to disable the logging within nginx-proxy?

We use logging solutions to pull logs from all docker containers console logs. We don't want the nginx-proxy logs.

I would disable collecting logs from that container but the solution we are using does not have 'exclude' container filters, only include.

Last resort is for me to build a dockerfile based off your image?

philicious:
Same for me:

I would require the original nginx access/error logs to feed them into Graylog via GELF or syslog logging driver. However nginx-proxy outputs them to STDOUT but with color codes like 
```
[0;36;1mnginx.1 | [0macme.com 141.8.143.143 - - [12/Jul/2016:13:34:36 +0000] "GET /rew...
```
So there are several ways I could get them how I want them, however I havent found a way to disable the nginx-proxy output. (which also includes the docker-gen output)

If theres no simple config to disable the output, I'd appreciate a hint on code where I could do it "manually"

wader:
Maybe https://github.com/kitpages/nginx-proxy/pull/1 is interesting? hopefully only thing needed is to also /dev/null docker-gen

philicious:
thx for the hint  @wader 

I also meanwhile found #165 which is quite similar and the suggested solution of running nginx/docker-gen in separate containers and thereby avoiding forego, seemed to work for that guy. 
So I'll try that.

I'm not sure if that would help @stevezau tho

afefelov:
Hi guys, I'm using graylog and rsyslog to pass logs into it.
I spent up to 4 hours to find out proper solution and finally came with following:

You should add rsyslog conf file like that /etc/rsyslog.d/30-docker.conf:

```
module(load="mmexternal")
if $syslogtag contains 'nginx' then action(type="mmexternal" binary="/usr/bin/strip_ansi_colors.py" interface.input="msg" )
```

while /usr/bin/strip_ansi_colors.py looks like:

```
#! /usr/bin/python

import sys
import re
import json

def onInit():
	""" Do everything that is needed to initialize processing (e.g.
	    open files, create handles, connect to systems...)
	"""
	pass


def onReceive(msg):
	"""This is the entry point where actual work needs to be done. It receives
	   the messge from rsyslog and now needs to examine it, do any processing
	   necessary. The to-be-modified properties (one or many) need to be pushed
	   back to stdout, in JSON format, with no interim line breaks and a line
	   break at the end of the JSON. If no field is to be modified, empty
[...]

PR #1113, opened 2018-03-28
Set DISABLE_ACCESS_LOGS env var to disable access logs

basro:
This PR removes the access_log lines from the generated nginx config if DISABLE_ACCESS_LOGS env var is set in the nginx-proxy container.

I think this would solve issue #503

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -88,6 +88,8 @@ proxy_set_header X-Forwarded-Port $proxy_x_forwarded_port;
 proxy_set_header Proxy "";
 {{ end }}
 
+{{ $access_log := (or (and (not $.Env.DISABLE_ACCESS_LOGS) "access_log /var/log/nginx/access.log vhost;") "") }}
+
 {{ $enable_ipv6 := eq (or ($.Env.ENABLE_IPV6) "") "true" }}
 server {
 	server_name _; # This is just an invalid value which will never trigger on a real hostname.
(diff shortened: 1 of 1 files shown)


Fohlen:
Can this be brought forward @jwilder? It is also a nice way to disable the default logging for an alternate logging format, thus making it possible to comply with the DGPR that's coming soon 👍

krivochenko:
I am also interested in these changes, because there are containers for which I do not want to log requests.

Slayer911:
good idea, but for a year there has been no solution

tjhiggins:
+1

tjhiggins:
@jwilder Thank you - that was amazing! How often do you update the docker tag?

gaastonsr:
Agreed, this is awesome, I would appreciate an updated image 🙏🏽. Thanks

modmoto:
would totally like to update to this version aswell =D

throny:
As this is already merged to the master branch, when will it be included in a new image?
Tried setting the env var in the desired container and in the nginx-container itself: neither does work. Would really love this feature!

tkw1536:
This should be available on the main branch and current images since [0.8.0](https://github.com/nginx-proxy/nginx-proxy/releases/tag/0.8.0). 

@throny  Can you try setting `DISABLE_ACCESS_LOGS=1` (as opposed to without a value)?

throny:
That did the trick! Thank you @tkw1536 ! 
However it only works for the nginx-container itself, not for the docker containers behind the proxy right?

tkw1536:
Correct, `nginx-proxy` cannot influence logging behavior of other docker containers.

PR #1116, opened 2018-03-30
fix fastcgi bug

qiqizjl:
fix fastcgi bug for debian

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -278,7 +278,7 @@ server {
 		uwsgi_pass {{ trim $proto }}://{{ trim $upstream_name }};
 		{{ else if eq $proto "fastcgi" }}
 		root   {{ trim $vhost_root }};
-		include fastcgi.conf;
+		include fastcgi_params;
 		fastcgi_pass {{ trim $upstream_name }};
 		{{ else }}
 		proxy_pass {{ trim $proto }}://{{ trim $upstream_name }};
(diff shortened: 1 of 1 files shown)


controlcde:
I think you missed the same entry in the line 328.

qiqizjl:
@controlcde   Thanks, Fixed

qiqizjl:
@gboor Because fastcgi.conf is not included in the nginx image, fastcgi.conf is included in alpine.

gboor:
So the alpine version works?

If that is true, I think the Dockerfiles for Debian need to change so that another tmpl can be injected.

qiqizjl:
@gboor  yes.alpine version is ok .  If you want to use Debian, you can make corresponding changes based on the current PR. I think it can be normal operation.

BlackSmith:
Hi folks, 
any progress with this problem?

G1K:
very annoying bug. @jwilder accept the correction please
@qiqizjl look at the build errors please

qiqizjl:
@G1K Fixed in #1124

SerikK:
@qiqizjl index.php not opening by default

TheLastProject:
Any reason this hasn't been merged yet?

jwilder:
@TheLastProject  CI checks failed.  I kicked them off again, but PR might need to be rebased.

Issue #1108, opened 2018-03-25
More control of the cipher suites used


Archcry:
Currently you can only use one of the presets, it would be nice to create your own custom cipher suite and define your own tls version (especially with TLS1.3 being available now).

I suggest adding two new environment variables for this:
`SSL_CIPHER_SUITE` in which you specify the Cipher Suite
`SSL_PROTOCOL` in which you specify your SSL protocol (defaults to TLS 1.0, 1.1 and 1.2, like mozilla intermediate)

Only when `SSL_CIPHER_SUITE` is specified we override the `SSL_POLICY` environment variable, because the `SSL_PROTOCOL` has a default we don't have to specify this one explicitly (and the ssl_protocol specified by `SSL_POLICY` is used when `SSL_CIPHER_SUITE` is not set). When nothing is specified be just fall back to the default Mozilla intermediate policy.

I might be able to make a pul-request about this, just tell me what you think of the idea.

Archcry:
Just giving a casual bump, @kamermans could you look into my suggestion :)

kamermans:
Hi @Archcry, thanks for the bump!  I did actually [suggest this as a possibility](https://github.com/jwilder/nginx-proxy/pull/957#issuecomment-342336668) when we implemented it last year, but we (@buchdag and @jwilder) agreed to use the more simplistic `SSL_POLICY` instead.  One of the main reasons I suggested avoiding a custom protocol/cipher suite is that the HSTS policy, DHParam length, OCSP Stapling, and possibly other features are also sort-of dependent on the policy you choose, so the user would need to get all of those settings right in order to have consistent SSL/TLS policy.  I would recommend you read through that thread to see how we came about the decision.  My general feeling is that we should continue to update the policies that we have explicit support for (the [Mozilla TLS Policies](https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility), for example.  @jwilder said something similar in the original thread back in November/December.

I think we should definitely add TLS 1.3 to the list of supported protocols for all the policies since they all should support the latest version of TLS.

Archcry:
Pull-request is merged now, please note that the alpine version of the nginx container is not up-to-date to support TLSv1.3 because alpine hasn't updated to a newer version (`>1.1.1`) of `openssl` yet.

Abdedine:
Hi

I've declared in my docker-compose those env. variables but TLS 1.0 is still enabled :(

[...]

PR #1123, opened 2018-04-22
Add TLSv1.3 support

kamermans:
As noted by @Archcry in #1108, `nginx-proxy` does not support TLSv1.3.  I'm not 100% sure of the status of TLSv1.3, but it appears to be an approved draft, which means it will be finalized eventually.  **This PR adds support for TLSv1.3**

Nginx [added support for TLS 1.3 in `1.13.0`](http://mailman.nginx.org/pipermail/nginx-announce/2017/000195.html), so it is available to `nginx-proxy` today.  The included `SSL_POLICY` options are a reflection of the policies provided by Mozilla and AWS.  Presumably, they will both add TLS 1.3 to all of their policies, although I suppose they could decide not to include it for unknown reasons.  It appears they are [waiting for the draft to be finalized](https://github.com/awslabs/s2n/issues/388) before announcing support.

I think it is reasonable for us to add support already today since all of our SSL/TLS policies include the latest version of TLS, and it's hard to imagine why we would not want people to be able to use it.


Archcry:
Any progress on this?

kamermans:
It's finished and ready to go.  Do you want to include TLSv1.3 support @jwilder?

Archcry:
I know I do :)

Archcry:
@jwilder, can we merge this?

PR #1120, opened 2018-04-11
Template code formatting

atnartur:
Added some template code formatting

PR #1083, opened 2018-03-03
Allow VIRTUAL_PATH routing (sub-VIRTUAL_HOST)

rodrigoaguilera:
Just a rebase of https://github.com/jwilder/nginx-proxy/pull/1011
and some fixes to check if the tests pass.

PR #1124, opened 2018-04-22
Bugfix: Test reads from out-of-scope var

kamermans:
I don't know why this only seems to fail in Docker `18.04 RC3`, but I can confirm that it consistently fails in that version and not in older versions.  This actually fixes two bugs:
1. In `docker-compose 1.11.1`, the code pulls some internal code from `pip`, but `pip` removed that code and so `docker-compose` broke.  This was fixed in `docker-compose 1.11.2`, so I've upgraded that, and had to upgrade `docker-py` as well, to meet the minimum requirements.
2. In one of our tests, we were opening a file handle via `with` and reading from it after the `with` was closed, so the variable was out of scope.  Fixed with a single indent.

@jwilder This PR will fix the failing builds in PRs #1123, #1083, #1116, #1120.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,8 @@ env:
   - TEST_TARGET: test-alpine
 
 before_install:
-  - sudo apt-get remove docker docker-engine
+  - sudo apt-get -y remove docker docker-engine docker-ce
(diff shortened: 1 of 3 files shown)


jwilder:
Thanks!

Issue #1017, opened 2018-01-04
Tag / version for Raspeberry Pi 3 & Alpine. 


IoTPlay:
Hi, could you add a tag for raspberry Pi 3 on Alpine please?

Jéan Roux.

buchdag:
Fixed as of #1470 + 4a1d5bfc58af35f5e90dcefb47b35a274c903150 + d120ce9fdebf09a6c7b90ecff7eda0b7d2760ef3

`arm64` and `arm/v7` images are availables from `jwilder/nginx-proxy` or the new registry `nginxproxy/nginx-proxy`

PR #1130, opened 2018-05-03
feat: Add rpi version

Javrd:
This PR adds a dockerfile version for rpi from arm32v7 nginx image
It's also use a new ENV var (ARCH_VERSION=armhf) to download correct version of docker-gen and build forego from source in order to build it in the correct architecture.

Related with #1017


buchdag:
Superseded by #1470

PR #1135, opened 2018-05-24
Adding the possibility to use anonymized logging

Mactory:
If the env var `ANONYMIZE_LOGGING` is set, the logging output of nginx is changed to anonymize the IP-Address and remove the referer and user agent.


Fohlen:
One could have referenced that this is the code from https://stackoverflow.com/a/27749834/3135319
Looks sane for simple anonymization cases.

Mactory:
You're right. I added the reference. I also merged upstream.

Issue #1136, opened 2018-05-26
vhost.d location config disables HSTS


grmat:
# Description:

Setting an `/etc/nginx/vhost.d/{VIRTUAL_HOST}_location` or `default_location` config overwrites the `location{}` block with the effect of disabling HSTS.

# Steps to reproduce:

1. Create a location config file per vhost or a `default_location`, mine looks as follows:

       add_header Referrer-Policy "no-referrer";

1. restart nginx-proxy
1. check that HSTS is disabled

# Workaround

Add `add_header Strict-Transport-Security "max-age=31536000" always;` to your per-vhost or default location config.

# Conclusion

It is not documented that HSTS is disabled with location configs, and at least I expected them to be complementary. Therefore, I'd suggest to add the workaround to the docs or see if it is possible to make both, the default HSTS config and the per-vhost/default location config complementary.

sgabe:
This works as designed, due to the fact that `add_header` directives are inherited from the previous level if and only if there are no `add_header` directives defined on the current level. Please refer to the [ngx_http_headers_module](https://nginx.org/en/docs/http/ngx_http_headers_module.html) documentation for more information.

PR #1137, opened 2018-05-26
update README.md: add workaround for #1136

grmat:


sgabe:
Kindly note that the proposed change does not accurately reflect the actual issue as you described in #1136 . Defining `add_header` directives on the location level will stop inheriting from previous levels. This is how Nginx works.

buchdag:
This won't be merged for the reasons given by @sgabe

PR #1154, opened 2018-07-23
Added client_max_body_size

ipepe:
I think that this nginx proxy configuration should be as transparent as possible. And not setting `client_max_body_size` leaves it on default with `1m` value. I suggest that it should be `client_max_body_size 0;`

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -205,6 +205,7 @@ server {
 server {
 	server_name {{ $host }};
 	listen 443 ssl http2 {{ $default_server }};
+	client_max_body_size 0;
 	{{ if $enable_ipv6 }}
 	listen [::]:443 ssl http2 {{ $default_server }};
 	{{ end }}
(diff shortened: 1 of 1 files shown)


delfer:
LGTM

indapublic:
Actual for me

ipepe:
I recommend using Traefik @indapublic and @delfer It's next level solution

indapublic:
Thanks @ipepe. Will look

azlux:
You should not disable it with setting up to zero, you're more exposed to ddos attack.
But a en variable making this value flexible should be better. I currently setting-up SOME backend with higher body_size for upload purpose, but never disabled.

selckin:
Better option to make it configurable in https://github.com/nginx-proxy/nginx-proxy/issues/981#issuecomment-1152924213

Prime1Code:
I agree with @azlux 
Make the config setting an env prop. 0 is unsafe in production.

buchdag:
This PR won't be merged because it introduces an unsafe default configuration.

Furthermore, even if it's not as dedicated and straightforward as a dedicated environment variable, nginx-proxy already has builtin capabilities to configure this, as remarked by @Prime1Code .

Issue #1138, opened 2018-05-29
Expose different containers on different ports of the url


jamiya9456:
I have a php based site say example.com.

I would like to serve the site on the default port: 80 - > example.com / example.com:80 -> Normal site displayed.
For phpMyAdmin I would like to use port: 8080 -> example.com:8080 -> phpMyAdmin displayed from port 80 of the pma container.

I tried tweaking the `/path/to/vhost.d/example.com` and exapmple.com_localtion

patrickqsos:
If you want to use your phmyadmin container in a different port than 80 then I think you don't need nginx proxy, just run the phmyadmin container as usual with `$ docker run -p 8080:80 ...` and your php app `$ docker run -p 80:80 ...` 

A better approach with nginx proxy would be like this:

Run your php app in example.com
 `$ docker run -e VIRTUAL_HOST=example.com  ...`

Run your phpmyadmin in phpmyadmin.example.com 
`$ docker run -e VIRTUAL_HOST=phpmyadmin.example.com  ...`

PR #1159, opened 2018-08-13
Feature: Listen Ports for HTTP/HTTPS

pauldorn:
This simple/backwards-compatible change lets the user specify alternatives to 80 and 443 for nginx to listen.  
It solves my issue (needing to listen on non-standard ports for both http/https)
This may also be a solution for #1138  

Documentation pull request for proxy companion pending...


pauldorn:
How do you handle out of sync changes?  That is, the documentation should probably not be updated on the companion until the images are updated on dockerhub, yes?

pauldorn:
I will replace this pull request with one that does not break the build.

PR #1170, opened 2018-09-19
allow one .htpasswd file for all VIRTUAL_HOST’s

mark-wff:
I use the nginx-proxy for my dev-Server. So I need a http-password for every vHost. To keep it simple I added the functionality to add a single `/etc/nginx/htpasswd/.htpasswd` file for all vHosts. This file can be overwritten by a specific File per vHost `/etc/nginx/htpasswd/$VIRTUAL_HOST`.


thiagotalma:
Please close.

PR #1176, opened 2018-10-17
disable basic authentication for HTTP OPTIONS for CORS

rparree:
This is to make CORS to work together with basic authentication. OPTIONS should not be restricted as browsers don't send the auth bearer for pre-flight requests

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -285,8 +285,10 @@ server {
 		{{ end }}
 
 		{{ if (exists (printf "/etc/nginx/htpasswd/%s" $host)) }}
-		auth_basic	"Restricted {{ $host }}";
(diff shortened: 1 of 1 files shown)


vemonet:
Is there any plan on accepting this fix? We are having the same issue to use CORS + basic nginx Auth

tkw1536:
I'm not sure this should be accepted. 

When configuring http basic authentication, by default I would expect *all* requests to only be forwarded after authentication; that way no data can be accidentally leaked to unauthenticated clients.

IMO if any request is excluded from this, it should at least be documented in the README. Perhaps it should also be hidden behind a feature flag / environment variable.

vemonet:
Indeed @tkw1536 it would make sense to make this disabled by default and only enabled if explicitly requested

Still interested @rparree ? I'll take a look when I have some time otherwise

PR #1179, opened 2018-10-18
Allow complete override of location blocks

### LINKED ISSUE #567: Support for VIRTUAL_PROTO=custom
In some special conditions, we may need to configure backends other than proxy_pass and uwsgi_pass (for example  fastcgi_pass, or serving static files). 
- As a solution, we can allow new env setting VIRTUAL_PROTO=custom,
  - If specified,  nginx-proxy should not set any backend for domain, and it should just leave that to user, and user should provide the backend configuration using {host}_location configuration file.
  - So any backend other than proxy_pass and uwsgi_pass are possible now.
- This change wont break any existing usage.
- I understand nginx-proxy was not created to support variety of backends, but by just allowing custom VIRTUAL_PROTO, users can tweak it for some complex requirement.
- I'll send pull request within an hour.

instead of 

```
{{ if eq $proto "uwsgi" }}
include uwsgi_params;
uwsgi_pass {{ trim $proto }}://{{ trim $host }};
{{ else }}
proxy_pass {{ trim $proto }}://{{ trim $host }};
{{ end }}
```

we can say,

```

{{ if eq $proto "custom" }}
{{ else if eq $proto "uwsgi" }}
include uwsgi_params;
uwsgi_pass {{ trim $proto }}://{{ trim $host }};
{{ else }}
proxy_pass {{ trim $proto }}://{{ trim $host }};
{{ end }}
```

https://github.com/jwilder/nginx-proxy/blob/ec7169c112db37f2869dc264ddda5854c9716bcc/nginx.tmpl#L171-L176
https://github.com/jwilder/nginx-proxy/blob/ec7169c112db37f2869dc264ddda5854c9716bcc/nginx.tmpl#L205-L210

### LINKED ISSUE #1003: Custom Locations
#So, I can re-write the vhost config are I can and settings to the root location.  However, how can I add settings for additional locations?

The specific use case here is using Collabora CODE with Nextcloud.  The required config can be found here: https://icewind.nl/entry/collabora-online/

Is there a way to add addition location based settings without doing a completely custom vhost config?

### LINKED ISSUE #1086: Add the ability to have no location for some virtual hosts
Since there is the ability to provide custom configuration for each vhost there is the possibility that a user is interested in rewriting the "location /".

I currently this modification and a random path:
https://github.com/jwilder/nginx-proxy/pull/1083

I am injecting this configuration for my vhost
https://docs.nextcloud.com/server/13/admin_manual/installation/nginx.html

Another fair case might be to add just a redirect for the "location /" instead of having a container just for the redirect.

I suggest the env VIRTUAL_PATH (from the PR I link) but the possibility of setting it to "none" so it does not print any location in the nginx configuration.

### LINKED ISSUE #1286: Implement rules for location /
I want to implement some custom nginx rules for location / but if I write this rule using vhost.de/domainname.conf I get that there are duplicate rules for / location, because this rule doesn't replace the existing rule in default.conf, but adds instead.

I have been looking into plenty of ways to do this. I have overwritten the rule in default.conf myself and it works if I run

`docker exec -it nginx-proxy nginx -s reload`

Restarting the container will make my change get just lost since docker-gen reconstructs this "default.conf". I think there is no current way to overwrite rules for / location. Is there any chance this gets implemented any time soon?

### LINKED ISSUE #1385: Pre-pending instead of appending to location block?
I am aware of https://github.com/jwilder/nginx-proxy/issues/1304, and also aware that the google group is not working. This is a combination of a question and a feature request.

In order to implement oauth2, I need my location block to look like this:-

    auth_request /oauth2/auth;
    error_page 401 = /oauth2/sign_in;

    proxy_pass http://web-server:80;

    proxy_set_header Host $host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Scheme $scheme;
    proxy_http_version 1.1;
    proxy_connect_timeout 300;
    proxy_send_timeout 30;
    proxy_read_timeout 300;
    proxy_set_header Connection "";

However if I use the current location drop-in (/etc/vhost.d/web-server_location) the content goes after proxy_pass. If I use the server drop-in (write the entire location block in /etc/vhost.d/web-server) then I have two root location blocks, which also does not work.

What's the right way to do this, or could there be an option to NOT generate the location block just for this host?

harvdogg:
Under the existing implementation, it would be impossible to change the actual behavior of the default location block, which attempts to proxy_pass onto the container.  This is probably acceptable for many use cases but not all.

This change adds some additional template logic that considers the existence of a `VIRTUAL_HOST_locations` (plural "locations" as opposed to the otherwise considered "location") configuration file.  When present, it will bypass generating any nginx `location` blocks and instead use those provided by the configuration file.  

This allows for some really advanced use cases with nginx's location filtering.  You can then forward the root location of a monitored domain onto a completely different container or even an entirely external host while still introducing logic that allows certain locations to forward to the container like normal.  

The README has been updated to reflect this. 

Fixes #1385
Fixes #1286 
Fixes #1086
Fixes #1003 
Fixes #567

### DIFF
--- a/README.md
+++ b/README.md
@@ -491,6 +491,32 @@ ln -s /path/to/vhost.d/www.example.com /path/to/vhost.d/example.com
 
 If you want most of your virtual hosts to use a default single `location` block configuration and then override on a few specific ones, add those settings to the `/etc/nginx/vhost.d/default_location` file. This file will be used on any virtual host which does not have a `/etc/nginx/vhost.d/{VIRTUAL_HOST}_location` file associated with it.
 
+#### Overriding `location` blocks
+
+The `${VIRTUAL_HOST}_${PATH_HASH}_location`, `${VIRTUAL_HOST}_location`, and `default_location` files documented above make it possible to *augment* the generated [`location` block(s)](https://nginx.org/en/docs/http/ngx_http_core_module.html#location) in a virtual host.  In some circumstances, you may need to *completely override* the `location` block for a particular combination of virtual host and path.  To do this, create a file whose name follows this pattern:
+
+```
+/etc/nginx/vhost.d/${VIRTUAL_HOST}_${PATH_HASH}_location_override
+```
+
+where `${VIRTUAL_HOST}` is the name of the virtual host (the `VIRTUAL_HOST` environment variable) and `${PATH_HASH}` is the SHA-1 hash of the path, as [described above](#per-virtual_path-location-configuration).
+
+For convenience, the `_${PATH_HASH}` part can be omitted if the path is `/`:
+
+```
+/etc/nginx/vhost.d/${VIRTUAL_HOST}_location_override
+```
+
+When an override file exists, the `location` block that is normally created by `nginx-proxy` is not generated.  Instead, the override file is included via the [nginx `include` directive](https://nginx.org/en/docs/ngx_core_module.html#include).
+
+You are responsible for providing a suitable `location` block in your override file as required for your service.  By default, `nginx-proxy` uses the `VIRTUAL_HOST` name as the upstream name for your application's Docker container; see [here](#unhashed-vs-sha1-upstream-names) for details.  As an example, if your container has a `VIRTUAL_HOST` value of `app.example.com`, then to override the location block for `/` you would create a file named `/etc/nginx/vhost.d/app.example.com_location_override` that contains something like this:
+
+```
+location / {
+    proxy_pass http://app.example.com;
+}
+```
+
 #### Per-VIRTUAL_HOST `server_tokens` configuration
 Per virtual-host `servers_tokens` directive can be configured by passing appropriate value to the `SERVER_TOKENS` environment variable. Please see the [nginx http_core module configuration](https://nginx.org/en/docs/http/ngx_http_core_module.html#server_tokens) for more details.
 
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -76,38 +76,46 @@
 {{- end }}
 
 {{- define "location" }}
+    {{- $override := printf "/etc/nginx/vhost.d/%s_%s_location_override" .Host (sha1 .Path) }}
+    {{- if and (eq .Path "/") (not (exists $override)) }}
+        {{- $override = printf "/etc/nginx/vhost.d/%s_location_override" .Host }}
+    {{- end }}
+    {{- if exists $override }}
+    include {{ $override }};
+    {{- else }}
     location {{ .Path }} {
-    {{- if eq .NetworkTag "internal" }}
+        {{- if eq .NetworkTag "internal" }}
         # Only allow traffic from internal clients
         include /etc/nginx/network_internal.conf;
-    {{- end }}
+        {{- end }}
 
-    {{- if eq .Proto "uwsgi" }}
+        {{- if eq .Proto "uwsgi" }}
         include uwsgi_params;
         uwsgi_pass {{ trim .Proto }}://{{ trim .Upstream }};
-    {{- else if eq .Proto "fastcgi" }}
+        {{- else if eq .Proto "fastcgi" }}
         root {{ trim .VhostRoot }};
         include fastcgi_params;
         fastcgi_pass {{ trim .Upstream }};
-    {{- else if eq .Proto "grpc" }}
+        {{- else if eq .Proto "grpc" }}
         grpc_pass {{ trim .Proto }}://{{ trim .Upstream }};
-    {{- else }}
+        {{- else }}
         proxy_pass {{ trim .Proto }}://{{ trim .Upstream }}{{ trim .Dest }};
-    {{- end }}
+        {{- end }}
 
-    {{- if (exists (printf "/etc/nginx/htpasswd/%s" .Host)) }}
+        {{- if (exists (printf "/etc/nginx/htpasswd/%s" .Host)) }}
         auth_basic "Restricted {{ .Host }}";
         auth_basic_user_file {{ (printf "/etc/nginx/htpasswd/%s" .Host) }};
-    {{- end }}
+        {{- end }}
 
-    {{- if (exists (printf "/etc/nginx/vhost.d/%s_%s_location" .Host (sha1 .Path) )) }}
+        {{- if (exists (printf "/etc/nginx/vhost.d/%s_%s_location" .Host (sha1 .Path) )) }}
         include {{ printf "/etc/nginx/vhost.d/%s_%s_location" .Host (sha1 .Path) }};
-    {{- else if (exists (printf "/etc/nginx/vhost.d/%s_location" .Host)) }}
+        {{- else if (exists (printf "/etc/nginx/vhost.d/%s_location" .Host)) }}
         include {{ printf "/etc/nginx/vhost.d/%s_location" .Host}};
-    {{- else if (exists "/etc/nginx/vhost.d/default_location") }}
+        {{- else if (exists "/etc/nginx/vhost.d/default_location") }}
         include /etc/nginx/vhost.d/default_location;
-    {{- end }}
+        {{- end }}
     }
+    {{- end }}
 {{- end }}
 
 {{- define "upstream" }}
--- a/test/test_location-override.py
+++ b/test/test_location-override.py
@@ -1,0 +1,39 @@
+def test_explicit_root_nohash(docker_compose, nginxproxy):
+    r = nginxproxy.get("http://explicit-root-nohash.nginx-proxy.test/port")
+    assert r.status_code == 418
+    r = nginxproxy.get("http://explicit-root-nohash.nginx-proxy.test/foo/port")
+    assert r.status_code == 200
+    assert r.text == "answer from port 82\n"
+
+def test_explicit_root_hash(docker_compose, nginxproxy):
+    r = nginxproxy.get("http://explicit-root-hash.nginx-proxy.test/port")
+    assert r.status_code == 418
+    r = nginxproxy.get("http://explicit-root-hash.nginx-proxy.test/foo/port")
+    assert r.status_code == 200
+    assert r.text == "answer from port 82\n"
+
+def test_explicit_root_hash_and_nohash(docker_compose, nginxproxy):
+    r = nginxproxy.get("http://explicit-root-hash-and-nohash.nginx-proxy.test/port")
+    assert r.status_code == 418
+    r = nginxproxy.get("http://explicit-root-hash-and-nohash.nginx-proxy.test/foo/port")
+    assert r.status_code == 200
+    assert r.text == "answer from port 82\n"
+
+def test_explicit_nonroot(docker_compose, nginxproxy):
+    r = nginxproxy.get("http://explicit-nonroot.nginx-proxy.test/port")
+    assert r.status_code == 200
+    assert r.text == "answer from port 81\n"
+    r = nginxproxy.get("http://explicit-nonroot.nginx-proxy.test/foo/port")
+    assert r.status_code == 418
+
+def test_implicit_root_nohash(docker_compose, nginxproxy):
+    r = nginxproxy.get("http://implicit-root-nohash.nginx-proxy.test/port")
+    assert r.status_code == 418
+
+def test_implicit_root_hash(docker_compose, nginxproxy):
+    r = nginxproxy.get("http://implicit-root-hash.nginx-proxy.test/port")
+    assert r.status_code == 418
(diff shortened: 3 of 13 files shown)


Divxtaman:
This is a must-have, please merge @jwilder

liudonghua123:
This is a must-have, please merge @jwilder

DominMuda:
It would be awesome if we could have this feature in master 🙏 @jwilder

sl45sms:
+1

thg303:
any idea why this is not merged yet?? :(

ImadSalimi:
Read the text here, it seems to be merged but this doesn't show on the docs. I have yet to test it
https://github.com/gudilap/nginx-proxy/pull/1/commits/c984ed2b18010db70ca2926967f8fc1e95d03a30

roysG:
any update?

Gymnae:
Please merge this

ngoonee (dismissed):
This would solve the issue I raise in https://github.com/jwilder/nginx-proxy/issues/1385. Testing on my own docker build here works fine (the resulting conf has some indentation issues though).

Zundrium:
2020 now, would be nice if this could be merged :) have to change the default.conf after each restart at the moment.

cmargroff:
Whats the hold up here?
I really need this feature to be able to replace the default location / {} block.

r-t-k:
Why hasn't this been merged?

banagale:
Thanks for this PR. I suspect the hold up is that this just takes time to deal with. :)

That said, I could also use this PR. Django complains bitterly about `[Django] ERROR (EXTERNAL IP): Invalid HTTP_HOST` if you do not specifically disallow requests from domains outside the `ALLOWED_HOST` configuration.

The current build of this repo sends requests for the server IP to the [first server block](https://github.com/jwilder/docker-gen/blob/master/templates/nginx.tmpl#L1) which results in 503's bubbling up to admin emails / sentry etc.  This is a real problem on a VPS host where IP ranges are constantly probed.

This is a great container. @jwilder, when time allows please consider merging this.

shoff:
> Why hasn't this been merged?

Indeed why has this not been merged?

buchdag:
@rhansen I'm considering merging this if we can test it.

I'm just not entirely sold on the `VIRTUAL_HOST_locations` vs existing `VIRTUAL_HOST_location` file name as I think it lacks explicitness. Maybe `VIRTUAL_HOST_root_location` ?

rhansen:
I'm also meh on the `_locations` name mostly because it's not visually very distinct so it's easy to misread.  Variants of singular `location` isn't really right because this feature replaces all `location` blocks.   Would something like `_all_locations` or `_all_locations_override` be better?

If I was a maintainer I could rebase this onto latest `main` and add tests...  :wink:

rhansen:
Thinking about this more, users of this feature are unlikely to use `VIRTUAL_PATH`, right?  So most of the time this would only override a single `location` block.  So maybe pluralized isn't the way to go.  How about `*_location_override`?

Users can already inject `server`-scoped config via `/etc/nginx/vhost.d/VIRTUAL_HOST` or `/etc/nginx/vhost.d/default`, so really all we need is a boolean that disables the `location` blocks.  Right?

If the user does use `VIRTUAL_PATH`, then the override/boolean would be overriding/disabling multiple `location` blocks, which might be surprising.  We could error out in that case.  If requested, we could add per-path overrides, like this:

```patch
diff --git a/nginx.tmpl b/nginx.tmpl
index e7d77c9..ddfb3f5 100644
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -52,6 +52,10 @@
 {{ end }}
 
 {{ define "location" }}
+    {{- $override := printf "/etc/nginx/vhost.d/%s_%s_location_override" .Host (sha1 .Path) }}
+    {{- if $override }}
+        include {{ $override }}
+    {{- else }}
 	location {{ .Path }} {
 	{{ if eq .NetworkTag "internal" }}
 		# Only allow traffic from internal clients
@@ -83,6 +87,7 @@
 	{{ else if (exists "/etc/nginx/vhost.d/default_location") }}
 	{{ end }}
+    {{- end }}
 }
 {{ end }}
 
```

rhansen:
I rebased onto latest `main` and cleaned up some newlines in `README.md`.  No other changes were made.

rhansen:
I pushed a couple of fixes (typo fix, move documentation) as well as a commit that changes the approach to be per-location instead of per-host.  Please let me know what you think; I'll add tests if there is consensus.

buchdag:
@rhansen I've given that a bit more thought, do you think an environment variable like `VIRTUAL_ROOT_LOCATION_BYPASS` or `VIRTUAL_ROOT_LOCATION_DISABLE` would be better ? What we actually need here is a way to disable the default root location, the existing custom location mechanism can handle the root location once the default one has been disabled.

rhansen:
> What we actually need here is a way to disable the default root location

Is it?  The problem we are trying to solve here is not clear to me.  The `VIRTUAL_PATH` feature did not exist when this PR was opened, so maybe the original problem has been largely mitigated and we are trying to solve something else?

If we only want to disable the default root location, then I propose extending the existing `DEFAULT_ROOT` environment variable instead:

```patch
diff --git a/README.md b/README.md
index 0ca08d7..a656637 100644
--- a/README.md
+++ b/README.md
@@ -152,10 +152,17 @@ The filename of the previous example would be `example.tld_8610f6c344b4096614eab
 
 This environment variable of the nginx proxy container can be used to customize the return error page if no matching path is found. Furthermore it is possible to use anything which is compatible with the `return` statement of nginx.
 
-For example `DEFAULT_ROOT=418` will return a 418 error page instead of the normal 404 one.
-Another example is `DEFAULT_ROOT="301 https://github.com/nginx-proxy/nginx-proxy/blob/main/README.md"` which would redirect an invalid request to this documentation.
-Nginx variables such as $scheme, $host, and $request_uri can be used. However, care must be taken to make sure the $ signs are escaped properly.
-If you want to use `301 $scheme://$host/myapp1$request_uri` you should use:
+Exception:  If this is set to the string `none`, no default `location /` directive will be generated.  This makes it possible for you to provide your own `location /` directive in your [`/etc/nginx/vhost.d/VIRTUAL_HOST`](#per-virtual_host) or [`/etc/nginx/vhost.d/default`](#per-virtual_host-default-configuration) files.
+
+If unspecified, `DEFAULT_ROOT` defaults to `404`.
+
+Examples (YAML syntax):
+
+  * `DEFAULT_ROOT: "none"` prevents `nginx-proxy` from generating a default `location /` directive.
+  * `DEFAULT_ROOT: "418"` returns a 418 error page instead of the normal 404 one.
+  * `DEFAULT_ROOT: "301 https://github.com/nginx-proxy/nginx-proxy/blob/main/README.md"` redirects the client to this documentation.
+
+Nginx variables such as `$scheme`, `$host`, and `$request_uri` can be used.  However, care must be taken to make sure the `$` signs are escaped properly.  For example, if you want to use `301 $scheme://$host/myapp1$request_uri` you should use:
 
 * Bash: `DEFAULT_ROOT='301 $scheme://$host/myapp1$request_uri'`
 * Docker Compose yaml: `- DEFAULT_ROOT: 301 $$scheme://$$host/myapp1$$request_uri`
diff --git a/nginx.tmpl b/nginx.tmpl
index fb4f76f..64a3edb 100644
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -453,7 +453,7 @@ server {
         {{- end }}
         {{- template "location" (dict "Path" $path "Proto" $proto "Upstream" $upstream "Host" $host "VhostRoot" $vhost_root "Dest" $dest "NetworkTag" $network_tag) }}
     {{- end }}
-    {{- if (not (contains $paths "/")) }}
+    {{- if and (not (contains $paths "/")) (ne $globals.default_root_response "none")}}
     location / {
         return {{ $globals.default_root_response }};
     }
```

buchdag:
> The problem we are trying to solve here is not clear to me.

From what I understood of the various issues mentioned, the problem is that nginx-proxy generate a `location / { [...] }` block no matter what you do with the per-location config files.  Some setup need to entirely bypass this default root location and provide a fully custom one but that's not possible at the moment because the default always generated one will always bypass whatever the user provide. IMHO the problem should be considered the same wether you use `VIRTUAL_PATH` without a container handling the root path or not.

> If we only want to disable the default root location, then I propose extending the existing `DEFAULT_ROOT` environment variable instead:

I really like this idea, using `DEFAULT_ROOT=none` to achieve this seems the best way to implement this.

rhansen:
> From what I understood of the various issues mentioned, the problem is that nginx-proxy generate a `location / { [...] }` block no matter what you do with the per-location config files. Some setup need to entirely bypass this default root location and provide a fully custom one but that's not possible at the moment because the default always generated one will always bypass whatever the user provide. IMHO the problem should be considered the same whether you use `VIRTUAL_PATH` without a container handling the root path or not.

I see a mix.  Some users seem to want to be able to disable the default root and some want to tweak the `location /` block generated by `VIRTUAL_PATH=/` in a way that is not currently supported by the `*_location` include file.  Specifically:

> #1554

I believe this can be fixed by setting `VIRTUAL_PATH: '~ \.php$$'` (Docker compose YAML syntax) and disabling the default root block, so the `DEFAULT_ROOT=none` proposal would be an adequate solution here.

> #1385

For this case, the `DEFAULT_ROOT=none` proposal would not be sufficient because the implicit `VIRTUAL_PATH=/` suppresses the default block anyway.

I guess this user could do the following:

  1. Set `VIRTUAL_PATH: '~ ^\b$'` (where `^\b$` is a regular expression that will never match anything) to move the `proxy_pass` from `location /` to a pointless `location` block.  This prevents the generation of a `location /` block for the `proxy_pass`, which would trigger the creation of the default root.  (A regular expression that never matches ensures that users can't craft special URLs to match the location block and thus bypass the custom `location /` config.)
  2. Disable the default root by setting `DEFAULT_ROOT=none`.
  3. Define a suitable `location /` block in their vhost-specific include file.

However, that's a very kludgy solution.

> #1286

The OP doesn't say whether the `*_location` include file meets their needs, so it's unclear whether any change is necessary at all.  However, a follow-up comment says, "I only want to change the `proxy_pass` value in `location /`", which implies that they have (implicit) `VIRTUAL_PATH=/`.  Thus, the `DEFAULT_ROOT=none` approach would not work for that commenter (unless they used the above kludge).

> #1086
> #1003

Not enough information to make a determination.

> #567

The user wants to override a `VIRTUAL_PATH`-generated `location` block, so `DEFAULT_ROOT` does not apply here either (except for the kludge).

> using `DEFAULT_ROOT=none` to achieve this seems the best way to implement this.

I don't think there's any reason we couldn't do both options (both per-path `location` overrides and a way to disable the default root).  I'll open a separate PR for `DEFAULT_ROOT=none`, and keep this PR about per-path `location` overrides.

rhansen:
> I'll open a separate PR for `DEFAULT_ROOT=none`, and keep this PR about per-path `location` overrides.

Done: #2146

buchdag:
>  I'll open a separate PR for `DEFAULT_ROOT=none`, and keep this PR about per-path `location` overrides.

Suits me, I approved #2146 

Regarding the per-path `location` overrides, I'm on board with the `*_location_override` files that simply disables the `location` block on inclusion, as you suggested a bit earlier. This feels more coherent with the existing custom location mechanisms than adding yet another environment variables as I suggested last week.

buchdag:
@rhansen looks good to me so far, with added tests this can be merged.

rhansen:
> @rhansen looks good to me so far, with added tests this can be merged.

Done.

buchdag:
Do you think you could add a test container that does not use `VIRTUAL_PATH` at all ?

Even if this is functionally the same as `VIRTUAL_PATH: /`, I'd prefer to have a dedicated test for the baseline use of the feature (no `VIRTUAL_PATH` and only `VIRTUAL-ROOT: root-nohash.nginx-proxy.test`).

rhansen:
> Do you think you could add a test container that does not use `VIRTUAL_PATH` at all ?

Done.  Turns out we have a bug unrelated to this PR:  If one container has `VIRTUAL_PATH=/foo` and another doesn't have `VIRTUAL_PATH` at all (but the same `VIRTUAL_HOST`), then nginx-proxy ignores the container without `VIRTUAL_PATH`.  (Maybe this is working as intended?)  I avoided this issue in the expanded test cases.

buchdag:
That might be convenient or okay but I'm pretty sure we never explicitly made it work that way 🤔

buchdag (approved):

PR #1185, opened 2018-10-24
Set proper X-Forwarded-Ssl for SSL-terminated setups

kressh:
Hello! Thanks for awesome project!

I use nginx on same machine to terminate SSL and serve static files. So my nginx-proxy instances are http-only and use only docker network. That's why I got `X-Forwarded-Ssl` set always to "off" because it based on `$scheme`. In my PR I utilise `$proxy_x_forwarded_proto` mapping (which based on `X-Forwarded-Proto` header and falls back to `$scheme` if empty) to detect ssl connection.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -51,8 +51,8 @@ server_names_hash_bucket_size 128;
 ssl_dhparam /etc/nginx/dhparam/dhparam.pem;
 {{ end }}
 
-# Set appropriate X-Forwarded-Ssl header
-map $scheme $proxy_x_forwarded_ssl {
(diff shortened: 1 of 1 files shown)


kamermans:
Hi @kressh, this looks like a good improvement.  It gets a thumbs-up from me @jwilder.

buchdag:
@kamermans stil relevant ?

kressh:
@buchdag I believe it is: https://github.com/nginx-proxy/nginx-proxy/blob/main/nginx.tmpl#L101

kamermans:
Yes, I think this is fine still @buchdag.

buchdag:
Thanks @kressh @kamermans, merging

PR #1194, opened 2018-11-12
support grpc protocol

umevoshi:


umevoshi:
I don't know why build is failed with updated README.md only.

PR #1197, opened 2018-11-14
Add support for Client-Side Certificate Authentication

Nobody84:
Adding support to enable client side certificate authentication per virtual host in the same way a the basic authentication works. If a certificate file with the name of a Virtual_Host.crt is located in directory /etc/nginx/ca, the server will activate ssl_verify_client using this crt file. If you also want to use a certificate revocation list you have to copy your crl file to the /etc/nginx/ca with the name Virtual_Host.crl.


maresb:
It would be great to see this functionality merged.  

In case anyone else is wondering how to do client certificates in the meantime, it's quite simple via config files.  For example, for proxy-wide client certificates, add the following lines to my_proxy.conf (as described in README.md):
```
ssl_client_certificate /etc/nginx/certs/ca.crt;
ssl_verify_client on;
```
This of course assumes that your client certificate CA has been added to the path above.

Sy14r:
Is this planning to be merged eventually? It doesn't appear that there are any blockers.

buchdag:
@Sy14r I want to have the test suite working again before starting to merge new features, see #1304 

Also, ideally, I'd like new features to have corresponding tests.

PR #1217, opened 2018-12-27
Creating index.html with a list of URLs to published containers

DimShadoWWW:
Creating an index.html file automatically with a list of URLs to the published containers


SilverFire:
From my point of view, this pull request leads to disclosure of domains that work behind the proxy.
It might be a good idea to create a kind of CLI script to list active container, but I definitely would not like to see this PR merged as is.

buchdag:
@DimShadoWWW closing this PR for reasons mentioned by @SilverFire, you don't want to automatically expose this kind of information.

PR #1234, opened 2019-02-15
 Added ability to override proxy listening ports.

stuckj:
With this change you can have nginx-proxy listen on a different port
from the default for HTTP and HTTPS. E.g., LISTEN_PORT=8080 sets the
HTTP port listen port to 8080 instead of 80. Likewise
LISTEN_PORT_SSL=8443 would use 8443 for HTTPS. This makes HTTP->HTTPS
redirection work properly when forwarding a port from a non-priviledged
VM (no ports < 1000) to nginx-proxy.

README.md is updated with the new variables available for config.

NOTE: To clarify, the entire reason for this feature is to allow proper automatic http->https redirection when nginx-proxy can't be bound to the standard 80 / 443 ports for http / https. Simply using docker port mapping (e.g., -p 8080:80 -p 8443:443) will properly map traffic onto the non-standard ports. But, the http->https redirection will use the standard ports. E.g., hitting http://someurl:8080 will end up redirecting to https://someurl with port redirection whereas it will hit https://someurl:8443 with this change (assuming you use 8080 and 8443).

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,12 @@ $ docker network connect my-other-network my-nginx-proxy
 
 In this example, the `my-nginx-proxy` container will be connected to `my-network` and `my-other-network` and will be able to proxy to other containers attached to those networks.
 
+### Listening Port
+
+If you want `nginx-proxy` to listen on a port different from the default port, use the environment variable `LISTEN_PORT`. E.g., to listen on port 8080, use `LISTEN_PORT=8080`. The default port is 80 for `HTTP` traffic.
+
+Likewise, you can set the default listen port for `HTTPS` traffic by setting `LISTEN_PORT_SSL`. E.g., `LISTEN_PORT_SSL=8443` would configure `nginx-proxy` to listen on port 8443 for `HTTPS` traffic. If you change the `HTTPS` port, the redirect for `HTTPS` traffic will also be configured to redirect to the custom port. The default port is 443 for `HTTPS` traffic.
+
 ### Internet vs. Local Network Access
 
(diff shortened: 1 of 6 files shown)


MarkNBroadhead:
This would be a great addition, would probably make sense to squash these commits? Anyways, I just a passerby that would love this patch. I hope this can get merged soon.

BoehmFloh:
Ohh man, I wish this feature was already built in when I launched my first proxy.

I hope your request get's merged soon.

flowercow:
how to test this patch?
just download the nginx.tmpl and replace the original one, and build the image?

I find there is a command to test the listen_port by {{ $listen_port := coalesce $.Env.LISTEN_PORT "80" }}
{{ $listen_port_ssl := coalesce $.Env.LISTEN_PORT_SSL "443" }}
is there any package that must be install to use the 'coalesce' command ?
and how to use the LISTEN_PORT and LISTEN_PORT_SSL environment variables in docker-compose.yml file?
I want to  bind the proxy listening por to port 8080 and 8443 of the host, is this patch fit for that?

flowercow:
i have made an updated docker image with this patch, and set the docker reverse proxy . The docker installed in nas in a private lan. The  proxy listing ports are 8080( http) and 8443( https) , I can access the website using an public domain name through dmz of the router which has a public IP address. All works fine.
But when i using a private IP of the nas, such as http://192.168.8.3:8080, browser will be redirect to https://192.168.8.3:8443, but after login action, I got an error "400 Bad Request The plain HTTP request was sent to HTTPS port", and the url in the browser became "http://192.168.8.2:8443/ login.php"
what wrong with this?
thanks for any suggestions.

eigood:
This PR is wrong.  The container(s) in all docker environments should *always* listen on their default ports(80:443 in this case).  But then, the external mapping, would be updated.

Like so: docker run .... -p 8080:80 8443:443 ...

stuckj:
@eigood, you may have missed this part of my comment in the PR or perhaps I didn't emphasize it enough:

> This makes HTTP->HTTPS redirection work properly when forwarding a port from a non-privileged VM (no ports < 1000) to nginx-proxy.

If you're using SSL nginx-proxy handles http->https redirects for you with rules it creates in the nginx config. If you just use docker port mappings, the configuration will still refer to ports 80 and 443. So, the redirect will end up going to 443 even if you hit the http page on 8080 (using the docker port mapping).

This solves that issue by changing the container to use the specified ports rather than just using port mappings. That was the whole purpose of this PR.

An alternative solution would be to have the generated config use different port mappings, but that seemed more complicated to me.

Cheers,
-Johnny

buchdag:
This PR has been adapted in #1662, closing

PR #1242, opened 2019-03-08
Enable gzip compression

lucasayres:
This configuration will reduce the size of files sent from server to increase the speed to which they are transferred to the browser.


buchdag:
Agreed with @khalilgharbaoui, this should be in a `conf.d/xxx.conf` file, not forced upon every users as a default.

PR #1243, opened 2019-03-08
Enable Leverage Browser Caching

lucasayres:
Benefits of Enable Leverage Browser Caching:
  - Primary benefit is speeding up website because static files will serve from your browser. it saves internet data of website visitor. it also saves bandwidth of website server and decrease load of server. Simply it decreases HTTP requests.
  - Score on Google PageSpeed Insights and Gtmetrix

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -57,6 +57,16 @@ map $scheme $proxy_x_forwarded_ssl {
   https on;
 }
 
+# Expires map
+map $sent_http_content_type $expires {
+  default                   off;
+  ~application/             30d;
+  ~text/                    30d;
+  ~audio/                   365d;
+  ~video/                   365d;
+  ~image/                   365d;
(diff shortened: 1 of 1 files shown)


khalilgharbaoui:
shouldn't this be in the `conf.d/xxx.conf` instead?

ProteanCode:
Can somebody look more into it? I am here 2nd time and this is the only one working solution for me since now when I do a reverse proxy to wordpress site anything like this:

```
location ~* \.(jpg|jpeg|png|gif|ico|css|js)$ {
     expires 365d;
}
```

Causes some heavy performance issues (more likely some kind of loop), and in the end returns 404.

So far this is the only one solution, but this is directly inside /etc/nginx/conf.d/default.conf which is not best solution.

Can we somehow manage to use it inside `conf.d`?

PR #1255, opened 2019-03-27
Add ability to configure SSL timeout per host

smcpeck:
I found a need for this and thought others might benefit from it as well.  This is a quick drive-by, so let me know if any documentation or tests needs to be updated.

🍻


smcpeck:
Hmmm.  New to Travis CI & can't really figure out how to dive into the failures here.  Any pointers in the right direction would be appreciated, @jwilder, as a learning experience so I can get this fixed up.

PR #1263, opened 2019-04-09
Change dhparam positional args to be inherited vars, standardize naming

jessejoe:
PR #913 added `DHPARAM_GENERATION` as a positional argument to generate-dhparam.sh. However, since it was the second positional argument,`DHPARAM_BITS` would also have to be defined or `DHPARAM_GENERATION` would be read into `DHPARAM_BITS`. This changes the arguments to be inherited variables which do not depend on order, just declaration.

Also change instances of `GENERATE_DHPARAM` to `DHPARAM_GENERATION` since it's unnecessary to have another variable. I think `GENERATE_DHPARAM` is actually a better name (verb vs. noun), but `DHPARAM_GENERATION` is already defined and may break someone if changed.

Addresses https://github.com/jwilder/nginx-proxy/pull/913#issuecomment-476014691

----

### Behavior before:
```
$ docker run -it --rm -e "DHPARAM_GENERATION=false" testing /bin/bash
dhparam: Can't parse "false" as a number
```
### Behavior after:
```
$ docker run --rm -e "DHPARAM_GENERATION=false" testing /bin/bash
Skipping Diffie-Hellman parameters generation and Ignoring pre-generated dhparam.pem
```

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -4,7 +4,7 @@ set -e
 # Warn if the DOCKER_HOST socket does not exist
 if [[ $DOCKER_HOST = unix://* ]]; then
 	socket_file=${DOCKER_HOST#unix://}
-	if ! [ -S $socket_file ]; then
+	if ! [ -S "$socket_file" ]; then
 		cat >&2 <<-EOT
 			ERROR: you need to share your Docker host socket with a volume at $socket_file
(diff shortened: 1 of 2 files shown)


kamermans:
[CC @jwilder] this looks good to me, thanks @jessejoe!

buchdag:
@jwilder this is a required fix, could you merge please ?

buchdag:
- rebased on `main`
- fixed merge conflict
- included @joren485 suggestion
- did a bit of linting on the shell scripts while we're at it

Issue #172, opened 2015-05-31
Use labels instead of env variables


Nadrieril:
With the new labels addition to docker 1.6 (https://docs.docker.com/userguide/labels-custom-metadata), it would be cleaner to use labels instead of environment variables.

For example, we could set the label 'com.jwilder.nginx-proxy.virtual-host' to the host we want, and so on.

It need not be a breaking change, you can handle both behaviours at the same time

thaJeztah:
+1 thinking of this, I think I actually suggested this in https://github.com/jwilder/nginx-proxy/issues/36#issuecomment-98262304

Labels were already suggested in https://github.com/jwilder/docker-gen/issues/81 and implemented in https://github.com/jwilder/docker-gen/pull/82

So, it looks like this should be possible, with a modified template?

md5:
@thaJeztah I think a new release of `docker-gen` still needs to be cut to include the latest changes.

jwilder:
Yeah, env vars were initially used because labels didn't exist at the time.  I'm in favor of supporting labels.  though.

jwilder:
docker-gen 0.4.0 has label support now.

mathroc:
just a recommendation for when choosing the labels name "It's recommended that you use reverse-DNS notation to prevent your labels from conflicting with those used by other software." ( https://github.com/docker/compose/blob/master/docs/yml.md#labels )

also, it could replace #60

thaJeztah:
> It's recommended that you use reverse-DNS notation

:+1: I mentioned that in my original https://github.com/jwilder/nginx-proxy/issues/36#issuecomment-98262304, but forgot to repeat it here. Thanks for the reminder!

iBobik:
Sorry for dump question.

Does in mean that it will be possible to change e.g. host without restarting container?

md5:
@iBobik Labels can only be set when a container is started (as far as I know), so supporting labels wouldn't help you change a hostname without restarting a container.

Boran:
@iBobik : I think labels can only be set when creating (not starting) a container,  i.e. same limits as environment variables. It would be grate if a container could be stopped, labels changed and restarted.

ibnesayeed:
What is stopping this from happening?

iBobik:
Approve from the @jwilder :-)



Jan Pobořil
http://honza.poboril.cz

mathroc:
[...]

PR #1269, opened 2019-04-17
Use labels instead of env variables

rhbvkleef:
Removes the use of environment variables for defining virtual hosts. Environment variables are still used for configuring the proxy container itself (and potentially the add-on container for docker-gen)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,13 +12,13 @@ To run it:
 
     $ docker run -d -p 80:80 -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy
 
-Then start any containers you want proxied with an env var `VIRTUAL_HOST=subdomain.youdomain.com`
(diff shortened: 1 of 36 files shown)


rhbvkleef:
I can't quite figure out why the tests are failing, but I have a suspicion that it is some kind of racing.

rhbvkleef:
Also, I'd like to reference #172.

rhbvkleef:
I should probably also rename the label from `@proxy/*` to a `proxy.nginx.*` format. After all, we're not doing Kubernetes here.

kamermans:
Unless I'm missing something, merging this would break every single installation of `nginx-proxy` (probably > 100k).  If we would entertain labels, it would almost certainly be alongside env vars, not as a replacement.

jwilder:
This is a breaking change and can’t be merged.

PR #1292, opened 2019-06-22
Old stye exceptions --> new style for Python 3

cclauss:
Old stye exceptions are ___syntax errors___ in Python 3 but new style exceptions work as expected in both Python 2 and Python 3.


buchdag:
Superseded by #1572

PR #1296, opened 2019-06-29
Use container names instead of container IPs

hasnat:
Results in less nginx reloads, offloads all the IP resolution to resolver instead of docker-gen
when using such keep `resolver 127.0.0.11 valid=5s;`
via `ENV RESOLVERS="127.0.0.11 valid=5s"`


jpic:
Tempting but fails tests, perhaps there is a docker version requirement that the test vms don't support ?

hasnat:
sorry haven't investigated why big number of tests fail. I can check a bit more if the idea is accepted, someone might comeback explaining why IPs are in use instead of container names

pablos91:
I use container names instead of IPs when using swarm in my fork.

https://github.com/pablos91/nginx-proxy/commit/85c2f1cfaec13c159d531c9bf1b88da07619f0d8

It is more friendly when services are only deployed to workers, not the manager node. Before the amendment I had to deploy at least one container on manager node for nginx-proxy to recognize there is a service running.

You might try my fork anyway it also has support for virtual paths so you can deploy apps like http://example.com/api just using another environmental variable VIRTUAL_PATH=/api

buchdag:
I get the idea but unfortunately this PR isn't useable in this state (no doc nor explanation for a new environment variable, failing tests and important merge conflicts).

> offloads all the IP resolution to resolver instead of docker-gen

docker-gen does not perform any DNS resolution, it's just looking up for the containers IPs obtained from the Docker endpoint at template render time.

PR #1306, opened 2019-07-20
add support for none standard https port 

ghw59:
set environment variable HTTPS_PORT to support https port other than 443,  HTTP redirects to this https port if redirect enabled.

-e HTTPS_PORT=9443


buchdag:
Superseded by #1353

PR #1313, opened 2019-08-08
Upgrade nginx to 1.17

JIAZHEN:
Similar to https://github.com/jwilder/nginx-proxy/pull/1298

Just fix the test


JIAZHEN:
@jwilder could you please have a look?

PR #1322, opened 2019-08-29
Update ssl configuration

came88:
This PR upgrade ssl configuration in order to use latest Mozilla recommendation (version 5.0) for the modern, intermediate and old profile.
Moreover, it add an option to change the default ssl policy defining the `SSL_POLICY` environment variable on the nginx-proxy container itself. Other containers can still change the ssl policy for their virtual host by setting the `SSL_POLICY` environment variable on itself as before.


nginx.tmpl
```diff
@@ -22,6 +22,48 @@
 
 {{ end }}
 
+{{ define "ssl_policy" }}
+	{{ if eq .ssl_policy "Mozilla-Modern" }}
+		ssl_protocols TLSv1.3;
+		{{/* ssl_ciphers is undefined in the Mozilla-Modern policy /*}}
```
deAtog: The Mozilla-Modern policy currently defines 3 ciphers as specified here: https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility

Nginx currently lacks the ability to change the default TLSv1.3 ciphers used by OpenSSL. According to the OpenSSL documentation, **all** TLSv1.3 ciphers are incompatible with TLSv1.2 and lower and **all** TLSv1.2 and lower ciphers are incompatible with TLSv1.3. To avoid enabling TLSv1.3 without any ciphers specified, the configuration for TLSv1.3 was separated from the TLSv1.2 and lower. Nginx sees this as an unworkable solution and refuses to use the new OpenSSL API to configure TLSv1.3 ciphers. The current workaround for this is to configure the list of TLSv1.3 ciphers globally via the openssl.conf file.

deAtog:
@jwilder Please merge this. The current defaults generated are outdated and the addition of a global option to specify the default SSL policy is a great addition. We are currently using this container in a production environment with a decent amount of containers. In order to conform to PCI standards we must disable TLSv1.0 but cannot currently do so without specifying a per-container SSL_POLICY setting.

nginx.tmpl
came88: Thanks for pointing it out, I've fixed the comment

Issue #1324, opened 2019-09-04
Config using environment variables


poma:
Currently the only way to set a global or per-vhost config is to mount a corresponding conf file into nginx container. Even though it can be done via creating custom container or mounting a volume with config files, using env vars would be more easy and effortless. My proposition is as follows:

At startup, scan all env vars starting with NGINX_* and add corresponding entries to /etc/nginx/conf.d/environment.conf and for each vhost do the same for hosts env vars. This way we would avoid hardcoding and maintaining an array of possible env variable names. Example usage from docker-compose:

```
services:
  nginx:
    image: jwilder/nginx-proxy
    environment:
      NGINX_CLIENT_MAX_BODY_SIZE: 30M
    ports:
      - 80:80
    ...
  app:
    image: app
    environment:
      VIRTUAL_HOST: example.com
      NGINX_ACCESS_LOG: 'off'
```

or from the command line:

```
 docker run -d -p 80:80 -e NGINX_CLIENT_MAX_BODY_SIZE=30M ... jwilder/nginx-proxy
```

There are 2 related issues #597 and #688 but they are a few years old and unanswered.

Is there interest for implementing this? I can try to do it. Maybe a similar solution already exists?

Or is it a bad idea? If so, why?

poma:
I've made a simple wrapper that adds env variables to global config using a shell script https://github.com/poma/docker-gen

drequivalent:
Need this

btxtiger:
Any update?

PR #1326, opened 2019-09-05
support nginx_* env variables on nginx container and vhosts

poma:
Implementation of #1324 feature request. Environment variables starting with `nginx_*` are added as nginx config statements, for example: `nginx_client_max_body_size=30M`. Variables of jwilder/nginx-proxy container are added to the global config, and variables on others containers are added to the respective vhosts.

Example `docker-compose.yml`:

```yaml
version: '2.1'

services:
  nginx:
    image: jwilder/nginx-proxy
    ports:
      - 80:80
      - 443:443
    volumes:
    environment:
      nginx_client_max_body_size: 42M

  app:
    image: app
    environment:
      VIRTUAL_HOST: example.com
      nginx_client_max_body_size: 1337M
```

Currently only lower case variables are supported (until jwilder/docker-gen#306 is merged). Also if there is an easy way to merge `Env` dicts from all containers with the same vhost, we should do that, currently `Env` map is taken from the first container for each vhost.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -191,6 +191,12 @@ log_format vhost '$host $remote_addr - $remote_user [$time_local] '
 
 access_log off;
 
+{{ range $key, $val := .Env }}
+{{ if and $val (hasPrefix "nginx_" (toLower $key)) }}
+{{ trimPrefix "nginx_" (toLower $key) }} {{ $val }};
+{{ end }}
+{{ end }}
+
 {{/* Get the SSL_POLICY defined by this container, falling back to "Mozilla-Intermediate" */}}
 {{ $ssl_policy := or ($.Env.SSL_POLICY) "Mozilla-Intermediate" }}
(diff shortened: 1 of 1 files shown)


pnbecker:
I like this PR and think it would be a helpful addition.

poma:
While we are waiting for this to be merged, you can use updated docker-gen from the repo [poma/docker-gen](https://github.com/poma/docker-gen) that already contains this PR

poma:
Rebased PR onto the current master, now also supports upper case "NGINX_*" env vars.

One problem I've encountered is that some containers export export "NGINX_*" env vars that are invalid in the context of nginx conf and this might lead to broken nginx config generation. Notably `nginx:alpine` exports NGINX_VERSION which is an invalid in nginx config, this might happen if you want to put another nginx instance behind the main nginx proxy. There is a workaround of overriding exported value with an empty string. But it's still a breaking change that can break a working server when people update to it, so be careful.

PR #1342, opened 2019-10-08
Make it possible to serve static files locally with fastcgi upstream

TheLastProject:
The idea of this pull request is to allow only sending a chosen list of extensions upstream, so that this container could serve static files itself and only make upstream (PHP-FPM for example) deal with the non-static files.

I have only tested this with fastcgi and PHP-FPM as upstream, but it works quite well here.


LittleHuba:
Would it be possible to merge this PR? Currently FastCGI is not really usable in any practical way.

GunoH:
Hoping this get merged soon; would like to use FastCGI backends but having no success :(

PR #1353, opened 2019-11-03
Add support for custom external HTTP/HTTPS ports

nanawel:
This PR allows to choose arbitrary ports for HTTP and HTTPS instead of 80 and 443.

While it does not change anything in the way ports can be mapped to host using native Docker capabilities, it fixes the "invalid" `X-Forwarded-Port` when the port on which nginx listens to is different from the one it is mapped - and accessed by clients - on the host.

Typical usage in docker-compose (here with the custom ports `1080` and `10443`):
```
  reverse-proxy:
    image: jwilder/nginx-proxy
    ports:
      - "1080:1080"
      - "10443:10443"
    environment:
      HTTP_PORT: "1080"
      HTTPS_PORT: "10443"
    volumes:
```

I'm open to suggestions.
Thanks for your amazing work on this image, it helps a lot.

Regards,

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -1,5 +1,8 @@
 {{ $CurrentContainer := where $ "ID" .Docker.CurrentContainerID | first }}
 
+{{ $external_http_port := coalesce $.Env.HTTP_PORT "80" }}
+{{ $external_https_port := coalesce $.Env.HTTPS_PORT "443" }}
+
 {{ define "upstream" }}
 	{{ if .Address }}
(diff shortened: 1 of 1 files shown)


thiagotalma:
Please close.

addshore:
Is there any timeline for this feature getting into the docker images?
We were just looking into finding something like this, and then discovered this PR merged a few hours ago!

DrRSatzteil:
Will this also create correct redirects from HTTP_PORT to HTTPS_PORT?

PR #1373, opened 2019-12-23
Allow to run on swarm mode setting VIRTUAL_PORT

LuisEnMarroquin:


hugokoopmans:
would be great to get this merged right?
looking for a solution to run in swrm mode @LuisEnMarroquin  could you elaborate a bit how to run nginx then on the swarm? Would this work with the letsencrypt-compagnion? thx

PR #1379, opened 2020-01-09
Add VIRTUAL_PATH and VIRTUAL_PATH_STRIP to run reverse-proxy as LB/proxy for pool of services

KeyWeeUsr:
Rebased #1257 and new feature that removes the incorrect behavior if the service expects deployment on the domain directly and under root (`/`) path.

Without stripping requests fail because the `VIRTUAL_PATH` is propagated into the service and service might not be created with such behavior in mind. Pretty much none of big projects nowadays because everyone wants separate DNS record that gets translated into `Host` for nginx to route :grimacing:.

Example:

    # accessing directly VIRTUAL_PATH with stripping
    GET /service  (nginx uses 302 to bump to the correct location)
    GET /service/ (nginx allows communication with container)
    GET /         (the request as visible within the service container)

    # accessing VIRTUAL_PATH without stripping
    GET /service  (nginx routes traffic directly to container)
    GET /service  (the request as visible within the service container)
    404/500/...   (service rejects request because can't handle path prefix)

Tests such as https://github.com/jwilder/nginx-proxy/blob/e7624687599a44a09b23e79c7fc8c614778eabae/test/test_multiple-hosts.py#L5 were failing, but it seems to be wrong anyway since port should be written with a colon (`:`) because the slash (`/`) is a part of a location/route for nginx.

Testing Alpine image available [on Docker Hub](https://hub.docker.com/repository/docker/keyweeusr/nginx-proxy).

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -297,16 +297,34 @@ server {
 	{{ $is_regexp := hasPrefix "^" $location }}
 	{{ $modifier := when $is_regexp "~" "" }}
 
+	{{ if (eq $container.Env.VIRTUAL_PATH_STRIP "TRUE") }}
 	location {{ $modifier }} '{{ $location }}' {
+		return 302 '{{ $location }}/';
+	}
+	{{ end }}
+
+	{{ if (eq $container.Env.VIRTUAL_PATH_STRIP "TRUE") }}
+	location {{ $modifier }} '{{ $location }}/' {
+	{{ else }}
+	location {{ $modifier }} '{{ $location }}' {
+	{{ end }}
 		{{ if eq $proto "uwsgi" }}
 		include uwsgi_params;
+		{{ if (eq $container.Env.VIRTUAL_PATH_STRIP "TRUE") }}
+		uwsgi_pass {{ trim $proto }}://{{ trim $container.Name }}/;
+		{{ else }}
 		uwsgi_pass {{ trim $proto }}://{{ trim $container.Name }};
+		{{ end }}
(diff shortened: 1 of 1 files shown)


nginx.tmpl
```diff
@@ -339,16 +363,41 @@ server {
...
 
-	location / {
+	{{ range $container := $containers }}
+	{{/* Determine whether this container has specified a virtual path, or default to the / pattern */}}
+	{{ $location := coalesce $container.Env.VIRTUAL_PATH "/" }}
+
+	{{ $is_regexp := hasPrefix "^" $location }}
+	{{ $modifier := when $is_regexp "~" "" }}
+
+	{{ if (eq $container.Env.VIRTUAL_PATH_STRIP "TRUE") }}
```
tyrantkhan: this call is erroring when VIRTUAL_PATH_STRIP is not set btw.

KeyWeeUsr: I think it's fixed. Also, I changed `TRUE` to `true` so it's the same as in other conditions.

tyrantkhan: awesome thanks :). I'm not too hopeful this PR will make it into master (just because previous attempts have not gotten feedback from the maintainer -) but can confirm i'm using this in my local version of the nginx-proxy and it is working well .

cwildfoerster:
Test is failing because multiple upstream with the same name will be created, which causes nginx to throw an error (duplicate upstream).

thiagotalma:
Please close.

buchdag:
Superseded by #1901

PR #1386, opened 2020-01-30
Add fallback to the proxy containers env for HTTPS_METHOD and HSTS

juliushaertl:
The current implementation requires each container to set the `HTTPS_METHOD` and `HSTS` environment variable individually. This pr adds an intermediate fallback to allow setting the value also on the main nginx-proxy container. If they are not set, the default will be the same as before.


thiagotalma:
Please close.

juliushaertl:
@thiagotalma No https://github.com/jwilder/nginx-proxy/pull/1388 was closed not merged.

thiagotalma:
My bad!

PR #1402, opened 2020-02-28
Upgrade to 1.17.8

sgabe:
This PR upgrades Nginx to version 1.17.8, see [CHANGES](https://nginx.org/en/CHANGES) for more information.


maxfi:
@jwilder When will this be pushed to docker hub? I'm using https://github.com/zodern/meteor-up which uses the docker hub image so unfortunately can't use an image that I build myself. Thanks in advance!

PR #1407, opened 2020-03-17
Fix dhparam parsing error

nathanbigaignon:
Parsing error raised when DHPARAM_BITS is blank. Fix by setting default value for DHPARAM_BITS to avoid DHPARAM_GENERATION to be send as first argument of generate-dhparam.sh


buchdag:
Closing in favor of #1263 approach

PR #1409, opened 2020-03-20
Bring ACME no redirection inline with companion

buchdag:
Add the following to the Let's Encrypt ACME challenge "no redirection to HTTPS"

https://github.com/nginx-proxy/docker-letsencrypt-nginx-proxy-companion/pull/570
https://github.com/nginx-proxy/docker-letsencrypt-nginx-proxy-companion/pull/335


sgabe:
Both `auth_request off;` and the `^~` modifier could save some headaches as discussed in the referenced threads. Both have been there in the companion container for quite some time now, without new issues, so LGTM.

PR #1429, opened 2020-04-26
Override location /, allow static files binding

PiTiLeZarD:


PiTiLeZarD:
The merge to master on my side updated the pull request and has now all my swarm updated, I'll try to create a pull request with only the location override and static binding

PR #1432, opened 2020-05-09
Fix docker-gen Docker Hub URL

KetchupBomb:
The Docker Hub links are stale. They take you to a page which loads the Docker Hub header and footer, but no content about the image.

The new URLs load the image description correctly.


sgabe:
@buchdag This one looks easy enough to merge.

PR #1444, opened 2020-06-01
Do not copy useless files into the image

pini-gh:
Move required files into a local 'app' folder and copy the folder
content into the image.

It does not make sense to have the dockerfiles and docker-compose files copied into the image. It wastes resources and it makes `docker-compose up --build` to rebuild the image each time a docker-compose file is changed.


sgabe:
Might be best to check the other PRs before merging this one.

buchdag:
@pini-gh I'd like to merge this as it somewhat solves the discussion we had in #1769 but there might unfortunately be a lot of people that rely on the current location of the `nginx.tmpl` file for three containers setups.

Maybe keep the `nginx.tmpl` file at the root and move everything else to `/app` ?

thaJeztah (changes requested):

Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN wget https://github.com/jwilder/docker-gen/releases/download/$DOCKER_GEN_VER
 
 COPY network_internal.conf /etc/nginx/
 
-COPY . /app/
+COPY app /app/
```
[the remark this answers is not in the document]
pini-gh: > Adding the trailing `/.` means "copy the _content_ of the directory to ", whereas _without_, it means "Copy to . Depending on if `<target>` exists or not, and wether or not `<target>` is a directory, this can cause `<source>` to either land in (in this case) `/app/` or `/app/app/`. See https://docs.docker.com/engine/reference/commandline/cp/#extended-description.

This is not true about the Dockerfile `COPY` command. See [1]:
> If `<src>` is a directory, the entire contents of the directory are copied, including filesystem metadata.
    Note
    The directory itself is not copied, just its contents.

[1] https://docs.docker.com/engine/reference/builder/#copy

pini-gh:
> @pini-gh I'd like to merge this as it somewhat solves the discussion we had in #1769 but there might unfortunately be a lot of people (and PR) that rely on the current location of the `nginx.tmpl` file for three containers setups.
> 
> Maybe keep the `nginx.tmpl` file at the root and move everything else to `/app` ?

I think people updating their instances could read release notes and easily adapt their configuration to this path change. But I understand your concerns.

Your call.

Dockerfile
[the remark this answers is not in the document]
pini-gh: > I guess we need to make the same changes in `Dockerfile.alpine` as well?

Sure.

junderw:
> I think people updating their instances could read release notes and easily adapt their configuration to this path change. But I understand your concerns.
> 
> Your call.

This is a breaking change. Usually projects hold off on breaking changes until some significant new feature that requires a breaking change is done, then they can include a bunch of smaller breaking changes.

This change as-is would only be included in the next major release, and no, this change by itself does not merit a major release imo.

pini-gh:
> This change as-is would only be included in the next major release, and no, this change by itself does not merit a major release imo.

So be it.

polarathene (commented):
Note that due to changes since the last commit here, `dhparam.pem.default` no longer exists, [there is now a directory `dhparam/`](https://github.com/nginx-proxy/nginx-proxy/tree/main/dhparam) that would need to be moved to `app/` instead. `generate-dhparam.sh` also has since been removed.

As for the breaking change concern, I imagine you can pin an open issue and include release notes in advance of the breaking release that notify the change in advance with enough time in-between before that release?

That said, the problem the PR is trying to solve since June 2020 could also just update the `.dockerignore` in the meantime, less discussion/concerns related to that approach and it'd probably have been merged long ago? (_probably still a good idea to eventually group the files intended for the image into a subdir instead of interleaved with other files/dirs at the root though_ :sweat_smile: )

.dockerignore
```diff
@@ -1,6 +1,10 @@
 .git
+.github
+test
 .dockerignore
-circle.yml
+.gitignore
+*.yml
+Dockerfile*
+LICENSE
```
tkw1536, line 8: Legally speaking, isn't it a license violation to not put the LICENSE file inside the Docker image?

The license says:
> The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software.


IANAL but I would say the Docker Image qualifies as a copy of the software and should thus contain a copy of the license.

thaJeztah: Yes, I think it should be included (I thought I left a comment about that, but must've been another repository 😓)

buchdag: Should it be included at the root of the image or inside `/app` ?

thaJeztah: To be honest; it's something that's often forgotten, and not sure if there's a clearly defined location for it, although for regular `deb` / `rpm` packages, such files (`LICENSE`, `NOTICE`) would be installed in (e.g.) `/usr/share/doc/<PACKAGE NAME>/`.


In addition, it's also possible to set one or more `LABEL` to provide some metadata (see https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys), although this can be somewhat tricky if the image is based on a base image and contains other software (with different licensing), so not sure if adding "partial" information is better than "not including it".

nduchon: The GitHub Actions workflow uses `docker/metadata-action@v3` to tag the image, so we're already setting some `org.opencontainers.image` labels on the images, including `licenses` :

```
org.opencontainers.image.title=nginx-proxy
org.opencontainers.image.description=Automated nginx proxy for Docker containers using docker-gen
org.opencontainers.image.url=https://github.com/nginx-proxy/nginx-proxy
org.opencontainers.image.source=https://github.com/nginx-proxy/nginx-proxy
org.opencontainers.image.version=latest
org.opencontainers.image.created=2022-03-07T15:23:02.325Z
org.opencontainers.image.revision=6b48e11e5cbe8066a619300fa7b40408da02387c
org.opencontainers.image.licenses=MIT
```

I really have no preference over putting the `LICENSE` file at the root or inside `/app`.

buchdag: The GitHub Actions workflow uses `docker/metadata-action@v3` to tag the image, so we're already setting some `org.opencontainers.image` labels on the images, including `licenses` :

```
org.opencontainers.image.title=nginx-proxy
org.opencontainers.image.description=Automated nginx proxy for Docker containers using docker-gen
org.opencontainers.image.url=https://github.com/nginx-proxy/nginx-proxy
org.opencontainers.image.source=https://github.com/nginx-proxy/nginx-proxy
org.opencontainers.image.version=latest
org.opencontainers.image.created=2022-03-07T15:23:02.325Z
org.opencontainers.image.revision=6b48e11e5cbe8066a619300fa7b40408da02387c
org.opencontainers.image.licenses=MIT
```

I really have no preference over putting the `LICENSE` file at the root or inside `/app`.

thaJeztah: Ah! Didn't check if it was using the metadata-action, nice!

Yes, no preference either where to stash the file(s) inside the image.

tkw1536: > not arguing against it but not including the License inside the Docker image does not seem to bother other MIT licensed projects like Traefik or Dokku.

I can't speak for other projects, but according to the license the full text and copyright notice should be included in the image. 
From what I can see the metadata-action only puts in the label 'MIT' and links to the repository - this isn't the same.
My personal preference is to include the LICENSE file directly inside the image.

buchdag: @tkw1536 @thaJeztah done in 4ea3437dfab423b6710b32b2aa9d50f9f750bed9

PR #1446, opened 2020-06-02
Imrove acme-challenge handling

pini-gh:
so that there is no need anomyre for the Let's Encrypt companion to fiddle
with vhosts nginx configuration.


pini-gh:
Closing and resubmitting to fix a typo

Issue #1468, opened 2020-07-10
test_dhparam_is_generated_if_missing test fails


tkw1536:
At the time of writing on the master branch the `test_ssl/test_dhparam_generation.py::test_dhparam_is_generated_if_missing ` test fails. 
I get the following output:
```
===================================================================================== test session starts =====================================================================================
platform linux2 -- Python 2.7.18, pytest-3.0.5, py-1.9.0, pluggy-0.4.0 -- /usr/local/bin/python
rootdir: [censored]/nginx-proxy/test, inifile: pytest.ini
collected 129 items 

test_DOCKER_HOST_unix_socket.py::test_unknown_virtual_host PASSED
test_DOCKER_HOST_unix_socket.py::test_forwards_to_web1 PASSED
test_DOCKER_HOST_unix_socket.py::test_forwards_to_web2 PASSED
test_composev2.py::test_unknown_virtual_host PASSED
test_composev2.py::test_forwards_to_whoami PASSED
test_default-host.py::test_fallback_on_default PASSED
test_events.py::test_nginx_proxy_behavior_when_alone PASSED
test_events.py::test_new_container_is_detected PASSED
test_ipv6.py::test_unknown_virtual_host_ipv4 PASSED
test_ipv6.py::test_forwards_to_web1_ipv4 PASSED
test_ipv6.py::test_forwards_to_web2_ipv4 PASSED
test_ipv6.py::test_unknown_virtual_host_ipv6 SKIPPED
test_ipv6.py::test_forwards_to_web1_ipv6 SKIPPED
test_ipv6.py::test_forwards_to_web2_ipv6 SKIPPED
test_multiple-hosts.py::test_unknown_virtual_host_is_503 PASSED
test_multiple-hosts.py::test_webA_is_forwarded PASSED
test_multiple-hosts.py::test_webB_is_forwarded PASSED
test_multiple-networks.py::test_unknown_virtual_host PASSED
test_multiple-networks.py::test_forwards_to_web1 PASSED
test_multiple-networks.py::test_forwards_to_web2 PASSED
test_nominal.py::test_unknown_virtual_host PASSED
test_nominal.py::test_forwards_to_web1 PASSED
test_nominal.py::test_forwards_to_web2 PASSED
test_nominal.py::test_ipv6_is_disabled_by_default SKIPPED
test_wildcard_host.py::test_wildcard_prefix[f00.nginx-proxy.test-81] PASSED
[...]

Issue #1370, opened 2019-12-19
Arm architecture not supported


samip5:
### What happened?
I tried to execute this container on armv7 platform, which unfortunately fails.

### What you expected to happen?
I expected it to work out-of-the box.

### Error log
```
standard_init_linux.go:211: exec user process caused "exec format error"
```

`docker image inspect jwilder/nginx-proxy |grep -i "Arch"`:
```
"Architecture": "amd64"
```

Related: #1262.

danywheeler:
what's the status? I use arm dev board do some test, but couldn't find nginx-proxy arm64 version!

samip5:
> what's the status? I use arm dev board do some test, but couldn't find nginx-proxy arm64 version!

There's no progress, but they're multiple nginx containers that have the support, like [linuxserver/nginx](https://hub.docker.com/r/linuxserver/nginx).

utrumo:
Hi!
I have some problem with arm processor on Zyxel Armor Z2:
`standard_init_linux.go:211: exec user process caused "exec format error"`

As an experiment i cloned this repository and changed references in Dockerfile:
- changed link to forego on ARM version 20180217041714 (i got it from https://github.com/ddollar/forego)
```
# Install Forego
RUN wget https://bin.equinox.io/c/ekMN3bCZFUn/forego-stable-linux-arm.tgz \
 && tar xvf forego-stable-linux-arm.tgz -C /usr/local/bin \
 && rm forego-stable-linux-arm.tgz
```

- changed docker-gen to armhf version:
```
RUN wget https://github.com/jwilder/docker-gen/releases/download/$DOCKER_GEN_VERSION/docker-gen-linux-armhf-$DOCKER_GEN_VERSION.tar.gz \
 && tar -C /usr/local/bin -xvzf docker-gen-linux-armhf-$DOCKER_GEN_VERSION.tar.gz \
 && rm docker-gen-linux-armhf-$DOCKER_GEN_VERSION.tar.gz
```

- then just built it:
```
docker build -t nginx-proxy:arm .
```
- and changed in my docker-conmpose.yml image from
```
  nginx-proxy:
    image: jwilder/nginx-proxy
```
to
```
  nginx-proxy:
    image: nginx-proxy:arm
```
And it starts working.

utrumo:
I made a little more concise decision:
[...]

Issue #1317, opened 2019-08-16
Start the Container under a arm32v7 environment


BMWfan:
Hello,

i have docker running on my raspberry pi 4b with 4 GB RAM.

I want to use the nginx-proxy but there was not separate arm32v7 container thatswhy i was pulling it from the git repository and changing the docker-compose.yml to:
#####################
version: '2'
services:
  nginx-proxy:
    build: .
    container_name: nginx-proxy
    ports:
      - "443:443"
    volumes:
      - /var/run/docker.sock:/tmp/docker.sock:ro
#####################

I started it with 'docker-compose up' and got the following message as latest:
`nginx-proxy    | WARNING: /etc/nginx/dhparam/dhparam.pem was not found. A pre-generated dhparam.pem will be used for now while a new one
nginx-proxy    | is being generated in the background.  Once the new dhparam.pem is in place, nginx will be reloaded.
nginx-proxy    | /app/docker-entrypoint.sh: line 34: /usr/local/bin/forego: cannot execute binary file: Exec format error
nginx-proxy    | /app/docker-entrypoint.sh: line 34: /usr/local/bin/forego: Success
nginx-proxy exited with code 1
`

Has someone a idea how i could fix this?

Greetings

Daniel

ahmedabt:
It requires arm32v7 docker image. it would be really nice if they supported ARM variant.

BMWfan:
@ahmedabt i totally agree it would be helpful i think not only for us. Many peoples are using a raspberry pi as a hardware plattform for example which requires a arm32v7 docker image. Do you know how we could proceed to get it?

ahmedabt:
@BMWfan No I don't know how to proceed, but I just had a look on the docker file and found it depends on Nginx, forego, and docker-gen.

Nginx already has arm32v7 version while forego and docker-gen are based on golang so theoretically you build armv7 version for both of them too.

If it is really needed by others as well, I can fork and maintain an armv7 version. Do you know few other people who would like to join?

BMWfan:
@ahmedabt here you find already opened issues which are mostly all depending on this behaviour:
[https://github.com/jwilder/nginx-proxy/issues?utf8=%E2%9C%93&q=raspberry](url)

IoTPlay:
Hello @jwilder, will we have an ARM docker impage on hub soon?
Thank you for your work.

valvanet:
bumb

o/

buchdag:
[...]

Issue #1297, opened 2019-06-29
Support for aarch64


Arsey:
I've tried nginx-proxy on Jetson Nano, where we have aarch64 and it doesn't work. Firstly I had to build the image manually but then when I start a container like `docker run --rm -it -p 80:80 -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy-local` it ends up with
```
WARNING: /etc/nginx/dhparam/dhparam.pem was not found. A pre-generated dhparam.pem will be used for now while a new one
is being generated in the background.  Once the new dhparam.pem is in place, nginx will be reloaded.
/app/docker-entrypoint.sh: line 34: /usr/local/bin/forego: cannot execute binary file: Exec format error
/app/docker-entrypoint.sh: line 34: /usr/local/bin/forego: Success
```

Knight1:
You have to download the forego arm version

buchdag:
Fixed as of #1470 + 4a1d5bfc58af35f5e90dcefb47b35a274c903150 + d120ce9fdebf09a6c7b90ecff7eda0b7d2760ef3

`arm64` and `arm/v7` images are availables from `jwilder/nginx-proxy` or the new registry `nginxproxy/nginx-proxy`

PR #1470, opened 2020-07-13
Enable multi-architecture docker image builds

tkw1536:
Previously, the Dockerfile downloaded 'docker-gen' and 'forego' binaries during build time. This caused a problem as it hard-coded the amd64 architecture for the images.

This PR updates both 'Dockerfile' and 'Dockerfile.alpine' to build the `forego` and `docker-gen` executables from source instead of downloading binaries directly. This is achieved using multi-stage builds [1]. Two separate stages first build the binaries, and are then copied over to the final stage.

The advantage of this change is two-fold: First, it enables building this image on architectures other than amd64. Second it adds trust by not adding external binaries to the docker image.

This modified version passes the tests both a linux desktop (`amd64`) as well as a raspberry pi (`armv7`) with some caveats:

- On `armv7`, a modified version of the `jwilder/docker-gen` image is required. See jwilder/docker-gen#327. 
- The `test_dhparam_is_generated_if_missing` test fails. This also doesn't currently pass on master. See #1468. 

See also:
- #1317 and #1370, #1297: These issues request an arm version of the image. Together with publishing arm images on DockerHub, this PR should fix all of them. 
- #1262, #1071: Both these PRs make a separate Dockerfile for the arm architecture. This PR is more general, in that it doesn't assume a specific architecture. In particular any architecture supported by the golang, debian and alpine base images can be used. 

[1] https://docs.docker.com/develop/develop-images/multistage-build/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,50 @@
-FROM nginx:1.19.3
+# setup build arguments for version of dependencies to use
+ARG NGINX_VERSION=1.19.3
+ARG GO_VERSION=1.14
+
+ARG DOCKER_GEN_VERSION=0.7.4
+ARG FOREGO_VERSION=0.16.1
+
+# Use a specific version of golang to build both binaries
+FROM golang:$GO_VERSION as gobuilder
+
+# Build docker-gen from scratch
+FROM gobuilder as dockergen
+
+# Download the sources for the given version
+ARG DOCKER_GEN_VERSION
+ADD https://github.com/jwilder/docker-gen/archive/${DOCKER_GEN_VERSION}.tar.gz sources.tar.gz
+
+# Move the sources into the right directory
+RUN tar -xzf sources.tar.gz && \
+   mkdir -p /go/src/github.com/jwilder/ && \
+   mv docker-gen-* /go/src/github.com/jwilder/docker-gen
+
+# Install the dependencies and make the docker-gen executable
+WORKDIR /go/src/github.com/jwilder/docker-gen
+RUN go get -v ./... && \
+   CGO_ENABLED=0 GOOS=linux go build -ldflags "-X main.buildVersion=${DOCKER_GEN_VERSION}" ./cmd/docker-gen
+
+# Build forego from scratch
+# Because this relies on golang workspaces, we need to use go < 1.8. 
+FROM gobuilder as forego
+
+# Download the sources for the given version
+ARG FOREGO_VERSION
+ADD https://github.com/jwilder/forego/archive/v${FOREGO_VERSION}.tar.gz sources.tar.gz
+
+# Move the sources into the right directory
+RUN tar -xzf sources.tar.gz && \
+   mkdir -p /go/src/github.com/ddollar/ && \
+   mv forego-* /go/src/github.com/ddollar/forego
+
+# Install the dependencies and make the forego executable
+WORKDIR /go/src/github.com/ddollar/forego/
+RUN go get -v ./... && \
+   CGO_ENABLED=0 GOOS=linux go build -o forego .
+
+# Build the final image
+FROM nginx:$NGINX_VERSION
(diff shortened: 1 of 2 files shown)


dustinrue:
I'd also really like to see this implemented

awehrfritz:
Excellent initiative and nice implementation! @jwilder is there a chance to get this merged into master or do you have any objections to this?

tkw1536:
I have yet to hear any feedback from @jwilder. I have nonetheless rebased this onto the current master, I only had to update the nginx version number.

plonkus:
This PR helped me get up and running on the Apple M1 processor. Thank you @tkw1536

kipusoep:
So any instructions to start using this on armv7?
I tried to build the docker image from your repo, which works, but still got errors running the image.
Then I saw your note about `jwilder/docker-gen`, how do I incorporate that fix in this fix as well?

tkw1536:
@kipusoep This image should just build and run on armv7, it did when I ran the tests under qemu locally. What is the error you are getting?

danil-smirnov:
I think this should be considered, very useful change

danil-smirnov:
@tkw1536 I believe there is an issue with this branch.

Original @jwilder code sets `DOCKER_GEN_VERSION` environment variable, which `letsencrypt-nginx-proxy-companion` product relies on, see

https://github.com/nginx-proxy/docker-letsencrypt-nginx-proxy-companion/blob/master/app/functions.sh#L241

After your multistage build this environment variable appeared unset, so `letsencrypt-nginx-proxy-companion` container fails.

tkw1536:
@danil-smirnov  I'm not convinced this is a problem. If you look at the diff of the PRs, neither of the old Dockerfiles set this variable and they seem to work fine on `linux/amd64` for many people.

danil-smirnov:
@tkw1536 Okay, let's look at the old Dockerfiles:

https://github.com/nginx-proxy/nginx-proxy/blob/master/Dockerfile#L21
https://github.com/nginx-proxy/nginx-proxy/blob/master/Dockerfile.alpine#L18

Both set this variable so `letsencrypt-nginx-proxy-companion` can rely on them.

I've added the line `ENV DOCKER_GEN_VERSION 0.7.4` to my custom image and `letsencrypt-nginx-proxy-companion` can started working after the fix.

tkw1536:
Oh you are correct. I forgot to read that `letsencrypt-nginx-proxy-companion` relies on this variable. 
I will add it to the branch.

tkw1536:
Added it to the branch, thanks @danil-smirnov.
Not all we have to do is wait for this PR to be merged.

severin-bruhat:
Hello, any update on this one?
Thanks

awehrfritz:
> Now all we have to do is wait for this PR to be merged.

We are waiting already for quite a while... more than half a year. It's a bit worrying that the last commit in this repository is from Oct 2020. Not sure who controls this repository or who else other than @jwilder has is able to merge PRs.

tkw1536:
Rebased onto the current master. With a new maintainer I would be happy to see this merged.

buchdag (changes requested):
I'd rather use directly `FROM golang:1.15.x AS gobuilder` rather than passing the base image version used with `ARG`, so that we can use `dependabot`. 

Other than that, looks good to me.

tkw1536:
I've updated this pull request with the changes requested.

buchdag:
@tkw1536 could you do the same for the `FROM nginx:$NGINX_VERSION`, for the same reason (detection of dependencies versions by dependabot) ?

tkw1536:
Done, updated. I don't think it's needed for `DOCKER_GEN_VERSION` or `FOREGO_VERSION`, as those are probably not covered by dependabot.

buchdag (approved):
LGTM

buchdag:
@tkw1536 if you spotted other arm related issue or PR that might be closed following the merging of this PR, don't hesitate to tag me on them.

PR #1477, opened 2020-07-22
Support no Docker proxy by env VIRTUAL_UPSTREAM

ccpaging:
If set env as below:

```
VIRTUAL_HOST=foo.example.com
VIRTUAL_UPSTREAM=192.168.1.8:8080,192.168.1.9:80 down
```

The default.conf's upstream sections is:

```
upstream foo.example.com {
    server 192.168.1.8:8080;
    server 192.168.1.9:80 down;
} 
```

VIRTUAL_HOST and VIRTUAL_UPSTREAM can be used in nginx-proxy container or any other fake container.

VIRTUAL_HOST do not support multiple hosts.
VIRTUAL_UPSTREAM support multiple address and deli with comma ','.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -178,6 +178,13 @@ server {
 upstream {{ $upstream_name }} {
 
 {{ range $container := $containers }}
+{{ $upstreams := or ($container.Env.VIRTUAL_UPSTREAM) "" }}
+{{ if ne $upstreams "" }}
+	# connect to extern VIRTUAL_UPSTREAM
+	{{ range $upstream := split $upstreams "," }}
+	server {{ $upstream }};
+	{{ end }}
+{{ else }}
 	{{ $addrLen := len $container.Addresses }}
 
 	{{ range $knownNetwork := $CurrentContainer.Networks }}
@@ -201,6 +208,7 @@ upstream {{ $upstream_name }} {
 			{{ end }}
 		{{ end }}
 	{{ end }}
+{{ end }}
 {{ end }}
 }
 


TimvdEijnden:
@jwilder @lemker @benchonaut Why isn't this being added?

This adds support to proxy containers using `network: host`

```
  homeassistant:
    image: homeassistant/raspberrypi4-homeassistant:stable
    container_name: home-assistant
    restart: always
    environment:
      - PUID=1000
      - PGID=1000
      - TZ=Europe/Amsterdam
      - VIRTUAL_HOST=hass.mydomain.com
      - VIRTUAL_PORT=8123
      - LETSENCRYPT_HOST=hass.mydomain.com
      - LETSENCRYPT_EMAIL=email@mydomain.com
    volumes:
      - ./home-assistant:/config
    ports:
      - 8123:8123
    network_mode: host
    ```

sausix:
I was just about to implement this and found it as open PR.

I'm also running the proxy on `network_mode: host` to allow connections to all services on all interfaces.

Some container ports can also be mapped to host's `127.0.0.1:8888` to avoid dynamic ip addresses problems.
And the proxy then can upstream to that local port without linking containers.

@ccpaging Great work. But indentation is messed up.

guysoft:
Any idea why Home Assistant needs --net=host? 
I see it has to set the proxy allowed, but wont that just work?

prashker:
> Any idea why Home Assistant needs --net=host? I see it has to set the proxy allowed, but wont setting that in the configuration be enough?

From HomeAssistant Docker page:

> Home Assistant can [discover](https://www.home-assistant.io/integrations/discovery/#mdns-and-upnp) and automatically configure [zeroconf](https://en.wikipedia.org/wiki/Zero-configuration_networking)/[mDNS](https://en.wikipedia.org/wiki/Multicast_DNS) and [UPnP](https://en.wikipedia.org/wiki/Universal_Plug_and_Play) devices on your network. In order for this to work you must create the container with --net=host.

buchdag:
@TimvdEijnden @sausix @prashker we're working on making `nginx-proxy` compatible with host networking on #2222, you feedback would be welcome if you're still running against this issue.

PR #1496, opened 2020-08-30
Supply max_fails=0 on all server entries

### LINKED ISSUE #1495: Including "down" server entries can lead to "no live upstreams while connecting to upstream"
When the host on which nginx-proxy is running has multiple networks, nginx-proxy's template generates an upstream on which some networks have an upstream marked as `down`, like:

```
upstream music.home.mati.ca {
				## Can be connected with "bridge" network
			# lms
			server 172.17.0.2:9000;
				# Cannot connect to network of this container
				server 127.0.0.1 down;
}
```

This is fine when everything works, but if the connectable upstream (on `bridge` in my example) has _any_ problems (like, say, a single 502 error), it triggers nginx's `max_fails`/`fail_timeout` behavior, and that upstream is marked as down for several seconds, with the error `no live upstreams while connecting to upstream`.

However, nginx actually has a fallback case that when an upstream only has _one_ `server` listed, then it skips the "no live upstreams" case, on the assumption that with only one upstream it's better to just keep trying it. 

Unfortunately, in the generated template above, nginx thinks it has _two_ `server`s listed, even though one is `down`, so the fallback case doesn't apply. (nginx should probably not do this when an upstream is marked `down`, but here we are.)

There's two options I think we could do to fix this:

* Don't output the `down` entry -- but I think this will produce an invalid nginx config if a container has _no_ valid endpoints, and we want nginx to keep working in that case, since it's presumably serving other containers; we'd have to handle the case where there are no upstreams at all and only output a single `down` upstream in that case
* Hardcode `max_fails` to 0 on every `server`, the idea being that we're not expecting there to be load-balancing behind `nginx-proxy`, just local upstream containers.

(I lean towards the second thing because it's easier templating, but I also suspect people use nginx-proxy for things I've never considered.)

wlonkly:
If nginx-proxy is run on a host whose docker config has multiple
networks, then it will generate an additional

  server XXX down;

entry in each upstream for networks on which the container _isn't_.
Having more than one server entry in an upstream invokes nginx's
max_fails/fail_timeout behavior, which will mark the only live
upstream server as down for 10s if it returns ONE error response.

Instead, we want nginx to keep retrying the live upstream server
regardless of what it returns, since there's nothing else to try.
Passing "max_fails=0" on each live server entry disables the
max_fails/fail_timeout behavior.

Fixes #1495.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -9,16 +9,16 @@
 		{{/* If we got the containers from swarm and this container's port is published to host, use host IP:PORT */}}
 		{{ if and .Container.Node.ID .Address.HostPort }}
 	# {{ .Container.Node.Name }}/{{ .Container.Name }}
-			server {{ .Container.Node.Address.IP }}:{{ .Address.HostPort }};
+	server {{ .Container.Node.Address.IP }}:{{ .Address.HostPort }} max_fails=0;
 		{{/* If there is no swarm node or the port is not published on host, use container's IP:PORT */}}
 		{{ else if .Network }}
 	# {{ .Container.Name }}
-	server {{ .Network.IP }}:{{ .Address.Port }};
+	server {{ .Network.IP }}:{{ .Address.Port }} max_fails=0;
 		{{ end }}
 	{{ else if .Network }}
 	# {{ .Container.Name }}
 		{{ if .Network.IP }}
-	server {{ .Network.IP }}:{{ .VirtualPort }};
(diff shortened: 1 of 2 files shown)


buchdag:
@wlonkly thanks for the PR

@pini-gh went the other route you mentioned in the original issue (#1495) :

> handle the case where there are no upstreams at all and **only output a single down upstream in that case**

So the original issue should be fixed by #1667

Issue #1020, opened 2018-01-06
No servers are inside upstream in /etc/nginx/conf.d/default.conf


valdemarrolfsen:
I am working on running my application over https using nginx and the letsencrypt-nginx-proxy-companion images, but I am experiencing some weird results. My docker-compose.yml looks like this:

```
version: '2.2'
services:
  ignite-api:
    restart: always
    build: ./ignite-api
    ports:
      - "8000"
    links:
      - postgres:postgres
      - rabbit:rabbit
      - worker:worker
    volumes:
      - /usr/src/app/
      - /usr/src/app/static
    env_file: .env
    command: /usr/local/bin/gunicorn api.wsgi:application -w 2 -b :8000 --reload
    networks:
      - api-net
      - nginx-proxy
    environment:
      VIRTUAL_PORT: 8000
      VIRTUAL_HOST: ignite-api.local
      LETSENCRYPT_HOST: example.com
      LETSENCRYPT_EMAIL: example@email.com

  postgres:
    ...

  rabbit:
    ...

  worker:
    ...

  nginx-proxy:
    image: jwilder/nginx-proxy
    ports:
      - "80:80"
      - "443:443"
    volumes:
      - "./nginx/vhost.d:/etc/nginx/vhost.d"
      - "./nginx/html:/usr/share/nginx/html"
      - "./nginx/certs:/etc/nginx/certs"
      - "/var/run/docker.sock:/tmp/docker.sock:ro"
    networks:
      - nginx-proxy

  letsencrypt-nginx-proxy-companion:
    image: jrcs/letsencrypt-nginx-proxy-companion
    volumes:
      - "/var/run/docker.sock:/var/run/docker.sock:ro"
    volumes_from:
      - "nginx-proxy"
    networks:
      - nginx-proxy

volumes:
  pgdata:
    driver: local
  rabbitmqdata:
    driver: local

networks:
  nginx-proxy:
    external:
      name: nginx-proxy
  api-net: 
    driver: bridge
```

When running  `docker-compose logs nginx-proxy` everything seems to be working fine with the following logs:

```
nginx-proxy_1                        | WARNING: /etc/nginx/dhparam/dhparam.pem was not found. A pre-generated dhparam.pem will be used for now while a new one
nginx-proxy_1                        | is being generated in the background.  Once the new dhparam.pem is in place, nginx will be reloaded.
nginx-proxy_1                        | forego     | starting dockergen.1 on port 5000
[...]

PR #1500, opened 2020-09-17
#1020 -- No upstream servers in ECS environments

thismatters:
Address #1020


buchdag:
This is caused by **jwilder/docker-gen**'s method of getting its own container ID, and should be fixed when `0.7.6` is released and we can upgrade the **docker-gen** version used in this project.

PR #1510, opened 2020-10-10
Nginx-Logs upgrade

tcarecolin:
Hello,

I have made an upgrade of the nginx-proxy Dockerfile in order to move access logs and error logs out of the container, in a volume mounted from remote host, so that a log collecting tool can easily be placed to setup a monitoring stack such as datadog or kibana

i have tested this successfully with filbeat as log collector and it fitted perfectly to my needs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,8 +30,12 @@ COPY . /app/
 WORKDIR /app/
 
 ENV DOCKER_HOST unix:///tmp/docker.sock
+ENV TERM xterm-mono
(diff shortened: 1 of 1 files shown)


deAtog:
This has zero chance of getting merged. Docker provides several options for managing logs from containers. See [Configuring logging drivers](https://docs.docker.com/config/containers/logging/configure/) for more information.

buchdag:
Hi. As @deAtog said, you should use Docker built in log management options to achieve this.

PR #1532, opened 2020-11-13
[Proposal] Ability to customize maintenance page

melya:
Nobody wants their service to be down, but when it happens, we want to make it in a more "happier" 🌈 way

Let's say instead of rendering
Screen Shot 2020-11-13 at 7 03 31 PM

We could render some human readable page, like
Screen Shot 2020-11-13 at 7 02 40 PM

**P.S.** _This is just a proposal.
I did a quick test and it works.

Would be great to here any feedback if it's relevant_

### DIFF
--- a/maintenance_internal.conf
+++ b/maintenance_internal.conf
@@ -1,0 +1,8 @@
+error_page 503 @maintenance;
+recursive_error_pages on;
+location @maintenance {
+    error_page 405 = /index.html;
+    add_header Cache-Control no-cache;
+    root /usr/share/nginx/html/maintenance;
+    rewrite ^(.*)$ /index.html break;
+}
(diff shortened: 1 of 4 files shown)


maintenance_internal.conf
```diff
@@ -0,0 +1,8 @@
+error_page 503 @maintenance;
+recursive_error_pages on;
+location @maintenance {
+    error_page 405 = /index.html;
+    add_header Cache-Control no-cache;
+    root /usr/share/nginx/html/maintenance;
```
melya, line 6: Maybe one page is enough and root is redundant.
Had thoughts if somebody needs a place for css

MyKrzysiu: Thanks! I really need it, and wait for merge it!

Prime1Code (approved):

buchdag:
Superseded by #2430, comments are welcome.

PR #1538, opened 2020-12-02
Update docker-entrypoint.sh comment on DHPARAM default key size

S0obi:
Default key size is not 2048 anymore but 4096 bits, reference : https://github.com/nginx-proxy/nginx-proxy/blob/master/generate-dhparam.sh#L4


thpham:
Hello,

could you please update the references in the main Readme, thanks.

buchdag:
Thanks @S0obi

Issue #1419, opened 2020-04-05
Request for documentation: IPv6 support + correct client ip addresses


tuxmainy:
Hi,
I had a rough time to get IPv6 working with the actual ip addresses of the clients and not the docker host IPv4 (yes, 4) address. The purpose of this issue is:

a) someone with more docker knowledge should check if I have done things right
b) Jason, I kindly ask you to add this to the documentation page of your nginx-proxy repository even if this is not fully nginx-proxy related but as your image provides a reverse proxy and lot of peoples are wondering what happens it would be nice if you could add this.

best regards
Daniel

------ Documentation -----
# IPv6 Support w/ actual client ip addresses
If you just bind the nginx HTTP(S) port to an IPv6 address on your Docker host you won't see the actual requesting IPv6 address of the client but the IPv4 address of Docker host (e.g. 172.17.0.1):
`nginx.1    | [2001:0db8::1] 172.17.0.1 - - [05/Apr/2020:14:46:56 +0000] "GET / HTTP/1.1" 200 197 "-" "Wget/1.20.3 (linux-gnu)"`

Following guide provides a way to see the actual requesting IPv6 client address by using a container which adds IPv6 NAT rules (e.g. 2001:0db8::abc):
`nginx.1    | [2001:0db8::1] 2001:0db8::abc - - [05/Apr/2020:16:19:20 +0000] "GET / HTTP/1.1" 200 197 "-" "Wget/1.20.3 (linux-gnu)"`

Thus, nginx-proxy will be able to set the actual client ip address in the HTTP_X_FORWARDED_FOR header which could be used by your web application.

## Docker configuration
Disable Docker userland-proxy and enable IPv6 support by using [ULA](https://en.wikipedia.org/wiki/Unique_local_address) adresses to provide container isolation

**/etc/docker/daemon.json**
`
"userland-proxy": false,
"ipv6": true,
"fixed-cidr-v6": "fd00:c0fe:babe::/48"
`
## IPv6 NAT
This will add IPv6 NAT support (like the one of IPv4 which is directly implemented by Docker). If you want to have a discussion about IPv6 NAT have a look at the [project repository](https://github.com/robbertkl/docker-ipv6nat/blob/master/README.md#nat-on-ipv6-are-you-insane).

[...]

PR #1557, opened 2021-02-06
Remove scoped IPv6 addresses from resolvers (fix #1424)

### LINKED ISSUE #1424: IPv6 link local interface scoped address breaks nginx config load
Related: #938

My host's `/etc/resolv.conf` looks like this:
```
# Generated by NetworkManager
search mydomain.net
nameserver 192.168.1.1
nameserver fe80::728b:cdff:fe86:a470%enp2s0
nameserver 2601:some:vlid:ipv6:addr::1
```

Which gets copied into the container.
Because of this:
https://github.com/nginx-proxy/nginx-proxy/blob/3cbc5417b7ac38037a9ec00fe80e0996d7da3d75/docker-entrypoint.sh#L22-L23
they get copied into the `nginx.conf`

When nginx loads it, it fails, and enters a restart loop:
```log
dockergen.1 | 2020/04/19 19:19:15 Running 'nginx -s reload'
dockergen.1 | 2020/04/19 19:19:15 Error running notify command: nginx -s reload, exit status 1
nginx.1    | 2020/04/19 19:19:46 [emerg] 31#31: invalid IPv6 address in resolver "[fe80::728b:cdff:fe86:a470%enp2s0]" in /etc/nginx/conf.d/default.conf:36
```

It seems that nginx can't use an address of the form `[ipv6%interface]`.

I have a work around like so:
```bash
export RESOLVERS=$(echo ${RESOLVERS} | sed 's/\[[a-f0-9:]*\%[a-zA-Z0-9]*\] //')
```
which removes the entry entirely. But maybe there is a way to fix it without removing it?

joe-p:
This is a fix for #1424. It removes any scoped IPv6 addresses from RESOLVERS in docker-entrypoint.sh. The alternative fix is to remove the interface from the address, but this could result in the wrong interface being used as a resolver which seems more dangerous and less intuitive than simply removing the address all together. 

I also updated the README addressing this change and included some details on IPv6 NAT in response to #1419.

`make test` has one failure, `test_dhparam_is_generated_if_missing`, but this also fails on the master branch.


buchdag:
Regarding IPv6 NAT, changes are coming to Docker:

> I can confirm that Docker 20.10.6's ipv6nat implementation works, and it seems to work exactly like how this container was doing it. The only difference I have seen is that the `docker ps` command now shows that the ports are mapped for both IPv4 and IPv6. The downside being that "experimental" mode needs to be turned on.

https://github.com/robbertkl/docker-ipv6nat/issues/65

PR #1558, opened 2021-02-09
nginx.tmpl: use closest vhost.d and htpasswd files (#1309)

### LINKED ISSUE #1309: Check closest host match when looking for vhost.d/config files before checking default
This would be the same thing that happens when looking for certificates, and it would allow wildcard host names to find a custom config file.

jjakob:
This changes the logic for finding a matching file to the same one used
by certificates. This allows wildcard host names to find a matching
config file.

The container prints "Template error: open /etc/nginx/htpasswd: no such
file or directory" if /etc/nginx/htpasswd doesn't exist, but this
doesn't impact functionality. A fix would be to make sure that directory
is always created.

Fixes #1309.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -280,6 +280,7 @@
 
 {{- define "location" }}
     {{- $vpath := .VPath }}
+    {{- /* TODO: largest suffix match location_override (same as for include_file and htpasswd_file) */}}
     {{- $override := printf "/etc/nginx/vhost.d/%s_%s_location_override" .Host (sha1 .Path) }}
     {{- if and (eq .Path "/") (not (exists $override)) }}
         {{- $override = printf "/etc/nginx/vhost.d/%s_location_override" .Host }}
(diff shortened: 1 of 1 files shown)


jjakob:
I'd like this functionality to be merged, as I still use it. But I'm not satisfied with using the 'closest' function, as it matches anywhere inside the filename string, it isn't anchored to the end, so e.g. a file 'b.c' will match 'a.b.c.d'.

I'd want something like a 'closestSuffix' function, which would anchor the match at the end of the string, so for example, only 'b.c.d' or 'c.d' will match 'a.b.c.d'.
This would allow setting vhost.d and htpasswd files for VIRTUAL_HOST with an arbitrary number of subdomains.

I think the only way to implement this is by adding a 'closestSuffix' function to docker-gen, but I don't have time to set up a build environment for it. Perhaps someone could implement the function there, and I could update this pull request to use it?

I'm envisioning something like:
``` go
# same as arrayClosest but anchored to the end of the string
func arrayClosestSuffix(values []string, input string) string {
        best := ""
        for _, v := range values {
                if strings.HasSuffix(input, v) && len(v) > len(best) {
                        best = v
                }
        }
        return best
}
```
This could be further improved by checking that the first subdomain is not a partial match, e.g. 'bc.def' should not match 'abc.def' (with the above function, it will), but at first, it's not necesarry.

In the meantime I will rebase to main and update the PR.
@buchdag

buchdag:
@jjakob I'm not opposed at all to add a function to docker-gen if this can help with a feature here, I'll look into it.

PR #1584, opened 2021-04-08
Added ssl client support when req'd files present

Sy14r:
This pull request resolves nginx-proxy/nginx-proxy#278 by modifying the nginx.tmpl to add support for either per host or global certificates and certificate revocation lists for client certificate verification.

To use a user must ensure that either a certificate authority is uploaded into /etc/nginx/certs when they start their proxied containers. To apply a ca globally the file must be called ca.crt and the crl must be ca.crl. To limit client certificate verification to a specific FQDN then the CA and optional CRL must be uploaded to /etc/nginx/certs with their name in the format of fqdn.ca.crt and fqdn.ca.crl (IE test.example.com.ca).

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -294,6 +294,26 @@ server {
 	ssl_dhparam {{ printf "/etc/nginx/certs/%s.dhparam.pem" $cert }};
 	{{ end }}
 
+	{{/* SSL Client Certificate Validation */}}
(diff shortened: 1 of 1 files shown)


Sy14r:
Just noticed this is a dup... not sure why the other PR hasn't been accepted yet since there's no blocking comments or negative feedback

PR #1594, opened 2021-04-19
Support for path-based routing

AlexanderLieret:
This is a rebase of #1083 with some improvements.
- `VIRTUAL_PATH`: route using this path
- `VIRTUAL_DEST`: rewrite the query path (optional)
- Support for custom config snippets files
- Add test cases

NOTE: This requires another pull request to function  jwilder/docker-gen#343 to function.

PR #1011, opened 2017-12-25
Allow VIRTUAL_PATH routing (sub-VIRTUAL_HOST)

trowj:
This is really an update to the fork at 'https://github.com/gregsymons/docker-nginx-proxy' that I believe I fully updated to current master and merged the functionality.  I will honestly say that I can't get the tests to run  due to some kind of python bug I think (I can't run them on this current repo either so it's not a change I made at least) so I cannot be sure everything is still passing, but I did my best in comparison locally between versions as to what was generated for default.conf

PR #599, opened 2016-10-07
Add support for path-based routing

gregsymons:
This PR adds the ability to serve multiple containers through the same vhost using path-based routes. This can be useful if you have several independent services that you want make appear as one.

The path-based proxies assume that the incoming request uri should be mapped to the root of the container, and should be rewritten to remove the prefix on the proxy host. This PR does not provide support for either passing the full request uri through to the backend nor for requesting any context other than the root of the backend container.

In other words, the incoming path is always passed to `http://<backend-address>:<backend-port>/`.

PR #1607, opened 2021-04-29
Support for path-based routing

buchdag:
This is a work in progress combination of #599 + #1011 + #1083 + #1594 with proper authoring credits.

- `VIRTUAL_PATH`: route using this path
- `VIRTUAL_DEST`: rewrite the query path (optional)
- Support for custom config snippets files
- Add test cases

NOTE: This requires another pull request to function  jwilder/docker-gen#343 to function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,12 @@ For each host defined into `VIRTUAL_HOST`, the associated virtual port is retrie
 
 You can also use wildcards at the beginning and the end of host name, like `*.bar.com` or `foo.bar.*`. Or even a regular expression, which can be very useful in conjunction with a wildcard DNS service like [xip.io](http://xip.io), using `~^foo\.bar\..*\.xip\.io` will match `foo.bar.127.0.0.1.xip.io`, `foo.bar.10.0.2.2.xip.io` and all other given IPs. More information about this topic can be found in the nginx documentation about [`server_names`](http://nginx.org/en/docs/http/server_names.html).
 
+### Path-based Routing
+
+You can have multiple containers proxied by the same `VIRTUAL_HOST` by adding a `VIRTUAL_PATH` environment variable containing the absolute path to where the container should be mounted. For example with `VIRTUAL_HOST=foo.example.com` and `VIRTUAL_PATH=/api/v2/service`, then requests to http://foo.example.com/api/v2/service will be routed to the container. If you wish to have a container serve the root while other containers serve other paths, make give the root container a `VIRTUAL_PATH` of `/`.  Unmatched paths will be served by the container at `/` or will return the default nginx error page if no container has been assigned `/`.
+
+The full request URI will be forwarded to the serving container in the `X-Forwarded-Path` header.
+
 ### Multiple Networks
 
 With the addition of [overlay networking](https://docs.docker.com/engine/userguide/networking/get-started-overlay/) in Docker 1.9, your `nginx-proxy` container may need to connect to backend containers on multiple networks. By default, if you don't pass the `--net` flag when your `nginx-proxy` container is created, it will only be attached to the default `bridge` network. This means that it will not be able to connect to containers on networks other than `bridge`.
@@ -365,6 +371,7 @@ proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 proxy_set_header X-Forwarded-Proto $proxy_x_forwarded_proto;
 proxy_set_header X-Forwarded-Ssl $proxy_x_forwarded_ssl;
 proxy_set_header X-Forwarded-Port $proxy_x_forwarded_port;
+proxy_set_header X-Forwarded-Path $request_uri;
 
 # Mitigate httpoxy attack (see README for details)
 proxy_set_header Proxy "";
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -47,6 +47,92 @@
 	{{ end }}
 {{ end }}
 
+{{ define "location" }}
+location {{ .Path }} {
+	{{ if eq .Proto "uwsgi" }}
+		include uwsgi_params;
+		uwsgi_pass {{ trim .Proto }}://{{ trim .Upstream }};
+	{{ else if eq .Proto "fastcgi" }}
+		root   {{ trim .Vhostroot }};
+		include fastcgi.conf;
+		fastcgi_pass {{ trim .Upstream }};
+	{{ else if eq .Proto "grpc" }}
+		grpc_pass {{ trim .Proto }}://{{ trim .Upstream }};
+	{{ else }}
+		proxy_pass {{ trim .Proto }}://{{ trim .Upstream }}/;
+	{{ end }}
+
+	{{ if (exists (printf "/etc/nginx/htpasswd/%s" .Host)) }}
+		auth_basic	"Restricted {{ .Host }}";
+		auth_basic_user_file	{{ (printf "/etc/nginx/htpasswd/%s" .Host) }};
+	{{ end }}
+
+	{{ if (exists (printf "/etc/nginx/vhost.d/%s_location" .Host)) }}
+		include {{ printf "/etc/nginx/vhost.d/%s_location" .Host}};
+	{{ else if (exists "/etc/nginx/vhost.d/default_location") }}
+		include /etc/nginx/vhost.d/default_location;
+	{{ end }}
+}
+{{ end }}
+
+{{ define "upstream-definition" }}
+	{{ $networks := .Networks }}
+	{{ $debug_all := .Debug }}
+	upstream {{ .Upstream }} {
+		{{ $server_found := "false" }}
+		{{ range $container := .Containers }}
+        {{ $debug := (eq (coalesce $container.Env.DEBUG $debug_all "false") "true") }}
+        {{/* If only 1 port exposed, use that as a default, else 80 */}}
+        {{ $defaultPort := (when (eq (len $container.Addresses) 1) (first $container.Addresses) (dict "Port" "80")).Port }}
+        {{ $port := (coalesce $container.Env.VIRTUAL_PORT $defaultPort) }}
+        {{ $address := where $container.Addresses "Port" $port | first }}
+        {{ if $debug }}
+        # Exposed ports: {{ $container.Addresses }}
+        # Default virtual port: {{ $defaultPort }}
+        # VIRTUAL_PORT: {{ $container.Env.VIRTUAL_PORT }}
+                {{ if not $address }}
+        # /!\ Virtual port not exposed
+                {{ end }}
+        {{ end }}
+			{{ range $knownNetwork := $networks }}
+				{{ range $containerNetwork := $container.Networks }}
+					{{ if (and (ne $containerNetwork.Name "ingress") (or (eq $knownNetwork.Name $containerNetwork.Name) (eq $knownNetwork.Name "host"))) }}
+        ## Can be connected with "{{ $containerNetwork.Name }}" network
+                                {{ if $address }}
+                                        {{/* If we got the containers from swarm and this container's port is published to host, use host IP:PORT */}}
+                                        {{ if and $container.Node.ID $address.HostPort }}
+                                                {{ $server_found = "true" }}
+        # {{ $container.Node.Name }}/{{ $container.Name }}
+        server {{ $container.Node.Address.IP }}:{{ $address.HostPort }};
+                                        {{/* If there is no swarm node or the port is not published on host, use container's IP:PORT */}}
+                                        {{ else if $containerNetwork }}
+                                                {{ $server_found = "true" }}
+        # {{ $container.Name }}
+        server {{ $containerNetwork.IP }}:{{ $address.Port }};
+                                        {{ end }}
+                                {{ else if $containerNetwork }}
+        # {{ $container.Name }}
+                                        {{ if $containerNetwork.IP }}
+                                                {{ $server_found = "true" }}
+        server {{ $containerNetwork.IP }}:{{ $port }};
+                                        {{ else }}
+        # /!\ No IP for this network!
+                                       {{ end }}
+						{{ end }}
+					{{ else }}
+        # Cannot connect to network '{{ $containerNetwork.Name }}' of this container
+					{{ end }}
+				{{ end }}
+			{{ end }}
+		{{ end }}
+		{{/* nginx-proxy/nginx-proxy#1105 */}}
+		{{ if (eq $server_found "false") }}
+        # Fallback entry
+        server 127.0.0.1 down;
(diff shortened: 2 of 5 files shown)


buchdag:
@AlexanderLieret okay this looks promising, we've got only two failing tests. I've reset my local branch to your last commit and fixed an error I did on a merge conflict resolution.

Next step is having jwilder/docker-gen#343 merged then taking a look at the two failing tests.

Given the radical changes this PR would make to nginx.tmpl, merging it will probably cause non trivial merge conflict on most if not all other feature PR. I have to think a bit about the best strategy for this, for now I came up with:

1. doing triage on the PR and merging as much low hanging fruits as possible before merging this one to `main`.
2. merging this PR to a new `edge` branch and release an `edge` version of the container, then merge it to `main` later, after it's been sufficiently field tested.

I'm leaning more toward 2.

AlexanderLieret:
@buchdag I had a look at those failed tests. The first was the recent change of the CI test pipeline.
The other does need a bit more troubleshooting.

Option 2 will push the merge conflicts to a later point. Being able to field test should be an advantage.

buchdag:
So the sole failing test is:

```
=================================== FAILURES ===================================
_ test_non_matching_host_is_503[web4.whatever.nginx-proxy.regexp-to-infinity-and-beyond] _

docker_compose = <docker.client.DockerClient object at 0x7f8c5a603130>
nginxproxy = <conftest.requests_for_docker object at 0x7f8c5a4399d0>
host = 'web4.whatever.nginx-proxy.regexp-to-infinity-and-beyond'

    @pytest.mark.parametrize("host", [
        "unexpected.nginx-proxy.tld",
        "web4.whatever.nginx-proxy.regexp-to-infinity-and-beyond"
    ])
    def test_non_matching_host_is_503(docker_compose, nginxproxy, host):
        r = nginxproxy.get(f"http://{host}/port")
>       assert r.status_code == 503, r.text
E       AssertionError: answer from port 84
E         
E       assert 200 == 503
E         +200
E         -503

test_wildcard_host.py:32: AssertionError
----------------------------- Captured stderr call -----------------------------
DEBUG:DNS:resolving domain name ('web4.whatever.nginx-proxy.regexp-to-infinity-and-beyond', 80, <AddressFamily.AF_UNSPEC: 0>, <SocketKind.SOCK_STREAM: 1>)
DEBUG:DNS:nginx_proxy_dns_resolver('web4.whatever.nginx-proxy.regexp-to-infinity-and-beyond')
INFO:DNS:resolving domain name 'web4.whatever.nginx-proxy.regexp-to-infinity-and-beyond' as IP address 172.17.0.2 of nginx-proxy container test_sut_1
------------------------------ Captured log call -------------------------------
DEBUG    DNS:conftest.py:218 resolving domain name ('web4.whatever.nginx-proxy.regexp-to-infinity-and-beyond', 80, <AddressFamily.AF_UNSPEC: 0>, <SocketKind.SOCK_STREAM: 1>)
DEBUG    DNS:conftest.py:169 nginx_proxy_dns_resolver('web4.whatever.nginx-proxy.regexp-to-infinity-and-beyond')
INFO     DNS:conftest.py:177 resolving domain name 'web4.whatever.nginx-proxy.regexp-to-infinity-and-beyond' as IP address 172.17.0.2 of nginx-proxy container test_sut_1
--------------------------- Captured stderr teardown ---------------------------
INFO:root:docker-compose -f /home/runner/work/nginx-proxy/nginx-proxy/test/test_wildcard_host.yml down
---------------------------- Captured log teardown -----------------------------
INFO     root:conftest.py:274 docker-compose -f /home/runner/work/nginx-proxy/nginx-proxy/test/test_wildcard_host.yml down
```

Maybe tied to 343cdd0 ?

buchdag:
Also I'm still not sure how I'll handle the final commit history that'll get merged.

The current one, while keeping proper authoring credit for everyone's specific commits, is super messy and forces to merge the target branch back into the source branch. I'd like to avoid that if at all possible.

I'm thinking about the following: 
- a single commit for `nginx.tmpl`
- a single commit for `README.md`
- a single commit for 343cdd02ac35cf57c070aa5de0ed86e48efe3c04
- a single commit for the tests you added @AlexanderLieret 

Each commit being authored by all the relevant contributors.

AlexanderLieret:
> So the sole failing test is:
Maybe tied to 343cdd0 ?

This commit does break the test which is now failing. Its purpose is to test forced full regex matches instead of partial matches. The commit message does mention #124 which is superseded. This commit should be removed.

> Also I'm still not sure how I'll handle the final commit history that'll get merged.
Each commit being authored by all the relevant contributors.

The main work was done by @gregsymons. The others and I rebased it on the current main branch and added some fixes. 

I added some QoL improvements to the "new" feature. Most of them were from the individual PRs and #254.

buchdag:
I'm not completely sold on the `VIRTUAL_DEST` behaviour.

I think most people who wanted this feature would expect that this:

`docker run -d --name app1 --env  VIRTUAL_HOST=domaint.tld --env VIRTUAL_PATH:/path nginx`

would mean this by default:

`http://domaint.tld/path` => `http:/app1/`

It seems to me that most of the work and discussions that happened around this feature was based on the asumption that the virtual path would route to `/` on the proxied container by default.

buchdag:
Scratch the last comment, I misread your code.

buchdag:
Okay now I'm lost.

The doc say the following ...

>     $ docker run -d -p 80:80 -v /var/run/docker.sock:/tmp/docker.sock:ro nginxproxy/nginx-proxy
>     $ docker run -d -name service1 -e VIRTUAL_HOST=example.tld -e VIRTUAL_PATH=/foo jwilder/whoami
>     $ docker run -d -name service2 -e VIRTUAL_HOST=example.tld -e VIRTUAL_PATH=/bar -e VIRTUAL_DEST=/ jwilder/whoami
> In this example, request to `http://example.tld/foo` are proxied to `http://service1/foo` and request to `http://example.tld/bar` to `http://service2/`.

... but the code does ...

```
{{ $paths := groupBy $containers "Env.VIRTUAL_PATH" }}
{{ $nPaths := len $paths }}

[...]

{{ if eq $nPaths 0 }}
    {{ template "location" (dict "Path" "/" "Proto" $proto "Upstream" $upstream_name "Host" $host "VHostRoot" $vhost_root "Dest" "") }}
{{ else }}
    [...]
{{ end }}
```

... which unless I'm mistaken means an unset `VIRTUAL_DEST` defaults to `/` and that the example is wrong ?

buchdag:
I've cleaned up the commit history to get something a bit more legible and removed the code change that made the test fail.

I haven't added back the support for custom location files per-VIRTUAL_HOST VIRTUAL_PATH yet.

I noticed two things in the README history of this feature:

> If you wish to have a container serve the root while other containers serve other paths, make give the root container a `VIRTUAL_PATH` of `/`. **Unmatched paths will be served by the container at `/`** or will return the default nginx error page if no container has been assigned `/`.

The part in bold does not work.

> When you want to forward many paths to the same container you can set the variable using a regular expression. For example `VIRTUAL_PATH=~^/(path1|path2)` will answer to requests that start with /path1 or /path2.

This doesn't work either.

Both might have been working at the time their respective documentations were written then broken by subsequent commits. The first was directly tested in @gregsymons original tests.

AlexanderLieret:
I will have another look at both problems. It seems like some features have been removed while rebasing on the current main.

While improving the test cases I should move all virtual_path tests into a single directory.

buchdag:
@AlexanderLieret let me know if you want me to merge this to the `dev` branch and push the image to Dockerhub under the `dev` tag.

I'm not sold on the fact that the custom location files per-VIRTUAL_HOST VIRTUAL_PATH needs a `sha1` of the virtual path in their name, but I don't know if there is another way to do it.

AlexanderLieret:
@buchdag I have finished rebasing @gregsymons original commit to see if other stuff had been changed in the rebase chain. The new attempt is in a [new branch](/AlexanderLieret/nginx-proxy/tree/vpath-retry). An updated documentation is needed.

I have found a bug which prevents dockergen from updating the generated config file when another container is added to or removed from a domain. This could explain your failed attempts at manual testing. The currently failing test case is caused by this bug.

> When you want to forward many paths to the same container you can set the variable using a regular expression. For example `VIRTUAL_PATH=~^/(path1|path2)` will answer to requests that start with /path1 or /path2.

This sentence was added by @rodrigoaguilera and only works if the routable-path is not to be overwritten by the proxypass (`VIRTUAL_DEST=""`). It is possible to overwrite the matched path with a regex. This is also one test case.

> I'm not sold on the fact that the custom location files per-VIRTUAL_HOST VIRTUAL_PATH needs a `sha1` of the virtual path in their name, but I don't know if there is another way to do it.

The hash is used to sanitize the `VIRTUAL_PATH` to be used as part of the file name. This is similar to how this works with regex domains.

A suitable default value for `VIRTUAL_DEST` is a tough decision and will need to be covered in the upcoming documentation. In short this depends on the handling of subpaths by the proxied application.

KunalSolanke:
Somehow I am not able to route all the paths statring with /test properly inside the container.When I set VIRTUAL_DEST=""
if request is sent to /test/work container recieves request at same /test/work path ,but I need to be at /work inside the target container.So I thought lets make VIRTUAL_PATH=/test/ and VIRTUAL_DEST=/ but with this only /test requests reach container other like /test/work are rejected and handled by / location block.

buchdag:
@KunalSolanke just to be sure: are you using the specific code from this PR ? It's based on but not the same as #599

KunalSolanke:
@buchbag yeah ,initially I landed on that PR and tried that,but later found this one is the latest one so I tried this code from this one.Sorry for that.I later realized that the way I wanted this to work wouldn't be helpful rather the default behaviour is better .So you can safely ignore previous comment.

KunalSolanke:
Sorry for bothering again but I am stuck with another things again .This may require a bit of understanding of django.
I am trying VIRTUAL_PATH=/test,VIRTUAL_PROTO=http


Screenshot from 2021-05-27 12-50-15

These are my django urls configuration.So the thing is

Now the issue:
When i navigate to "/test/swagger/" with ending slash it works fine but when I do "/test/swagger"(without end slash) nginx redirects to "/swagger/",request doesn't even reach container.It like it is not even recognizing the url.
This is my issue .But now the weird part,

when I do /test/debug(without end slash) nginx somehow redirects to /test/debug/ and I am sure this redirect is done by nginx and not by django.
No idea how that is happening.

A basic sol for my problem would be to redirect everything from nginx to /uri to /uri/ ,but I have no idea how to do it here.



This is autogenrated nginx.conf file
``nginx

	# localhost/test
	upstream localhost-upstream-f133a4599372cf531bcdbfeb1116b9afe8d09b4f {
						## Can be connected with "nginx-proxy" network
			# evy_web_1
			server 172.19.0.3:8000;
	}
        server {
	        listen 80 ;
                server_name localhost;
                    location /test {
		                    proxy_pass http://localhost-upstream-f133a4599372cf531bcdbfeb1116b9afe8d09b4f;
                    }
	            location / {
		            return 404;
	            }
        }
``

AlexanderLieret:
@KunalSolanke all of these redirect are done by your django application.

With `VIRTUAL_PATH=/test` all request which begin with `/test` are proxied to your application. Nginx does not do any redirects in this situation. You need to test your application first that it is able to work on the subpath first.

The only part I did see where Nginx propably was redirecting was something like `VIRTUAL_PATH=/foo/ VIRTUAL_DEST=/` and then only redirecting `/foo` to `/foo/`. It should only add a slash when the resulting proxy request has a slash in the end.

KunalSolanke:
@AlexanderLieret The reason why I said I am sure nginx is doing those redirects so requests don't even reach the target in those cases. Ok now I know this might me naive observation, so there is middleware called commonMiddleware in django which handles these redirects, I just copied that middleware and made my own with debug statements and yeah the requests didn't reach there if I don't append slash.It might be so that my django application is indeed redirecting,I will try to debug more in my django app. Thanks for your reply😥😥.

AlexanderLieret:
@buchdag I have finished redoing the commits. The new test cases are covering more parts of the code. Furthermore the documentation should be easier to read and contain more information. This is still in the [new branch](/AlexanderLieret/nginx-proxy/tree/vpath-retry).

The failing test case covers one previously not covered scenario. After removing one container from the configuration file is generated but the reload is not triggered by docker-gen. This only happens if at least one path for this subdomain is remaining. A workaround is to manually restart the nginx container or triggering the reload.

AlexanderLieret:
@buchdag It's ready to push to dev.

buchdag:
@AlexanderLieret ideally do you think you could split [this commit](https://github.com/AlexanderLieret/nginx-proxy/commit/3cac2ddea2abd6be1a7c5de93ec52a6acfecd52e) into a commit for each separate features ?

buchdag:
To make things easier I propose doing the following:

1) limiting the current PR to:
    - 31e72fb (which is identical to your branch's efe03ef1f5d7b2e1bac49c7f5edd2b581b09a3ef minus the fact that it is based upon `main` instead of `dev`)
    - ac8c2f95d1ecf7584f13af78df440598c57cc4d0 (tests for the base path based routing feature)
    - 6490246 (docs, reworked to cover only the base path based routing feature)
2) merging this PR to `dev` and pushing the image to Dockerhub for field tests.
3) creating another branch on your fork  for the additional features and tests in 3cac2ddea2abd6be1a7c5de93ec52a6acfecd52e, based on `dev` and not on `main` and with a commit for each feature.
4) opening a PR targeting `dev` for those additional features

eugene-katsevman:
Any update on this one?

buchdag:
@eugene-katsevman I'll resume work on this next week.

warren-ru:
I get

> Template error: template: nginx.tmpl:365:23: executing "nginx.tmpl" at <$paths>: wrong type for value; expected map[string]string; got map[string][]interface {}

when I try to use VIRTUAL_PATH with "/" symbols in env_file

buchdag:
@warren-ru this PR hasn't been merged and require the latest version of docker-gen to work.

AlexanderLieret:
@buchdag I did split the commits into smaller ones. I am not sure what went wrong in your attempt.

My branch `virtual-path` has the original code rebased/merged with the current main and split into 3 commits (implementation, tests and documentation). My other branch `virtual-path-additions` contains my additions in separate commits.

buchdag:
> I am not sure what went wrong in your attempt.

@AlexanderLieret me neither, I'll try to clean this up based on your branch tomorrow.

buchdag:
@AlexanderLieret PR merged to `dev`, I'll start to work on publishing the dev image to DockerHub immediately.

Thanks for the excellent work.

buchdag:
Further tests and remarks in #1733

Issue #1132, opened 2018-05-08
server ip is marked as down when using different internal networks


gtaspider:
Got a docker-compose file which will have multiple internal docker networks. The proxy seems to find the correct ip and set it to the correct upstream but the server is marked as down, eventhough the commet says nginx was able to connect to the right network:

default.conf:
```
#[...]
# cloud.website.com
upstream cloud.website.com {
				# Cannot connect to network of this container
				server 127.0.0.1 down;
				## Can be connected with "test_net-dashboard" network
		# test_nextcloud_1
			server 172.31.0.4 down;
}
server {
	server_name cloud.website.com;
	listen 80 ;
	access_log /var/log/nginx/access.log vhost;
	location / {
		proxy_pass http://cloud.website.com;
	}
}
```

When I remove the `down`, safe it and then reload nginx with `docker exec -it test_proxy_1 nginx -s reload` it works like a charm. But if a container starts/stops the file will be rewritten (this is why I want to use this tool..) 

I created a simple docker-compose file, to show the problem: 
```
version: '3'

services:
  db:
    image: postgres:alpine
    restart: always
    volumes:
      - db:/var/lib/postgresql/data
    environment:
      - POSTGRES_PASSWORD=
      - POSTGRES_DB=nextcloud
      - POSTGRES_USER=usr
    networks:
      - net-dashboard
  nextcloud:  
    image: nextcloud:apache
    restart: always
    volumes:
      - nextcloud:/var/www/html
    environment:
      - VIRTUAL_HOST=cloud.website.com
      - POSTGRES_HOST=db
      - MM_USERNAME=usr
      - MM_PASSWORD=
      - MM_DBNAME=nextcloud
    depends_on:
      - db
    networks:
     - net-dashboard
  #NGINX Proxy
  proxy:
    image: jwilder/nginx-proxy:alpine
    restart: always
    ports:
      - 80:80
    volumes:
#      - certs:/etc/nginx/certs:ro
      - vhost.d:/etc/nginx/vhost.d
      - conf.d:/etc/nginx/conf.d
      - html:/usr/share/nginx/html
      - /var/run/docker.sock:/tmp/docker.sock:ro
    networks:
      - net-dashboard
      - default
 

volumes:
  db:
  nextcloud:
#  certs:
  vhost.d:
  conf.d:
  html:


networks:
  net-dashboard:
    internal: true
```

This seems to be a bug, so I posted it here.
Thanks in advance,
Spider

[...]

PR #1609, opened 2021-04-30
Honor VIRTUAL_PORT + DEBUG flag + fallback entry

### LINKED ISSUE #1105: nginx-proxy chokes on containers from invisible networks

pini-gh:
The VIRTUAL_PORT environment variable should always be honored.
Even when the related port is not exposed.
Fix for #1132.

This commit also add the DEBUG environment variable which enables more
verbose comments in the nginx comfiguration file to help troubleshooting
unreachable containers.

Finaly it fixes #1105 as well by defining only one
fallback entry per upstream block.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -2,24 +2,25 @@
 
 {{ $external_http_port := coalesce $.Env.HTTP_PORT "80" }}
 {{ $external_https_port := coalesce $.Env.HTTPS_PORT "443" }}
+{{ $debug_all := $.Env.DEBUG }}
 
 {{ define "upstream" }}
 	{{ if .Address }}
 		{{/* If we got the containers from swarm and this container's port is published to host, use host IP:PORT */}}
 		{{ if and .Container.Node.ID .Address.HostPort }}
-			# {{ .Container.Node.Name }}/{{ .Container.Name }}
+	# {{ .Container.Node.Name }}/{{ .Container.Name }}
 			server {{ .Container.Node.Address.IP }}:{{ .Address.HostPort }};
 		{{/* If there is no swarm node or the port is not published on host, use container's IP:PORT */}}
 		{{ else if .Network }}
-			# {{ .Container.Name }}
-			server {{ .Network.IP }}:{{ .Address.Port }};
+	# {{ .Container.Name }}
+	server {{ .Network.IP }}:{{ .Address.Port }};
 		{{ end }}
 	{{ else if .Network }}
-		# {{ .Container.Name }}
+	# {{ .Container.Name }}
 		{{ if .Network.IP }}
-			server {{ .Network.IP }} down;
+	server {{ .Network.IP }}:{{ .VirtualPort }};
 		{{ else }}
-			server 127.0.0.1 down;
+	# /!\ No IP for this network!
 		{{ end }}
 	{{ end }}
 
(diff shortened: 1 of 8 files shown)


buchdag:
@pini-gh after some thoughts, I think this PR in its current state might cause might issues than it solves.

First I don't think we should break the previous "If only 1 port exposed, use that" behaviour.

I agree on you that ideally that should have been the target behaviour from the start:

> - VIRTUAL_PORT should always be honored (least surprise principle)
> - the server shouldn't be marked down unless it is unreachable
> - a comment could be added when VIRTUAL_PORT is not exposed (troubleshooting hint)

... but we don't know how many setup we might break by changing the first point because they relied implicitly on the previous behaviour, and I don't want to do something in the likes of https://github.com/moby/libnetwork/issues/2607 that I criticised myself.

Regarding nginx the `server 127.0.0.1 down` entry, are we safe including it by default or should we try to include It only when required (ie when the proxy and proxied containers don't share any common networks) ?

I would also prefer not to loose the `# Cannot connect to network of this container` indication because that's a pretty common misconfiguration, especially when using Docker Compose.

I'll think a bit more about it and come back to you with a suggestion we can discuss, if that's okay with you.

buchdag:
And if completely misunderstood the way this PR works, please correct me.

pini-gh:
> ... but we don't know how many setup we might break by changing the first point because they relied implicitly on the previous behaviour, and I don't want to do something in the likes of [moby/libnetwork#2607](https://github.com/moby/libnetwork/issues/2607) that I criticised myself.

Well, I really can't see any use case for defining a `VIRTUAL_PORT` variable which would not be the wanted port. This behavior (always honor `VIRTUAL_PORT`) is a kind of safety check: when only one port is exposed, either this is the same port, or the configuration is wrong on one side.

pini-gh:
> Regarding nginx the `server 127.0.0.1 down` entry, are we safe including it by default or should we try to include It only when required (ie when the proxy and proxied containers don't share any common networks) ?

I've been using this feature during several months with no problem so far. Do you think about a use case where it wouldn't be safe?

pini-gh:
> I would also prefer not to loose the `# Cannot connect to network of this container` indication because that's a pretty common misconfiguration, especially when using Docker Compose.

Agreed. How about printing the network's name as well then?
```
# Cannot connect to network '{{ $containerNetwork.Name }}' of this container
```

buchdag:
@pini-gh I'll come back to this PR by the end of the week

nginx.tmpl
```diff
@@ -180,29 +181,32 @@ server {
 upstream {{ $upstream_name }} {
 
 {{ range $container := $containers }}
-	{{ $addrLen := len $container.Addresses }}
-
+	{{ $debug := (eq (coalesce $container.Env.DEBUG $debug_all "false") "true") }}
+	{{/* If only 1 port exposed, use that as a default, else 80 */}}
+        {{ $defaultPort := (when (eq (len $container.Addresses) 1) (first $container.Addresses) (dict "Port" "80")).Port }}
```
buchdag, line 186: Could you fix the indentation on this line ? 😃

buchdag:
Looks good to me in this state and after clarification on `VIRTUAL_PORT`, but ideally I'd like tests to be adjusted to verify that behaviour.

The debug feature should also be tested, on the model of [location per vhost test](https://github.com/nginx-proxy/nginx-proxy/blob/aa9386129d7caee7165395e3c816171114416c0f/test/test_custom/test_location-per-vhost.py) :

```python
def test_custom_block_is_present_in_nginx_generated_conf(docker_compose, nginxproxy):
    assert b"include /etc/nginx/vhost.d/web1.nginx-proxy.local_location;" in nginxproxy.get_conf()
```

Please tell me if you need help with this.

pini-gh:
> Please tell me if you need help with this.

Yes, please bear with me. I'm new to this ci/cd thing. Is there a way to run these tests on my box, so I could proceed by try / error iterations?

pini-gh:
@buchdag I've added a test to check that the server is unreachable when VIRTUAL_PORT is defined different from the single exposed port. Is that what you have in mind?

buchdag:
> Yes, please bear with me. I'm new to this ci/cd thing. Is there a way to run these tests on my box, so I could proceed by try / error iterations?

Yes, instructions for running tests localy are in [`/tests`'s README ](https://github.com/nginx-proxy/nginx-proxy/tree/main/test/README.md), but since you added a test I guess your already found out.

In addition I'd suggest using a Python virtual environment (with something like `virtualenvwrapper`) to avoid clogging your local Python with unwanted packages.

buchdag:
>  I've added a test to check that the server is unreachable when VIRTUAL_PORT is defined different from the single exposed port.

Thanks, I think we're good with the `VIRTUAL_PORT` tests:

- [x] when a single port is exposed, this is the default port
- [x] when multiple port are exposed, the default port is `80`
- [x] `VIRTUAL_PORT` works
- [x] `VIRTUAL_PORT` is always honnored even wen it's different from the default port

We need another test for the `DEBUG` feature and we're good to go.

Something like this for instance

`/test/test_debug.yml`
```yaml
web:
  image: web
  expose:
    - "80"
    - "81"
  environment:
    WEB_PORTS: "80 81"
    VIRTUAL_HOST: "web.nginx-proxy.tld"
    VIRTUAL_PORT: "82"
    DEBUG: "true"

sut:
  image: nginxproxy/nginx-proxy:test
  volumes:
    - ../lib/ssl/dhparam.pem:/etc/nginx/dhparam/dhparam.pem:ro
```

`/test/test_debug.py`
```python
import pytest

def test_debug_info_is_present_in_nginx_generated_conf(docker_compose, nginxproxy):
    assert b"# Exposed ports: [{80 tcp} {81 tcp}]" in nginxproxy.get_conf()
    assert b"# Default virtual port: 80" in nginxproxy.get_conf()
    assert b"# VIRTUAL_PORT: 82" in nginxproxy.get_conf()
```

We want to test that `DEBUG=true` actually adds the expected comments in the generated nginx conf file.

pini-gh:
I've added two tests:
* one with DEBUG set at the proxy level
* the other with DEBUG set at the webserver level

buchdag:
LGTM, thanks ! 👍

buchdag:
@pini-gh regarding the default inclusion of the `server 127.0.0.1 down` entry:

> Do you think about a use case where it wouldn't be safe?

Well I found one: #1495

The fix proposed by @wlonkly in #1496 works and pass tests, but as stated in #1495 it would have the side effect of breaking load balancing behind the proxy. I know this isn't an advertised feature but looks like people are using it anyway (https://github.com/nginx-proxy/nginx-proxy/discussions/1654).

Do you think there would be an easy-ish way to output the `server 127.0.0.1 down` entry only when required ?

pini-gh:
It's not simple. Because the variables' scope is too limited with this template language. We need to be able to build a list from these nested loops:
```
        {{ range $knownNetwork := $CurrentContainer.Networks }}
                {{ range $containerNetwork := $container.Networks }}
                        {{ if (and (ne $containerNetwork.Name "ingress") (or (eq $knownNetwork.Name $containerNetwork.Name) (eq $knownNetwork.Name "host"))) }}                                                   
```
Then it's easy. But we need this list first, and I have no brilliant idea right now :/

buchdag:
I thought about using global variables (which is possible in the go templating language), but I could not find specifically if setting a global variables inside a `template` execution actions is possible or not.

pini-gh:
The only solution I can come up with is something along these lines:
* dereference the `upstream` template and hardcode it inside the nested loop
* use a `$server_found` variable set to `"false"` before the loops
* set this variable to `"true"` when a `server` line is issued
* print the fallback `server 127.0.0.1 down;` line only when `$server_found` equals `"false"`

PR #1610, opened 2021-05-02
Fix tests for virtual path feature

AlexanderLieret:
This fixes the failed tests for the WIP #1607.

test_custom/test_location-per-vpath:
This test was written before the pytest version bump and needed to be updated too.

test_wildcard_host:
The reason of this commit is not obvious. This might have been a quick fix for a problem solved in upstream.


buchdag:
Thanks for pinpointing the issue with the two tests, back to #1607

PR #1612, opened 2021-05-04
Automatic Dockerfile Image Updater

ducarrozb:
`nginx-proxy` changed recently. This pull request ensures you're using the latest version of the image and changes `nginx-proxy` to the latest tag: `latest`

New base image: `nginx-proxy:latest`


buchdag:
@ducarrozb not providing a tag after an image name make Docker defaults to `latest` anyway.

PR #1644, opened 2021-05-30
Update SSL policies, SSL Client certificates, Quic/HTTP3

patrickdk77:
Adds newer SSL policies
Adds SSL Client certificate support for verification and passthough
Adds quic/http3 configuration support
Makes http2 optional (packet loss makes http2 unusable)
Fix leftover port 80 definition with external_http_port definition


buchdag:
Hi, thanks for the PR. 👍 

> Adds newer SSL policies

Could you add the new SSL policies to the docs ?

> Makes http2 optional (packet loss makes http2 unusable)

Could you elaborate on this ? I'm seeing more and more people having issues with http2 without really understanding why.

Also this should be documented and, if possible, tested.

> Adds SSL Client certificate support for verification and passthough
> Adds quic/http3 configuration support

Same thing, those new features need doc update and tests, if it is actually possible to test them with our current CI.

patrickdk77:
Good writeup on http2 issues:
https://www.twilio.com/blog/2017/10/http2-issues.html
cloudflare also has a report, and why they developed quic/http3 so fast after http2 came out.

I personally had the issue from clients in china, having high rates of packet loss causing issues when I was using aws alb's and had to disable http2 on them.

I'm not sure if quic/http3 can be tested, as it's not in the offical nginx build yet.

I'll see what I can do about the rest.

buchdag:
> I'm not sure if quic/http3 can be tested, as it's not in the offical nginx build yet.

The CI rely on the [`requests` package](https://docs.python-requests.org/en/master/) which can only do HTTP/1.1, so I guess that answers that : we can't test HTTP/2 or HTTP/3 for now, don't bother.

[`httpx`](https://github.com/encode/httpx/) should be out of beta sometime in 2021, has a `requests` compatible api and supports HTTP/2. I'll take another look at it when it's released.

Things missing from the current PR:

- [ ] documentation update for the new SSL policies
- [ ] documentation on how to disable HTTP/2 (maybe with a few words or links about potential issues with it ?)
- [ ] documentation on HTTP/3 and how to enable it
- [ ] documentation on the client certificates feature
- [ ] tests for the client certificates feature

Also, should HTTP/2 and HTTP/3 be mutually exclusive ?

patrickdk77:
No you can run http/2 and http/3 at the same time.
http/2 run over https 443/tcp as normal, it mostly just multiplex the requests into a single tcp stream (like imap)

http/3 is the same deal as http/2, but runs over 443/udp, so it removes the issue with packet loss causing tcp to slow the connection down. There is an extra header on the http/1 http/2 response that tells the client if your http/3 enabled and what versions you support, so the client can switch over to it.

buchdag:
> http/3 is the same deal as http/2, but runs over 443/udp

That would require an additional port published on the proxy container like `--publish  443:443/udp`, right ?

patrickdk77:
Ok, added documentation and I see nginx preview http3 release is using the http3 tag instead of quic tag to enable it, so I set that as the default.
I don't know python to write a test script for client certs.

buchdag:
> I don't know python to write a test script for client certs.

No problem, I'll take a look at how we could/should write a test for this and come back to you when I'll have something.

In the meantime maybe we can merge this PR without the SSL client certificate feature and doc ?

patrickdk77:
Pulled the client certificate code and docs

buchdag:
@patrickdk77 with your approval I'm going to rebase the PR so when have a clean history prior to merge.

Fell free to open another PR for client certificate verification.

buchdag:
Ok I've given this a bit a more thought, I'm afraid that I'm not sure now is the time to start merging support for HTTP/3 if it is still so experimental that its isn't available in **nginx** yet. The aim of this image is to be based on and used with upstream nginx, not with experimental third party nginx images.

rbrueckner82:
Hey everyone,

is it now possible to enable http3 Support with the "ENABLE_HTTP3=quic" tag at the enviroment section?

thanks for the reply

buchdag:
@patrickdk77 discussion about QUIC and HTTP/3 support : #2271

buchdag:
Closed by #2274 + #2278

PR #1659, opened 2021-06-09
Improve logging

exussum12:
Currently its almost impossible to know which host actually handled the request, this extra variable logs the upstream server too

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -108,7 +108,8 @@ gzip_types text/plain text/css application/javascript application/json applicati
 
 log_format vhost '$host $remote_addr - $remote_user [$time_local] '
                  '"$request" $status $body_bytes_sent '
-                 '"$http_referer" "$http_user_agent"';
+                 '"$http_referer" "$http_user_agent" '
+		 '"$upstream_addr"';
 
 access_log off;
 
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -109,7 +109,7 @@ gzip_types text/plain text/css application/javascript application/json applicati
 log_format vhost '$host $remote_addr - $remote_user [$time_local] '
                  '"$request" $status $body_bytes_sent '
                  '"$http_referer" "$http_user_agent" '
-		 '"$upstream_addr"';
+                 '"$upstream_addr"';
 
 access_log off;
 


buchdag:
Hi ! Thanks for the PR !

Do you have a before PR / after PR example of the log ?

exussum12:
Sure.
Easiest to set up scenario is

Terminal 1
docker run  -p 80:80 -v /var/run/docker.sock:/tmp/docker.sock:ro jwilder/nginx-proxy

Terminal 2
docker run -e VIRTUAL_HOST=test.local nginx:latest

Terminal 3
docker run -e VIRTUAL_HOST=test.local nginx:latest


The proxy will load balance between the two
Terminal 4
curl 127.0.0.1 -H "Host: test.local"
curl 127.0.0.1 -H "Host: test.local" 

One request is served by each and the resulting logs are
```
nginx.1    | test.local 172.17.0.1 - - [11/Jun/2021:10:09:37 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.74.0"
nginx.1    | test.local 172.17.0.1 - - [11/Jun/2021:10:09:41 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.74.0"
```

After this change

```
nginx.1    | test.local 172.17.0.1 - - [11/Jun/2021:10:14:41 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.74.0" "172.17.0.4:80"
nginx.1    | test.local 172.17.0.1 - - [11/Jun/2021:10:14:42 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.74.0" "172.17.0.3:80"
```

The last param makes it easier work see where the request ended up.

I added to the end to prevent breaking any existing awk scripts which people have have written

buchdag:
Seems okay to me, I have no objection to merging this but I'm a bit concerned by people relying on load balancing with nginx-proxy. This was never intended as a feature, is untested and kinda just work "by accident".

exussum12:
Its not just load balancing it helps with, another use case is confirming you are going to the correct container in a default fall back. Eg nginx doesn't match as expected, and serves the default vhost instead. 

The load balancing case just makes it clear that the number changes.

PR #1667, opened 2021-06-15
Use `server 127.0.0.1 down` entry only when required

pini-gh:
See https://github.com/nginx-proxy/nginx-proxy/pull/1609#issuecomment-858150285.

Additionally, fix a warning due to a missing backslash escaping from previous merge request.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -4,28 +4,6 @@
 {{ $external_https_port := coalesce $.Env.HTTPS_PORT "443" }}
 {{ $debug_all := $.Env.DEBUG }}
 
-{{ define "upstream" }}
-	{{ if .Address }}
-		{{/* If we got the containers from swarm and this container's port is published to host, use host IP:PORT */}}
-		{{ if and .Container.Node.ID .Address.HostPort }}
-	# {{ .Container.Node.Name }}/{{ .Container.Name }}
-			server {{ .Container.Node.Address.IP }}:{{ .Address.HostPort }};
-		{{/* If there is no swarm node or the port is not published on host, use container's IP:PORT */}}
-		{{ else if .Network }}
-	# {{ .Container.Name }}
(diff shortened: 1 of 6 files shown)


buchdag:
Do you think the test cases could be expanded to include the following:

- multiple (three ?) containers with the same `VIRTUAL_HOST`, one on an unreachable network: assert that we still don't get `server 127.0.0.1 down` in this setup
- single container with multiple unreachable networks (I don't know if it's possible): assert that we only get `server 127.0.0.1 down` once

Or maybe it's overkill ?

Excellent work anyway, thank you.

pini-gh:
>     * multiple (three ?) containers with the same `VIRTUAL_HOST`, one on an unreachable network: assert that we still don't get `server 127.0.0.1 down` in this setup
This implies that they use the same VIRTUAL_PORT, doesn't it?

>     * single container with multiple unreachable networks (I don't know if it's possible): assert that we only get `server 127.0.0.1 down` once
Will give it a try.

buchdag:
> This implies that they use the same VIRTUAL_PORT, doesn't it?

Not necessarily. The idea was to cover the (undocumented, admittedly) use case of people doing load balancing behind nginx-proxy as described in #1654, but failing this test won't mean that this use case would stop working. Both this test and the other one are more unit testing than functional testing, hence the overkill question.

I have no problem merging this now and adding further tests later of you agree that they'd be nice to have.

pini-gh:
> single container with multiple unreachable networks (I don't know if it's possible): assert that we only get `server 127.0.0.1 down` once

@buchdag I don't know how to do it using docker-compose v1 syntax. And and my box `test_composev2` fails because `test_nginx-proxy_1` and `test_web_1` are on the auto allocated `test_default` network while the `nginx-proxy-tester` instance is on the `bridge` network. Then the tester can't talk to the usecase instances. Any idea?

PR #1699, opened 2021-07-15
Virtual path additions

buchdag:
@AlexanderLieret can I let you introduce this PR ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,20 @@ The full request URI will be forwarded to the serving container in the `X-Forwar
 
 **NOTE**: Your application needs to be able to generate links starting with `VIRTUAL_PATH`. This can be achieved by it being natively on this path or havin an option to prepend this path. The application does not need to expect this path in the request.
 
+#### VIRTUAL_DEST
+
+This environment variable can be used to rewrite the `VIRTUAL_PATH` part of the requested URL to proxied application. The default value is empty (off).
+Make sure that your settings won't result in the slash missing or being doubled. Both these versions can cause troubles.
+
+If the application runs natively on this sub-path or has a setting to do so, `VIRTUAL_DEST` should not be set or empty.
+If the requests are expected to not contain a sub-path and the generated links contain the sub-path, `VIRTUAL_DEST=/` should be used.
+
+```console
+$ docker run -d -e VIRTUAL_HOST=example.tld -e VIRTUAL_PATH=/app1/ -e VIRTUAL_DEST=/ --name app1 app
+```
+
+In this example, the incoming request `http://example.tld/app1/foo` will be proxied as `http://app1/foo` instead of `http://app1/app1/foo`.
+
 #### DEFAULT_ROOT
 
 This environment variable of the nginx proxy container can be used to customize the return error page if no matching path is found. Furthermore it is possible to use anything which is compatible with the `return` statement of nginx.
--- a/README.md
+++ b/README.md
@@ -109,6 +109,16 @@ You can have multiple containers proxied by the same `VIRTUAL_HOST` by adding a 
 
 The full request URI will be forwarded to the serving container in the `X-Forwarded-Path` header.
 
+**NOTE**: Your application needs to be able to generate links starting with `VIRTUAL_PATH`. This can be achieved by it being natively on this path or havin an option to prepend this path. The application does not need to expect this path in the request.
+
+#### DEFAULT_ROOT
+
+This environment variable of the nginx proxy container can be used to customize the return error page if no matching path is found. Furthermore it is possible to use anything which is compatible with the `return` statement of nginx.
+
+For example `DEFAUL_ROOT=418` will return a 418 error page instead of the normal 404 one.
+Another example is `DEFAULT_ROOT="301 https://github.com/nginx-proxy/nginx-proxy/blob/main/README.md"` which would redirect an invalid request to this documentation.
+
+
 ### Multiple Networks
 
 With the addition of [overlay networking](https://docs.docker.com/engine/userguide/networking/get-started-overlay/) in Docker 1.9, your `nginx-proxy` container may need to connect to backend containers on multiple networks. By default, if you don't pass the `--net` flag when your `nginx-proxy` container is created, it will only be attached to the default `bridge` network. This means that it will not be able to connect to containers on networks other than `bridge`.
--- a/README.md
+++ b/README.md
@@ -140,6 +140,11 @@ This environment variable of the nginx proxy container can be used to customize 
 
 For example `DEFAUL_ROOT=418` will return a 418 error page instead of the normal 404 one.
 Another example is `DEFAULT_ROOT="301 https://github.com/nginx-proxy/nginx-proxy/blob/main/README.md"` which would redirect an invalid request to this documentation.
+Nginx variables such as $scheme, $host, and $request_uri can be used. However, care must be taken to make sure the $ signs are escaped properly.
+If you want to use `301 $scheme://$host/myapp1$request_uri` you should use:
+
+* Bash: `DEFAULT_ROOT='301 $scheme://$host/myapp1$request_uri'`
+* Docker Compose yaml: `- DEFAULT_ROOT: 301 $$scheme://$$host/myapp1$$request_uri`
 
 
 ### Multiple Networks
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ It is also possible to specify multiple paths with regex locations like `VIRTUAL
 
 The full request URI will be forwarded to the serving container in the `X-Forwarded-Path` header.
 
-**NOTE**: Your application needs to be able to generate links starting with `VIRTUAL_PATH`. This can be achieved by it being natively on this path or havin an option to prepend this path. The application does not need to expect this path in the request.
+**NOTE**: Your application needs to be able to generate links starting with `VIRTUAL_PATH`. This can be achieved by it being natively on this path or having an option to prepend this path. The application does not need to expect this path in the request.
 
 #### VIRTUAL_DEST
 
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ You can also use wildcards at the beginning and the end of host name, like `*.ba
 ### Path-based Routing
 
 You can have multiple containers proxied by the same `VIRTUAL_HOST` by adding a `VIRTUAL_PATH` environment variable containing the absolute path to where the container should be mounted. For example with `VIRTUAL_HOST=foo.example.com` and `VIRTUAL_PATH=/api/v2/service`, then requests to http://foo.example.com/api/v2/service will be routed to the container. If you wish to have a container serve the root while other containers serve other paths, give the root container a `VIRTUAL_PATH` of `/`.  Unmatched paths will be served by the container at `/` or will return the default nginx error page if no container has been assigned `/`.
+It is also possible to specify multiple paths with regex locations like `VIRTUAL_PATH=~^/(app1|alternative1)/`. For further details see the nginx documentation on location blocks. This is not compatible with `VIRTUAL_DEST`.
 
 The full request URI will be forwarded to the serving container in the `X-Forwarded-Path` header.
 
@@ -124,6 +125,14 @@ $ docker run -d -e VIRTUAL_HOST=example.tld -e VIRTUAL_PATH=/app1/ -e VIRTUAL_DE
 ```
 
 In this example, the incoming request `http://example.tld/app1/foo` will be proxied as `http://app1/foo` instead of `http://app1/app1/foo`.
+
+#### Per-VIRTUAL_PATH location configuration
+
+The same options as from [Per-VIRTUAL_HOST location configuration](#Per-VIRTUAL_HOST-location-configuration) are available on a `VIRTUAL_PATH` basis.
+The only difference is that the filename gets an additional block `HASH=$(echo -n $VIRTUAL_PATH | sha1sum | awk '{ print $1 }')`. This is the sha1-hash of the `VIRTUAL_PATH` (no newline). This is done filename sanitization purposes.
+The used filename is `${VIRTUAL_HOST}_${HASH}_location`
+
+The filename of the previous example would be `example.tld_8610f6c344b4096614eab6e09d58885349f42faf_location`.
 
 #### DEFAULT_ROOT
 
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -70,6 +70,8 @@ location {{ .Path }} {
 
 	{{ if (exists (printf "/etc/nginx/vhost.d/%s_location" .Host)) }}
 		include {{ printf "/etc/nginx/vhost.d/%s_location" .Host}};
+	{{ else if (exists (printf "/etc/nginx/vhost.d/%s_%s_location" .Host (sha1 .Path) )) }}
+		include {{ printf "/etc/nginx/vhost.d/%s_%s_location" .Host (sha1 .Path) }};
 	{{ else if (exists "/etc/nginx/vhost.d/default_location") }}
 		include /etc/nginx/vhost.d/default_location;
 	{{ end }}
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -3,6 +3,7 @@
 {{ $external_http_port := coalesce $.Env.HTTP_PORT "80" }}
 {{ $external_https_port := coalesce $.Env.HTTPS_PORT "443" }}
 {{ $debug_all := $.Env.DEBUG }}
+{{ $default_root_response := coalesce $.Env.DEFAULT_ROOT "404" }}
 
 {{ define "ssl_policy" }}
 	{{ if eq .ssl_policy "Mozilla-Modern" }}
@@ -386,6 +387,11 @@ server {
 			{{ $upstream := printf "%s-%s" $upstream_name $sum }}
 			{{ template "location" (dict "Path" $path "Proto" $proto "Upstream" $upstream "Host" $host "Vhostroot" $vhost_root) }}
 		{{ end }}
+		{{ if (not (contains $paths "/")) }}
+			location / {
+				return {{ $default_root_response }};
+			}
+		{{ end }}
 	{{ end }}
 }
 
@@ -422,6 +428,11 @@ server {
 			{{ $sum := sha1 $path }}
 			{{ $upstream := printf "%s-%s" $upstream_name $sum }}
 			{{ template "location" (dict "Path" $path "Proto" $proto "Upstream" $upstream "Host" $host "Vhostroot" $vhost_root) }}
+		{{ end }}
+		{{ if (not (contains $paths "/")) }}
+			location / {
+				return {{ $default_root_response }};
+			}
 		{{ end }}
 	{{ end }}
 }
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -68,10 +68,10 @@ location {{ .Path }} {
 		auth_basic_user_file	{{ (printf "/etc/nginx/htpasswd/%s" .Host) }};
 	{{ end }}
 
-	{{ if (exists (printf "/etc/nginx/vhost.d/%s_location" .Host)) }}
-		include {{ printf "/etc/nginx/vhost.d/%s_location" .Host}};
-	{{ else if (exists (printf "/etc/nginx/vhost.d/%s_%s_location" .Host (sha1 .Path) )) }}
+	{{ if (exists (printf "/etc/nginx/vhost.d/%s_%s_location" .Host (sha1 .Path) )) }}
 		include {{ printf "/etc/nginx/vhost.d/%s_%s_location" .Host (sha1 .Path) }};
+	{{ else if (exists (printf "/etc/nginx/vhost.d/%s_location" .Host)) }}
+		include {{ printf "/etc/nginx/vhost.d/%s_location" .Host}};
 	{{ else if (exists "/etc/nginx/vhost.d/default_location") }}
 		include /etc/nginx/vhost.d/default_location;
 	{{ end }}
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -60,7 +60,7 @@ location {{ .Path }} {
 	{{ else if eq .Proto "grpc" }}
 		grpc_pass {{ trim .Proto }}://{{ trim .Upstream }};
 	{{ else }}
-		proxy_pass {{ trim .Proto }}://{{ trim .Upstream }}/;
+		proxy_pass {{ trim .Proto }}://{{ trim .Upstream }}{{ trim .Dest }};
 	{{ end }}
 
 	{{ if (exists (printf "/etc/nginx/htpasswd/%s" .Host)) }}
@@ -380,12 +380,13 @@ server {
 	{{ end }}
 
 	{{ if eq $nPaths 0 }}
-		{{ template "location" (dict "Path" "/" "Proto" $proto "Upstream" $upstream_name "Host" $host "Vhostroot" $vhost_root) }}
+		{{ template "location" (dict "Path" "/" "Proto" $proto "Upstream" $upstream_name "Host" $host "Vhostroot" $vhost_root "Dest" "") }}
 	{{ else }}
 		{{ range $path, $container := $paths }}
 			{{ $sum := sha1 $path }}
 			{{ $upstream := printf "%s-%s" $upstream_name $sum }}
-			{{ template "location" (dict "Path" $path "Proto" $proto "Upstream" $upstream "Host" $host "Vhostroot" $vhost_root) }}
+			{{ $dest := (or (first (groupByKeys $container "Env.VIRTUAL_DEST")) "") }}
+			{{ template "location" (dict "Path" $path "Proto" $proto "Upstream" $upstream "Host" $host "Vhostroot" $vhost_root "Dest" $dest) }}
 		{{ end }}
 		{{ if (not (contains $paths "/")) }}
 			location / {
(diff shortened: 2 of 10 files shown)


AlexanderLieret:
This PR add three features to the path-based routing.

### Customizable default response
If there is no container proxied for the path `/`, a user customizable response will be sent. The original code relied on the nginx behavior (return a 404 error page) if there is no matching location block. The new environment variable allows a custom return statement, the default still is `404`. It is possible to return any other valid response code, i.e. `418` or `301 /application1`.

### Customizable path stripping
The old behavior is to always strip the virtual path prefix. This is not compatible with all container solutions and regex matches. The environment variable `VIRTUAL_DEST` specifies what the replacement is and whether to replace at all. More complex replacements need the next feature.

### Custom location block per virtual path
It is possible to inject custom config files for the location blocks. For example this allows to apply regex based path rewriting or adding custom config as already possible.

warren-ru:
In my case, VIRTUAL_PATH feature doesn't work.
Even if I declare the one container with VIRTUAL_PATH=/, such container isn't accessible.
When I remove the VIRTUAL_PATH parameter leaving only VIRTUAL_HOST, everything works properly.

buchdag:
@warren-ru could you share your setup (compose files or command line used for both nginx-proxy and the proxied apps) ?

warren-ru:
@buchdag sure.
I use a Django app as the service behind the proxy.
I try to use new feature, but can't get access to my service.
Just I comment out the VIRTUAL_PATH parameter in .env file, service becomes reachable.

docker-compose.yml
```yml
version: '3'

services:
  proxy:
    image: nginxproxy/nginx-proxy:dev-alpine
    restart: always
    ports:
      - 80:80
      - 443:443
    labels:
      com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_proxy: "true"
    volumes:
      - certs:/etc/nginx/certs
      - vhost.d:/etc/nginx/vhost.d
      - html:/usr/share/nginx/html
      - dhparam:/etc/nginx/dhparam
    networks:
      - proxy-tier
  letsencrypt-companion:
    image: nginxproxy/acme-companion
    restart: always
    volumes:
      - certs:/etc/nginx/certs
      - vhost.d:/etc/nginx/vhost.d
      - acme:/etc/acme.sh
      - html:/usr/share/nginx/html
    networks:
      - proxy-tier
    depends_on:
      - proxy
  cron:
    build: https://github.com/############
    restart: always
    env_file:
      - ./passportdb.env
  passport:
    image: ghcr.io/##############
    command: gunicorn passport.wsgi:application --bind 0.0.0.0:8001
    restart: always
    volumes:
      - passport_static_volume:/home/app/web/static
    expose:
      - 8001
    env_file:
      - ./passportdb.env
    depends_on:
      - db_passport
    networks:
      - proxy-tier
      - default
  db_passport:
    image: postgres:12.0-alpine
    restart: always
    volumes:
      - postgres_passport_data:/var/lib/postgresql/data/
    env_file:
      - ./passportdb.env
volumes:
  postgres_passport_data:
  passport_static_volume:
  vhost.d:
  acme:
  certs:
  html:
  dhparam:
  
networks:
  proxy-tier:
```

env file mentioned in docker-compose
```env
PGDATA=/var/lib/postgresql/data/pgdata
C_FORCE_ROOT=true

#Postgres settings
PASSPORT_DB=passport_db
POSTGRES_DB=passport_db
POSTGRES_USER=passportdbuser
POSTGRES_PASSWORD=######################
POSTGRES_HOST=db_passport

#Django settings
PASSPORT_SECRET_KEY=########################
PASSPORT_SERVICE_TOKEN=###################

#nginx_proxy settings
VIRTUAL_HOST=some.server.com
VIRTUAL_PORT=8001
#VIRTUAL_PATH=/

#Lets Encrypt settings
DEFAULT_EMAIL=email@server.com
LETSENCRYPT_HOST=some.server.com
```

OliverRhyme:
The adding VITUAL_PATH breaks ssl via the acme_companion image. The ssl generated resolves to the default ssl.

README.md
```diff
@@ -106,9 +106,42 @@ You can also use wildcards at the beginning and the end of host name, like `*.ba
...
+It is also possible to specify multiple paths with regex locations like `VIRTUAL_PATH=~^/(app1|alternative1)/`. For further details see the nginx documentation on location blocks. This is not compatible with `VIRTUAL_DEST`.
 
 The full request URI will be forwarded to the serving container in the `X-Forwarded-Path` header.
 
+**NOTE**: Your application needs to be able to generate links starting with `VIRTUAL_PATH`. This can be achieved by it being natively on this path or havin an option to prepend this path. The application does not need to expect this path in the request.
```
junderw, line 113:
```suggestion
**NOTE**: Your application needs to be able to generate links starting with `VIRTUAL_PATH`. This can be achieved by it being natively on this path or having an option to prepend this path. The application does not need to expect this path in the request.
```

junderw:
2 questions:

1. It seems the above issue is caused by acme-companion, I also have a similar setup and would like this feature to work with acme-companion. Is this a problem with the issue reporter's config or is it confirmed to be an issue with this PR?
2. Can I use per host variables in DEFAULT_ROOT? I would like to do `301 https://$host/app1$path` and forward root paths to app1 path.

AlexanderLieret:
@junderw 

>     1. It seems the above issue is caused by acme-companion, I also have a similar setup and would like this feature to work with acme-companion. Is this a problem with the issue reporter's config or is it confirmed to be an issue with this PR?

I did reproduce this error using a simplified docker-compose file. The one posted by @warren-ru does a few problematic things, i.e. having a volume for the acme.sh or specifying the `VIRTUAL_PATH=/` for both the app and its db. 

The generated nginx configuration from both the dev build and this PR did not show any unwanted differences. The problem is that the acme-companion does create a self-signed certificate with `CN=letsencrypt-nginx-proxy-companion`. Maybe it tries to get this signed? Needs more troubleshooting in this direction.

> 2. Can I use per host variables in DEFAULT_ROOT? I would like to do 301 https://$host/app1$path and forward root paths to app1 path.

Yes this works. The second example could be modified to mention this.

junderw:
> Yes this works. The second example could be modified to mention this.

Awesome! Yeah, I think that should be shown that the vars will work.

> Needs more troubleshooting in this direction.

Hmm, let me know if there's anything I can do to help. I am not too familiar with how the whole thing works, but I have time and a desire to get this feature working.

junderw:
Also, is there any rough ballpark as to which version / when this will be in a numbered release?

AlexanderLieret:
> > Yes this works. The second example could be modified to mention this.
> 
> Awesome! Yeah, I think that should be shown that the vars will work.

@junderw `301 https://$host/app1$request_uri` should be the correct way. nginx does not have a variable $path but $request_uri. Without testing the only problem I see is that docker-compose can interfere with variable expansion. This might require $$ instead of a single $.

nginx.tmpl
```diff
@@ -379,12 +382,18 @@ server {
...
 
 	{{ if eq $nPaths 0 }}
-		{{ template "location" (dict "Path" "/" "Proto" $proto "Upstream" $upstream_name "Host" $host "Vhostroot" $vhost_root) }}
+		{{ template "location" (dict "Path" "/" "Proto" $proto "Upstream" $upstream_name "Host" $host "Vhostroot" $vhost_root "Dest" "") }}
 	{{ else }}
 		{{ range $path, $container := $paths }}
 			{{ $sum := sha1 $path }}
-			{{ $upstream := printf "%s-%s" $host $sum }}
-			{{ template "location" (dict "Path" $path "Proto" $proto "Upstream" $upstream "Host" $host "Vhostroot" $vhost_root) }}
+			{{ $upstream := printf "%s-%s" $upstream_name $sum }}
```
AlexanderLieret, line 389: This line change fixes the error found by @warren-ru. The error affects https entirely.

buchdag: I've commited the `$host` -> `$upstream_name` fix to the `dev` branch.

buchdag:
Rebased `dev` branch to `main` and this PR to `dev` branch, HSTS tests might be failing.

warren-ru:
> 
>  
> @warren-ru could you test again with the new `dev` image ?

As @AlexanderLieret mentioned I split my env-file into two separate env-files, for the app service and for the DB service, so that the DB service doesn't expose nginx-proxy environments.

My Django app uses "/passport/' location.
When I declare VIRTUAL_PATH as "/" everything works well. But when I use "VIRTUAL_PATH=/passport/" I see "The requested resource was not found on this server" nginx error.

junderw:
I was able to verify VIRTUAL_DEST, VIRTUAL_PATH, and DEFAULT_ROOT behavior.

As you said I needed to double $ the string. Another thing that tripped me up was that docker-compose.yml does not need " or \ for spaces, and the " and \ were included in the generated conf file... so I used `- DEFAULT_ROOT= 301 https://$$host/app1$$request_uri` in my docker-compose.yml

acme companion worked fine.

My application went a tad wonky, but that's on my side.

tACK d8db897

junderw:
Here's my docker-compose btw. The nginx-proxy image was built locally from this branch.

```yaml
version: "3"

services:

  teslamate:
    image: teslamate/teslamate:latest
    container_name: teslamate
    hostname: teslamate
    restart: unless-stopped
    expose:
      - "4000"
    environment:
      - DATABASE_USER=teslamate
      - DATABASE_PASS=${INTERNAL_PASSWORD}
      - DATABASE_NAME=teslamate
      - DATABASE_HOST=database
      - DATABASE_PORT=5432
      - DISABLE_MQTT=true
      - LETSENCRYPT_EMAIL=${SSL_EXPIRE_EMAIL}
      - VIRTUAL_HOST=${TESLAMATE_MAIN_DOMAIN}
      - VIRTUAL_PATH=/
      - VIRTUAL_PORT=4000
      - LETSENCRYPT_HOST=${TESLAMATE_MAIN_DOMAIN}
      - TZ=${TESLAMATE_TZ}
    volumes:
      - ${CONFIG_BASE}/import:/opt/app/import
    cap_drop:
      - all
    network_mode: "bridge"
    links:
      - "database"

  grafana:
    image: teslamate/grafana:latest
    container_name: grafana
    hostname: grafana
    restart: unless-stopped
    expose:
      - "3000"
    environment:
      - DATABASE_USER=teslamate
      - DATABASE_PASS=${INTERNAL_PASSWORD}
      - DATABASE_NAME=teslamate
      - DATABASE_HOST=database
      - LETSENCRYPT_EMAIL=${SSL_EXPIRE_EMAIL}
      - VIRTUAL_HOST=${TESLAMATE_MAIN_DOMAIN}
      - VIRTUAL_PATH=/grafana/
      - VIRTUAL_DEST=/
      - VIRTUAL_PORT=3000
      - LETSENCRYPT_HOST=${TESLAMATE_MAIN_DOMAIN}
      - GF_SERVER_ROOT_URL=https://${TESLAMATE_MAIN_DOMAIN}/grafana
      - GF_SERVER_SERVE_FROM_SUB_PATH=true
    volumes:
      - ${CONFIG_BASE}/teslamate-grafana-data:/var/lib/grafana
    network_mode: "bridge"
    links:
      - "database"

  database:
    image: postgres:13
    container_name: database
    hostname: database
    expose:
      - "5432"
    restart: unless-stopped
    environment:
      - POSTGRES_USER=teslamate
      - POSTGRES_PASSWORD=${INTERNAL_PASSWORD}
      - POSTGRES_DB=teslamate
    volumes:
      - ${CONFIG_BASE}/teslamate-db:/var/lib/postgresql/data
    network_mode: "bridge"

  nginx-proxy:
    image: nginx-proxy-path
    restart: unless-stopped
    container_name: nginx-proxy
    ports:
      - "80:80"
      - "443:443"
    volumes:
      - ${CONFIG_BASE}/nginx/htpasswd:/etc/nginx/htpasswd
      - ${CONFIG_BASE}/nginx/conf:/etc/nginx/conf.d
      - ${CONFIG_BASE}/nginx/vhost:/etc/nginx/vhost.d
      - ${CONFIG_BASE}/nginx/html:/usr/share/nginx/html
      - ${CONFIG_BASE}/nginx/dhparam:/etc/nginx/dhparam
      - ${CONFIG_BASE}/nginx/certs:/etc/nginx/certs:ro
    network_mode: "bridge"

  acme-companion:
    image: nginxproxy/acme-companion:latest
    restart: unless-stopped
    container_name: nginx-proxy-acme
    volumes_from:
      - nginx-proxy
    volumes:
      - ${CONFIG_BASE}/nginx/certs:/etc/nginx/certs:rw
      - ${CONFIG_BASE}/nginx/acme:/etc/acme.sh
    network_mode: "bridge"
```

buchdag:
> My Django app uses `/passport/` location. When I declare `VIRTUAL_PATH` as "/" everything works well. But when I use `VIRTUAL_PATH=/passport/` I see "The requested resource was not found on this server" nginx error.

When you access your django application without using `VIRTUAL_PATH` at all, what URL do you use ?

`www.yourdomain.tld` ?

`www.yourdomain.tld/passport` ?

warren-ru:
> > My Django app uses `/passport/` location. When I declare `VIRTUAL_PATH` as "/" everything works well. But when I use `VIRTUAL_PATH=/passport/` I see "The requested resource was not found on this server" nginx error.
> 
> When you access your django application without using `VIRTUAL_PATH` at all, what URL do you use ?
> 
> `www.yourdomain.tld` ?
> 
> `www.yourdomain.tld/passport` ?

www.yourdomain.tld/passport/

junderw:
I think warren's issue would be fixed by this PR's VIRTUAL_DEST=/passport/

buchdag:
@junderw unless I completely missed something, at this point the default behaviour of `VIRTUAL_PATH`, even without this PR, is the following:

with `VIRTUAL_HOST=www.yourdomain.tld` and `VIRTUAL_PATH=/passport` on container `foobarbuz`, incoming requests to `http://www.yourdomain.tld/passport` are proxied to `http://foobarbuz/passport`

@AlexanderLieret can you confirm / infirm ? I think we went in that direction in #1607 but maybe I got that wrong.

junderw:
looking at the diff, the place where Dest goes is newly added, whereas if it were automatically forwarding to the same path you'd see something replaced, not newly added.

looks like dev VIRTUAL_PATH always forwards to http://foobarbuz root.

junderw:
I tried `dev` and the file it generates does not include any paths at the end of the proxy_pass to upstream.

Whereas this PR + usage of VIRTUAL_DEST will include the path of VIRTUAL_DEST.

Neither ones included anything path related (besides the sha1 of VIRTUAL_PATH) in the upstream definitions.

nginx.tmpl
```diff
@@ -379,12 +382,18 @@ server {
...
 
 	{{ if eq $nPaths 0 }}
-		{{ template "location" (dict "Path" "/" "Proto" $proto "Upstream" $upstream_name "Host" $host "Vhostroot" $vhost_root) }}
+		{{ template "location" (dict "Path" "/" "Proto" $proto "Upstream" $upstream_name "Host" $host "Vhostroot" $vhost_root "Dest" "") }}
 	{{ else }}
 		{{ range $path, $container := $paths }}
 			{{ $sum := sha1 $path }}
 			{{ $upstream := printf "%s-%s" $upstream_name $sum }}
-			{{ template "location" (dict "Path" $path "Proto" $proto "Upstream" $upstream "Host" $host "Vhostroot" $vhost_root) }}
+			{{ $dest := (or (first (groupByKeys $container "Env.VIRTUAL_DEST")) "") }}
```
junderw, line 390:
```suggestion
			{{ $dest := (or (first (groupByKeys $container "Env.VIRTUAL_DEST")) $path) }}
```

I think this will create the behavior you expected @buchdag

AlexanderLieret:
> @junderw unless I completely missed something, at this point the default behaviour of `VIRTUAL_PATH`, even without this PR, is the following:
> 
> with `VIRTUAL_HOST=www.yourdomain.tld` and `VIRTUAL_PATH=/passport` on container `foobarbuz`, incoming requests to `http://www.yourdomain.tld/passport` are proxied to `http://foobarbuz/passport`
> 
> @AlexanderLieret can you confirm / infirm ? I think we went in that direction in #1607 but maybe I got that wrong.

The behaviour of the dev branch is to strip the path. This is not compatible with more advanced `VIRTUAL_PATH` situations like regex matches.
In the example above the request `http://www.yourdomain.tld/passport/kitten.jpg` will be proxied to `http://foobarbuz/kitten.jpg`.

This PR adds the `VIRTUAL_DEST` variable to control the rewriting. With the default empty value the example will be proxied to `http://foobarbuz/passport/kitten.jpg`. This would fix @warren-ru issue. To restore the original behaviour add the enviroment variable `VIRTUAL_DEST=/`.

Changing the default value to `/` is possible if setting an empty value `VIRTUAL_DEST=` does take precedence.

junderw:
I find the current state of this PR to be the optimal method. (my two cents)

nginx.tmpl
AlexanderLieret: @junderw This change is unnecessary and can cause problems. Setting $dest to $path will rewrite the matched location path with $path. This is unnecessary and breaks if $path is an regex. An empty $dest will not do any rewriting.

junderw: I agree. Let's keep it as is.

warren-ru:
@AlexanderLieret yes. I supposed VIRTUAL_DEST to be optional. If no VIRTUAL_DEST is set proxy passes request to VIRTUAL_PATH location appending it to container address.

buchdag:
> I find the current state of this PR to be the optimal method. (my two cents)

Okay I think we all agree on that.

@warren-ru I built and pushed an image based on this PR to `nginxproxy/nginx-proxy:1699` so you can test it.

warren-ru:
> 
> 
> > I find the current state of this PR to be the optimal method. (my two cents)
> 
> Okay I think we all agree on that.
> 
> @warren-ru I built and pushed an image based on this PR to `nginxproxy/nginx-proxy:1699` so you can test it.
> 
> @OliverRhyme the issue you experienced should be fixed in both `nginxproxy/nginx-proxy:dev` and `nginxproxy/nginx-proxy:1699` is you want to try this feature again.

I've just tried out the 1699 build. Now everything is ok!

README.md
```diff
@@ -106,9 +106,42 @@ You can also use wildcards at the beginning and the end of host name, like `*.ba
...
+
+The filename of the previous example would be `example.tld_8610f6c344b4096614eab6e09d58885349f42faf_location`.
+
+#### DEFAULT_ROOT
+
+This environment variable of the nginx proxy container can be used to customize the return error page if no matching path is found. Furthermore it is possible to use anything which is compatible with the `return` statement of nginx.
+
+For example `DEFAUL_ROOT=418` will return a 418 error page instead of the normal 404 one.
+Another example is `DEFAULT_ROOT="301 https://github.com/nginx-proxy/nginx-proxy/blob/main/README.md"` which would redirect an invalid request to this documentation.
+
```
junderw, line 143:
```suggestion
Nginx variables such as $scheme, $host, and $request_uri can be used. However, care must be taken to make sure the $ signs are escaped properly (ie. `\` for bash commands and `$$` for docker-compose yaml files).
Yaml files will include `""` or `\`, so if you want to use `301 $scheme://$host/myapp1$request_uri` you should use:

* Bash: `DEFAULT_ROOT="301 \$scheme://\$host/myapp1\$request_uri"`
* Docker Compose yaml: `- DEFAULT_ROOT: 301 $$scheme://$$host/myapp1$$request_uri`

```

buchdag: For bash the other solution is to use single quotes; ie `--env DEFAULT_ROOT='301 $scheme://$host/myapp1$request_uri'`

buchdag: > Yaml files will include `""` or `\`

This might be confusing: docker compose file does not include anything specific, that's the other way around where bash needs quoting (either with single/double quotes or backslash) to remove the special meaning of certain characters, in this case `$` and `<space>`.

OliverRhyme:
> > I find the current state of this PR to be the optimal method. (my two cents)
> 
> Okay I think we all agree on that.
> 
> @warren-ru I built and pushed an image based on this PR to `nginxproxy/nginx-proxy:1699` so you can test it.
> 
> @OliverRhyme the issue you experienced should be fixed in both `nginxproxy/nginx-proxy:dev` and `nginxproxy/nginx-proxy:1699` if you want to try this feature again.

Does the docker image for the dev branch updated already?

buchdag:
> Does the docker image for the dev branch updated already?

Yep.

`dev` corresponds to #1607 rebased to the `main` branch
`1699` corresponds to the current state of this PR

warren-ru:
@buchdag I can't use VIRTUAL_PATH and configure locations through the "default_location" file in /etc/nginx/vhost.d/ at the same time.
Did I get it right?

OliverRhyme:
> > Does the docker image for the dev branch updated already?
> 
> Yep.
> 
> `dev` corresponds to #1607 rebased to the `main` branch
> `1699` corresponds to the current state of this PR

Thanks, can also confirm dev-alpine image working as expected

buchdag:
> I can't use `VIRTUAL_PATH` and configure locations through the `default_location` file in `/etc/nginx/vhost.d/` at the same time. Did I get it right?

Nope, if present `/etc/nginx/vhost.d/default_location` should still be included in every location generated wether through a `VIRTUAL_HOST` alone or through a `VIRTUAL_HOST` + `VIRTUAL_PATH` (that's what allows the acme-companion to work without modification).

warren-ru:
> 
> 
> > I can't use `VIRTUAL_PATH` and configure locations through the `default_location` file in `/etc/nginx/vhost.d/` at the same time. Did I get it right?
> 
> Nope, if present `/etc/nginx/vhost.d/default_location` should still be included in every location generated wether through a `VIRTUAL_HOST` alone or through a `VIRTUAL_HOST` + `VIRTUAL_PATH`.
> 
> Same thing for `/etc/nginx/vhost.d/default` at the `server` block level, that's what allows the acme-companion to work without modification.

It isn't in my case though. I have some default_location file mounted to /etc/nginx/vhost.d using a docker volume. When I add a VIRTAUL_PATH parameter it breaks not only locations from the default_location file, but a location set by the VIRTUAL_PATH parameter either.

junderw:
> It isn't in my case though. I have some default_location file mounted to /etc/nginx/vhost.d using a docker volume. When I add a VIRTAUL_PATH parameter it breaks not only locations from the default_location file, but a location set by the VIRTUAL_PATH parameter either.

Do you have any files like `/etc/nginx/vhost.d/${VIRTUAL_DOMAIN}_location` or `/etc/nginx/vhost.d/${VIRTUAL_DOMAIN}_xxxxx_location` with the same domain as the one using VIRTUAL_PATH?

Or is the only content of `/etc/nginx/vhost.d/` default_location?

warren-ru:
@junderw only content of /etc/nginx/vhost.d/ is the default_location file.

junderw:
What is the content of the default_location file like?

warren-ru:
@junderw I use a default_location file to host Django static files.
In my case it looks so:
```
    location /staticfiles/ {
        alias /usr/share/nginx/html/static/;
    }

    location /static/ {
        alias /usr/share/nginx/html/static/;
    }
```

junderw:
@warren-ru In your case you will have something like this then:

```
location /passport/ {

   ... ... ...

    location /staticfiles/ {
        alias /usr/share/nginx/html/static/;
    }

    location /static/ {
        alias /usr/share/nginx/html/static/;
    }
}
```

Nested locations are not relative, so you must make sure it is like:

```
location /passport/ {

   ... ... ...

    location /passport/staticfiles/ {
        alias /usr/share/nginx/html/static/;
    }

    location /passport/static/ {
        alias /usr/share/nginx/html/static/;
    }
}
```

junderw:
I think maybe you want to use `default` and not `default_location`

The former will be included once outside all the VIRTUAL_PATH locations, the former (with _location) will be included in EACH location of VIRTUAL_PATH.

The reason it works without VIRTUAL_PATH is because the location becomes / and a nested location within / is fine as is.

junderw:
If you rename your `default_location` file to `default` it will serve static files at //yourdomain/static and //yourdomain/staticfiles and your container at //yourdomain/passport

warren-ru:
> 
> 
> I think maybe you want to use `default` and not `default_location`
> 
> The former will be included once outside all the VIRTUAL_PATH locations, the latter (with _location) will be included in EACH location of VIRTUAL_PATH.
> 
> The reason it works without VIRTUAL_PATH is because the location becomes / and a nested location within / is fine as is.

Thank you! I used 'default' and it works great.
But what if someone uses different VIRTUAL_PATH with many services. What VIRTUAL_PATH would the location be nested in?

junderw:
> But what if someone uses different VIRTUAL_PATH with many services. What VIRTUAL_PATH would the location be nested in?

1. get the sha1 of the VIRTUAL_PATH (ie. for /passport/ use `echo -n /passport/ | sha1sum | awk '{ print $1 }'`)
2. create file called `/etc/nginx/vhost.d/${VIRTUAL_HOST}_${SHA1HASH}_location`
3. make sure any location inside that file has the VIRTUAL_PATH at the beginning (like I showed you)

nginx.tmpl
```diff
@@ -69,6 +70,8 @@ location {{ .Path }} {
 
 	{{ if (exists (printf "/etc/nginx/vhost.d/%s_location" .Host)) }}
 		include {{ printf "/etc/nginx/vhost.d/%s_location" .Host}};
+	{{ else if (exists (printf "/etc/nginx/vhost.d/%s_%s_location" .Host (sha1 .Path) )) }}
+		include {{ printf "/etc/nginx/vhost.d/%s_%s_location" .Host (sha1 .Path) }};
```
junderw, line 74: This should be in order of most specific.

If you have a host-wide location, and a single path_location for one path, then that path should prioritize the path location imo

```suggestion
	{{ if (exists (printf "/etc/nginx/vhost.d/%s_%s_location" .Host (sha1 .Path) )) }}
		include {{ printf "/etc/nginx/vhost.d/%s_%s_location" .Host (sha1 .Path) }};
	{{ else if (exists (printf "/etc/nginx/vhost.d/%s_location" .Host)) }}
		include {{ printf "/etc/nginx/vhost.d/%s_location" .Host}};
```

junderw: Current behavior means if you have a host_location, all host_path_locations are ignored.

AlexanderLieret: That makes more sense.

README.md
junderw: Updated the suggestion.

junderw:
tACK ef9cd4c42e50ca15ca1f7d84c8648d67bd24fee1

junderw:
I think the quality of this PR is sufficient to be merged into dev branch in its current state.

buchdag:
Agreed.

I think the next step would be to add more tests, we should have caught something like the SSL being broken with virtual paths.

buchdag:
Further tests and remarks / discussions in #1733

PR #1725, opened 2021-08-04
Use sha1 of hostname as upstream names

### LINKED ISSUE #1693: Raw Ipv6 `VIRTUAL_HOST` broken on 0.9.1
As of updating to nginx-proxy version 0.9.1 it is no longer possible to set a `VIRTUAL_HOST` corresponding to a raw ipv6 address. 

When attempting this, the entire nginx proxy configuration errors out and falls back to the default error page:

```
Welcome to nginx!
If you see this page, the nginx web server is successfully installed and working.
Further configuration is required.
```

The logs show the following error message:
```
nginx-proxy_1                        | 2021/07/14 15:02:32 [emerg] 52#52: invalid host in upstream "[2001:638:(rest of ipv6 address censored)]-upstream" in /etc/nginx/conf.d/default.conf:108
nginx-proxy_1                        | nginx: [emerg] invalid host in upstream "[2001:638:(rest of ipv6 censored)]-upstream" in /etc/nginx/conf.d/default.conf:108
```

I rolled back to 0.9.0 and confirmed the issue does not occur there. 

I suspect the culprit is https://github.com/nginx-proxy/nginx-proxy/pull/1163, as `[]`-bracketed ipv6 addresses do not seem to be valid upstream names when adding an `upstream` suffix.

buchdag:
This PR adds tests for raw IPv4/IPv6 `VIRTUAL_HOST` and fixes #1693 by using sha1 hashes of the hostname as `upstream` names (which was already the case for regexp `VIRTUAL_HOST`) rather than suffixing the hostname with `-upstream`.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -153,8 +153,7 @@ server {
 {{ range $host, $containers := groupByMulti $ "Env.VIRTUAL_HOST" "," }}
 
 {{ $host := trim $host }}
-{{ $is_regexp := hasPrefix "~" $host }}
-{{ $upstream_name := (print (when $is_regexp (sha1 $host) $host) "-upstream") }}
+{{ $upstream_name := sha1 $host }}
 
 # {{ $host }}
 upstream {{ $upstream_name }} {
--- a/test/test_raw-ip-vhost.py
+++ b/test/test_raw-ip-vhost.py
@@ -1,0 +1,15 @@
+import pytest
+
+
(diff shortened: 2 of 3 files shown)


buchdag:
@tkw1536 could you take a look at this when you have the time ?

tkw1536:
I'll try and have a look in the next couple of days.

tkw1536 (approved):
LGTM.

I've also locally run the image, and this no longer triggers #1693 for me.

Issue #1162, opened 2018-08-21
Upstream naming pre-empts proxy-passing to other load-balanced services


cheald:
I have a cluster of hosts which run things like:

app.domain.com
microservice1.domain.com

I have nginx-proxy set up so that certain location blocks get proxy_passed to the microservice:

```
location /service {
    proxy_pass http://microservice1.domain.com:1234/
}
```

However, this causes nginx to get confused, because the upstream for the microservice is _also_ named `microservice1.domain.com`, which results in errors like:

```
[emerg] 34#34: upstream "microservice1.domain.com" may not have port 1234 in /etc/nginx/vhost.d/app.domain.com:9
```

What I want to do is for that microservice request to resolve DNS and hit my load balancer, which will then get distributed to one of my upstreams running nginx-proxy. What happens is that nginx-proxy instead sees an upstream with that domain name and attempts to proxy to it. You can't proxy_pass to a defined upstream with a port, so it gets confused.

An easy fix would be to suffix upstream naming with some string so they can't be confused with FQDNs, like:

### Before
```
upstream app.domain.com {
  # ...
}

upstream microservice1.domain.com {
  # ...
}

server {
  location /service {
    proxy_pass http://microservice1.domain.com:1234/
  }
  proxy_pass http://app.domain.com/
}
```

### After
```
upstream app.domain.com-upstream {
  # ...
}

upstream microservice1.domain.com-upstream {
  # ...
}

server {
  location /service {
    proxy_pass http://microservice1.domain.com:1234/
  }
  proxy_pass http://app.domain.com-upstream/
}
```

PR #1736, opened 2021-08-17
Revert to predictable upstream names, optional SHA1 names

buchdag:
I decided to backpedal and revert both #1163 and #1725 so that people that relied on predictable `upstream` blocks names won't have their setup break on them.

As an alternative fix to #1162, I added an optional `SHA1_UPSTREAM_NAME` environment variable to nginx-proxy which, when set to `true`, will forces the `upstream` block's names to SHA1 hashes of the hostnames in the same manner as #1725 did.

@cheald I believe this PR will solve your original issues without breaking stuff for others like the previous solution did.

@akpircher, @tkw1536, @acortelyou, @EricSoroos sorry for the back and forth on this one.


tkw1536:
I am fine with the content of this PR, but I'm not sure the word "predictable" in the documentation is correct. IMO the sha1 hash of a domain name is perfectly predictable - maybe `unhashed vs sha1` instead?

buchdag:
@tkw1536 I think I'll go with `Unhashed vs SHA1 upstream names`.

PR #1737, opened 2021-08-18
feat: redirect non-GET methods using 308 instead of 301

### LINKED ISSUE #1738: Non GET HTTP to HTTPS redirect does not work properly.
Calling post request to http endpoint will instead call get method on the reverse proxy.

See logs
```
nginx-proxy          | nginx.1     | <REDACTED> - - [18/Aug/2021:09:23:31 +0000] "POST /api/v1/register HTTP/1.1" 301 169 "-" "PostmanRuntime/7.28.2" "-"
asi-server           | 2021-08-18 09:23:32.081 [DefaultDispatcher-worker-1] TRACE ktor.application - 404 Not Found: GET - /register
nginx-proxy          | nginx.1     | <REDACTED> - - [18/Aug/2021:09:23:32 +0000] "GET /api/v1/register HTTP/1.1" 404 0 "http://<REDACTED>/api/v1/register" "PostmanRuntime/7.28.2" "172.19.0.5:8080"
```
But calling https directly works just fine

```
asi-server           | 2021-08-18 09:27:17.602 [DefaultDispatcher-worker-3] TRACE ktor.application - 400 Bad Request: POST - /register
nginx-proxy          | nginx.1     | <REDACTED> - - [18/Aug/2021:09:27:17 +0000] "POST /api/v1/register HTTP/1.1" 400 62 "-" "PostmanRuntime/7.28.2" "<REDACTED>"
```

My docker-compose env variables is 

```
- VIRTUAL_HOST=REDACTED
- VIRTUAL_PORT=8080
- VIRTUAL_PATH=/api/v1/
- VIRTUAL_DEST=/
```

_Originally posted by @OliverRhyme in https://github.com/nginx-proxy/nginx-proxy/discussions/1733#discussioncomment-1199655_

junderw:
Reference: This post in discussions: https://github.com/nginx-proxy/nginx-proxy/discussions/1733#discussioncomment-1199655

Fixes #1738 

Behavior:
- POST requests to http with redirect are changed to GET requests and the request body is dropped.
Expected:
- POST requests should be forwarded to as is.

Reason: 301 responses cause clients to change the method to GET.

Solution: For non-GET methods, use 307 redirect, which guarantees method and request info is maintained.


junderw:
Got the idea from here: https://serverfault.com/a/1056120

OliverRhyme:
> Google Drive uses a non-standard 308, "Resume Incomplete"...
 > Yeah, maybe the solution is: "don't use http to POST stuff"
 > (Edit: ... maybe the internet was a big mistake... _existential crisis_)
 
 what do you think we should do or workarounds?

junderw:
301 for GET and 308 for everything else is at least better than current behavior...
Just hope that no one ever sets up a nginx proxy in front of the Google Drive API for some reason.

Currently:
non-GET = broken in all cases

With 308 for non-GET:
non-GET = broken in rare cases

buchdag:
I'm not yet 100% comfortable redirecting everything (`GET` included) with 308 instead of 301.

Nevertheless, [SEO wise 308 should be treated exactly like 301 by Google](https://developers.google.com/search/docs/advanced/crawling/301-redirects).

buchdag:
@junderw could you rework this PR to target / be based on `main` instead ?

buchdag:
Maybe the 301 / 308 redirect behavior should be documented ?

junderw:
- Base on main
- Added documentation
- Changed PR target to main

junderw:
Added tests for all available methods on the nginxproxy object.

OliverRhyme:
@junderw can you link issue #1738? So that ill close automatically

junderw:
Ok, done.

AlexanderLieret:
This feature should not be enabled by default. This just hides the underlying problem that the request uri is plain http. The website should generate proper https links instead. 
The intial request still contains the (private) payload as well. This creates a false impression of security.

For example the user opens the login page which was loaded with https and thus secured. After entering the credentials the login button create a post request with the http request uri. The current behavior would create errors and thus alerts the admin of this problem. This PR will just redirect it properly and thus hide the problem. Any further request still can be unencrypted and thus leak any private data. Any malicious attacker can intercept the traffic without the user noticing at all.

TLDR: the http to https redirect should only be intended to redirect the initial unencrypted request as it currently is.

OliverRhyme:
> This feature should not be enabled by default. This just hides the underlying problem that the request uri is plain http. The website should generate proper https links instead. 
> The intial request still contains the (private) payload as well. This creates a false impression of security.
> 
> For example the user opens the login page which was loaded with https and thus secured. After entering the credentials the login button create a post request with the http request uri. The current behavior would create errors and thus alerts the admin of this problem. This PR will just redirect it properly and thus hide the problem. Any further request still can be unencrypted and thus leak any private data. Any malicious attacker can intercept the traffic without the user noticing at all.
> 
> TLDR: the http to https redirect should only be intended to redirect the initial unencrypted request as it currently is.

I don't think that's our problem but rather the problem of whoever consumes the api to always use https. Nevertheless the current behavior is the same where the initial request will use http, so I think this concern has nothing to do with this pr as this does not change the current default behavior.

AlexanderLieret:
>I don't think that's our problem but rather the problem of whoever consumes the api to always use https. Nevertheless the current behavior is the same where the initial request will use http, so I think this concern has nothing to do with this pr as this does not change the current default behavior.

@OliverRhyme this changes the current default behavior such that the admin will most likely never notice the problem.

The user making the intial request to the website is another problem. But this PR hides the fact that for example the login button makes a plain http request.

If the admin is comfortable to enable this redirect feature, it should be able to be activated, like the no-redirect behavior of http requests.

OliverRhyme:
> >I don't think that's our problem but rather the problem of whoever consumes the api to always use https. Nevertheless the current behavior is the same where the initial request will use http, so I think this concern has nothing to do with this pr as this does not change the current default behavior.
> 
> @OliverRhyme this changes the current default behavior such that the admin will most likely never notice the problem.
> 
> The user making the intial request to the website is another problem. But this PR hides the fact that for example the login button makes a plain http request.
> 
> If the admin is comfortable to enable this redirect feature, it should be able to be activated, like the no-redirect behavior of http requests.
> 

Correct me if I'm wrong but is the default behavior of nginx-proxy when getting http request is to redirect it to the https? So currently it still does hide the fact that for example the login button will still send plain http request? 301 is basically the same with 308 with the caveat that it preserves the http method used

junderw:
> such that the admin will most likely never notice the problem.

This assumes that private data is not sent as query parameters as a GET request.
Any form that sends GET requests with private data will have the same problem.

Your argument is actually supporting for nohttp as default and only supporting https.

Relying on a quirk of client implementations (301 response being redirected with GET always is a client implementation detail) for security is never a good idea.

If we want to prevent insecure things from happening, we should disable them by default on the server side.

If I explicitly set "redirect http to https" I am not saying "only redirect GET requests please."

So I agree that current behavior can save some people in some rare cases, we are picking and choosing who is saved and who is not, and relying on clients to interpret these errors and pick up on why they're bad. (ie. an admin might pick up on the POST failure and solve it by noticing that GET goes through fine, and changing the login form into a GET request.)

The expected behavior is "redirect all http to https" and this PR brings the feature in line with that expectation.

I think a different discussion could be made to reject all http by default. (besides the ACME wellknown path)

junderw:
I think maybe another solution could be to reject all methods besides GET for http.

This way at least the admin will be more likely to arrive to a conclusion of "oh, maybe I'm not supposed to use POST with http"

Though they might still stumble upon GET still working, and adding `method="get"` to their login form to get around it...

OliverRhyme:
I agree with creating another issue to enable disabling redirect http to https all together. I also believe that we can at least alleviate "such that the admin will most likely never notice the problem." by having a strongly worded documentation on what will happen and how to fix or workaround it.

junderw:
tbh though... even if we reject all http requests. They'll still be sending the packets over the clear to the server anyways... so really, I don't think there's much we can do to help people here...

I'd be open to explicitly rejecting non-GET as well rather than sending 301 and having a hard-to-understand error. A rejection is easier to understand... why isn't it there? *looks at URL* OH I forgot https!

as compared to "why is it sending GET when I sent POST? hmmm... *googles "login form get request"* *gets results of "how to set form to get method"* *tries it* *works*

OliverRhyme:
> tbh though... even if we reject all http requests. They'll still be sending the packets over the clear to the server anyways... so really, I don't think there's much we can do to help people here...
> 
> I'd be open to explicitly rejecting non-GET as well rather than sending 301 and having a hard-to-understand error. A rejection is easier to understand... why isn't it there? _looks at URL_ OH I forgot https!
> 
> as compared to "why is it sending GET when I sent POST? hmmm... _googles "login form get request"_ _gets results of "how to set form to get method"_ _tries it_ _works_

But then, this becomes opinionated and maybe we should instead provide them an option.

AlexanderLieret:
Yes, entirely disabling the http part will add more security, like closing port 80. But depending on the browser a single 301 redirect is necessary to have all further requests made using https. Some just try https is http is not working.

For example just entering `github.com` will first create a plain http get request which is answered with `301 https://github.com/`. As long as all (internal) links are https, no http request will be made.

buchdag:
> I agree with creating another issue to enable disabling redirect http to https all together.

@OliverRhyme this feature is already in nginx-proxy

> To serve traffic in both SSL and non-SSL modes without redirecting to SSL, you can include the environment variable `HTTPS_METHOD=noredirect` (the default is `HTTPS_METHOD=redirect`). You can also disable the non-SSL site entirely with `HTTPS_METHOD=nohttp`, or disable the HTTPS site with `HTTPS_METHOD=nohttps`. `HTTPS_METHOD` can be specified on each container for which you want to override the default behavior or on the proxy container to set it globally. If `HTTPS_METHOD=noredirect` is used, Strict Transport Security (HSTS) is disabled to prevent HTTPS users from being redirected by the client. If you cannot get to the HTTP site after changing this setting, your browser has probably cached the HSTS policy and is automatically redirecting you back to HTTPS. You will need to clear your browser's HSTS cache or use an incognito window / different browser.

buchdag:
By the way maybe this PR should be a new `HTTPS_METHOD` option, e.g `redirect308`

OliverRhyme:
> By the way maybe this PR should be a new `HTTPS_METHOD` option, e.g `redirect308`

and make it default i guess? maybe also redirect307 as to avoid conflict with different services that dosent use the 308 as intended

junderw:
The paths I'm seeing are:

1. Add 308 for non-GET. no options added.
2. Add 308 for non-GET. option to disable (making non-GET reject completely (404? 400?)).
3. Make non-GET reject completely. add option to enable with 308.

Maybe 3 is best.

1. `redirect` default will be 301 for GET and 400 for non-GET.
2. `redirectallmethods` will be 308 for all methods (since it won't be default, might as well use 308 for all)

One major PITA is the fact that nginx doesn't support variables for response codes, so the more branches we add, the more if {} statements we need, since we can't use map or anything. Otherwise it would be easier to just have a var for REDIRECT_CODE etc.

buchdag:
At the moment I'm not sold on rejecting non `GET` and not 100% sure about changing the default behavior with `if`s in `location` blocks (something notoriously discouraged by nginx, even if we're in one of the two supposedly 100% safe cases) for an issue that I don't recall seeing here before. Adding the 308 for non GET as a new non default option would be fine.

I'm not saying this is final, I'll have to think a bit more about this one. Any example of `if ($request_method = WHATEVER) {` in `location` blocks being used in production without issue would be welcome.

junderw:
It seems there's a limitation on the return code in nginx, so you can't use a variable as the return code.

Therefore, the only way to change the return code based on some dynamic data is to split the logic using some other way.

The nginx documentation does state that if statements are safe if they `return` or `rewrite` because the only way if statements can get you is if you have a bunch of if statements and expect them all to run.

ie.

```
if ($request_method = GET) {
    set_header X Y
}
if ($scheme = http) {
  set_header A B
}
```

In this case, if you are expecting "when it is GET and http, header X and A should both be set." this is wrong. Because if {} is wonky.

However, if every if {} statement conditions can not possibly overlap, and each if {} ends in return/rewrite, it is safe.

If we could use variables on the return code it would be easy to use map {} to just map the request method to a code.

---

> for an issue that I don't recall seeing here before

I'm totally fine with a "Won't fix" here. This really feels like "I tried shooting myself in the foot, and the bullet didn't come out and it made a weird sound" and we're arguing whether "the bullet should come out" or "we should prevent the bullet from coming out explicitly, without a weird sound."

Maybe the answer is: "Don't try to shoot yourself in the foot." if this was a wide spread problem, you're right, someone would have mentioned the issue by now, since this is an issue from the beginning of the project.

buchdag:
Back to this PR, has anyone used it in production so far ?

buchdag:
@junderw sorry for letting this sit in the dust for so long, are you still interested in getting it merged ?

junderw:
Sure!

I actually stopped using the thing that was using nginx-proxy last year when I had to take down my home lab for various reasons, but I'm still alive and well and willing to push this through. I think it will be helpful.

buchdag:
> If you don't feel like merging it, I'm fine with a "Won't fix" or "Not a bug" decision here.

No, it actually did make sense to me and a few people manifested interest for it, I was rather deterred by the very opinionated discussion about what should or should not be done because of security implications.

I think I got the feature where I wanted it ~3 years ago :
- optional and disabled by default (as suggested by @AlexanderLieret)
- uncoupled from `HTTPS_METHOD`
- with choice between `307` and `308` redirect (AWS use a `307` for the same purpose on Cloudfront)
- with the `HEAD` method using the same redirect as `GET` (again, what AWS does on Cloudfront in this case)

buchdag (approved):

docs/README.md
```diff
@@ -573,6 +573,8 @@ The default behavior for the proxy when port 80 and 443 are exposed is as follow
 - If the virtual host does not have a usable cert, but `default.crt` and `default.key` exist, those will be used as the virtual host's certificate.
 - If the virtual host does not have a usable cert, and `default.crt` and `default.key` do not exist, or if the virtual host is configured not to trust the default certificate, SSL handshake will be rejected (see [Default and Missing Certificate](#default-and-missing-certificate) below).
 
+The redirection from HTTP to HTTPS use by default a [`301`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301) response for every HTTP methods (except `CONNECT` and `TRACE` which are disabled on nginx). If you wish to use a custom redirection response for the `OPTIONS`, `POST`, `PUT`, `PATCH` and `DELETE` HTTP methods (for example to use a ), you can either do it globally with the environment variable `NON_GET_REDIRECT` on the proxy container or per virtual host with the `com.github.nginx-proxy.nginx-proxy.non-get-redirect` label on proxied containers. Valid values are [`307`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) and [`308`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308).
```
buchdag, line 576: typo
```suggestion
The redirection from HTTP to HTTPS use by default a [`301`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301) response for every HTTP methods (except `CONNECT` and `TRACE` which are disabled on nginx). If you wish to use a custom redirection response for the `OPTIONS`, `POST`, `PUT`, `PATCH` and `DELETE` HTTP methods, you can either do it globally with the environment variable `NON_GET_REDIRECT` on the proxy container or per virtual host with the `com.github.nginx-proxy.nginx-proxy.non-get-redirect` label on proxied containers. Valid values are [`307`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) and [`308`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308).
```

junderw:
LGTM so far

buchdag (dismissed):

buchdag:
Rebased on main, I think this can be merged when the tests pass.

PR #1747, opened 2021-08-24
Add more testcases and fixes to Virtual Path

AlexanderLieret:
This PR adds more test cases for the Virtual Path feature (#1607 + #1699).
The error found [here](https://github.com/nginx-proxy/nginx-proxy/discussions/1733#discussioncomment-1227273) is fixed.

Reference: Discussion in #1733.

### DIFF
--- a/network_internal.conf
+++ b/network_internal.conf
@@ -3,4 +3,5 @@ allow 127.0.0.0/8;
 allow 10.0.0.0/8;
 allow 192.168.0.0/16;
 allow 172.16.0.0/12;
+allow fc00::/7; # IPv6 local address range
 deny all;
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -50,6 +50,11 @@
 
 {{ define "location" }}
 location {{ .Path }} {
+	{{ if eq .NetworkTag "internal" }}
+	# Only allow traffic from internal clients
+	include /etc/nginx/network_internal.conf;
+	{{ end }}
+
 	{{ if eq .Proto "uwsgi" }}
 		include uwsgi_params;
 		uwsgi_pass {{ trim .Proto }}://{{ trim .Upstream }};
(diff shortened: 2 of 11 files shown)


network_internal.conf
```diff
@@ -3,4 +3,5 @@ allow 127.0.0.0/8;
 allow 10.0.0.0/8;
 allow 192.168.0.0/16;
 allow 172.16.0.0/12;
+allow fc00::/7; # IPv6 local address range
```
junderw, line 6: Should also include `::1/128` loopback IPv6 (there's only one, in comparison to IPv4's huge loopback block)

junderw (changes requested):
Other than one addition, it LGTM.

Not sure how we could test the internal network issue though, can we mock it so that nginx thinks python is requesting from a non-local IP?

network_internal.conf
AlexanderLieret: I am not sure if we need any loopback at all. Loopback in this scenario would mean that the proxy container itself makes a request to localhost. It could be added.

AlexanderLieret:
> Not sure how we could test the internal network issue though, can we mock it so that nginx thinks python is requesting from a non-local IP?

Currently there are no test at all for the internal network feature. Spoofed IPs require some configuration to the host OS.

One simpler approach is to mount a modified file which adds a line to the response header and then check if it is added or not.

AlexanderLieret:
> One simpler approach is to mount a modified file which adds a line to the response header and then check if it is added or not.

I did add such test cases.

@buchdag can you approve the awaiting workflow?

junderw:
utACK c225c82

Tests failing due to test module collision.

buchdag:
@AlexanderLieret I'll take a look at the PR by next week.

JMLX42:
@buchdag any chance it can get reviewed/merged any time soon? Thanks a lot!

PR #1779, opened 2021-09-13
Add an environment variable to enable CORS requests with basic authentication

vemonet:
Added the environment variable `ENABLE_CORS_AUTH` to disable basic authentication for HTTP OPTIONS to enable CORS requests preflights.

The default behavior does not change, authentication for OPTIONS is still enabled by default

This feature has been mentioned here: https://github.com/nginx-proxy/nginx-proxy/issues/1150
And a first pull request proposed here: https://github.com/nginx-proxy/nginx-proxy/pull/1176

I am not sure about the name of the env variable: `ENABLE_CORS_AUTH` to disable auth when CORS OPTIONS... Might be a bit misleading. Could be also `ENABLE_PASS_CORS_AUTH`, let me know if we should change it

Thanks @tkw1536 for the feedback on the issue

And thanks for this amazing project

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,6 +317,14 @@ docker run -d -p 80:80 -p 443:443 \
 
 You'll need apache2-utils on the machine where you plan to create the htpasswd file. Follow these [instructions](http://httpd.apache.org/docs/2.2/programs/htpasswd.html)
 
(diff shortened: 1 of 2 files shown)


junderw:
Needs tests

PR #1797, opened 2021-10-10
feat: Use RFC 7919 DH groups + Remove DH generation

polarathene:
This PR replaces the Diffie-Hellman parameter generation with standardized RFC 7919 DH groups (_2048, 3072, 4096 bit lengths_). These files are available from [this repo](https://github.com/internetstandards/dhe_groups) These DH groups are advised to use vs generating your own. The [openssl wiki advises adopting RFC 7919 too](https://wiki.openssl.org/index.php/Diffie-Hellman_parameters#RFC_7919_Groups) and provides the ASCII PEM encoded content with which you can verify against the param files in this PR as the same.

- The DH params feature was [originally introduced in 2017](https://github.com/nginx-proxy/nginx-proxy/pull/589).
- [In June 2020, the default DH group size was increased to 4096-bit](https://github.com/nginx-proxy/nginx-proxy/pull/1452/files), and the `openssl` command to generate params had the `-dsaparam` arg added - which trades some security for reducing generation time (_[2016 CVE advisory](https://www.openssl.org/news/secadv/20160128.txt), OpenSSL [docs warn that these generated params are only suitable for **single use only**](https://www.openssl.org/docs/man1.1.1/man1/dhparam.html#OPTIONS) - although that is a non-issue for DHE usage_). There was no justification or discussion related to this in the PR, with the project author merging it regardless..

### Commit Overview

For ease of review, you will find logical commits with associated commit messages. Some of these commits are sequenced changes to better grok the PR diff.

You may find it more comfortable to review in this manner via a git GUI, I use GitKraken personally.

- The default group size remains at `4096-bit`, however `DHPARAM_BITS` now only supports lengths of `2048|3072|4096`. Requesting other sizes will fail with an error logged and then exit the container. These should be sufficient, otherwise support for providing a custom `dhparam.pem` file remains.
- `DHPARAM_GENERATION=false` has been replaced in favor of `DHPARAM_SKIP=1`. There is no generation involved now, nor overhead to be concerned about. This ENV allows preventing the pre-generated DH params from being copied over to `/etc/nginx/dhparam`, which prevents `DHE` cipher suites from being used (_provided no custom DH params file was volume mounted_).
- The removal of the generation feature now requires any need for 1024-bit DH params (_which is generally considered ill-advised, but required for some legacy software_) to no longer rely on `nginx-proxy` to provide/generate the dhparam file for that size. Users requiring this tend to be from enterprise backgrounds and should know how to create their own DH key to supply `nginx-proxy` with explicitly.
- Originally [a pre-generated `dhparam.pem` was added to `nginx-proxy` from a contributor](https://github.com/nginx-proxy/nginx-proxy/pull/589#issuecomment-272072531) (_used as a fallback while generating a new parameters file_). This was generated on their machine,  and has now been removed in favor of the RFC 7919 files added to repo folder at `dhparam/` (_located at `/app/dhparam/` inside the container_). I have left the copy in the `test/lib/ssl/dhparam.pem` alone. All tests (_except for one specific to DH_) have dropped the volume mount for that test `dhparam.pem` file.
- The DH generation test has been removed. The remaining DH test has been notably improved to cover additional test cases that weren't previously handled, and to fix an invalid test (_asserts the wrong value, which was misleading_). 
- The Dockerfile `VOLUME` directive created anonymous volumes unless you explicitly provided named volumes. This is undesirable and introduced early on by the original author of this project without justification. I believe it was unintentional. The `/etc/nginx/dhparam` location still needs to exist at run-time, so it is now created at build-time.
- `docker-entrypoint.sh` has been refactored. It now includes the DH setup logic, and splits out two separate checks into their own wrapped functions. The conditional exit failure has also been relocated and handled better ([re-implements an earlier 2015 approach](https://github.com/nginx-proxy/nginx-proxy/commit/1aac9960197f56d502a6147337337fd3b1bfa365)),
- The README has been updated accordingly too.


buchdag:
Impressive work, thank you. 👍

buchdag:
> The entrypoint change I referenced overlooked that there was an intention to output the warnings/errors regardless of what CMD was given to the container. It may be desirable to restore that?

In my opinion the way you did it in the PR is the correct one, if we `exec` or `run` let's say `bash` into the container, we don't want any of the initialisation / verifications functions from the entrypoint wrapper script to be run. This is the way **acme-companion** works and what I intended to do with **nginx-proxy** at some point.

buchdag:
>  `DHPARAM_GENERATION=false` has been replaced in favor of `DHPARAM_SKIP=1`. There is no generation involved now, nor overhead to be concerned about. This ENV allows preventing the pre-generated DH params from being copied over to `/etc/nginx/dhparam`, which prevents `DHE` cipher suites from being used (_provided no custom DH params file was volume mounted_).

Ideally we should maintain backward compatibility with the old environment variable and / or output a deprecation warning if it's set.

buchdag:
> The removal of the generation feature now requires any need for 1024-bit DH params (_which is generally considered ill-advised, but required for some legacy software_) to no longer rely on `nginx-proxy` to provide/generate the dhparam file for that size. Users requiring this tend to be from enterprise backgrounds and should know how to create their own DH key to supply `nginx-proxy` with explicitly.

I think this warning / explanation should be output by the container if `DHPARAM_BITS` is set to `1024` bits (in addition to exiting).

polarathene:
> I think this warning / explanation should be output by the container if `DHPARAM_BITS` is set to `1024` bits (in addition to exiting).

That's technically already handled with this error:

> "ERROR: Unsupported DHPARAM_BITS size: ${DHPARAM_BITS}. Use: 2048, 3072, or 4096 (default)."

The user then knows to use only a supported value, or they'll presumably report an issue about it before checking the README or release notes.

If you get a few repetitive issues/discussions about it, that could make sense to be more explicit about `1024`, but I doubt it'll happen much in practice.

The entrypoint outputting the detected config issues for different CMD (without exiting like the default CMD would), was added specifically because many users were running into those problems that the entrypoint detects, but apparently not being aware of it if they launched the container with a shell to investigate :man_shrugging: 

I guess likewise, if that continues to be an actual issue for some reason, bring that back. I agree that it shouldn't be necessary, just thought I'd make you aware that I deviated from that decision/behaviour.

---

> Agreed, I think we should also mkdir -p '/etc/nginx/certs' just to be sure.

Are you aware of a case where that'd be necessary? The related tests config I tried to remove the certs directory but learned that it was required to enable implicit HTTPS support via the generated template.

Users are instructed to volume mount this location for persistence as well, so I'm not sure if that's necessary. I didn't check if the directory location already exists in the image by default though.

Might be worth logging that HTTPS is not enabled if that directory has no content to trigger the HTTPS template logic, as when working on this PR it took me a while to realize the trigger (_I had to look up the template source_) :sweat_smile: 

`/etc/nginx/dhparam` on the other hand didn't exist and needs to at run-time for the DH logic to copy over the files. Seemed appropriate to just create that in the image up front as users would not be mounting that location locally.

---

### Regarding experience contributing tests

It might be worth referencing this test `.yml` and `.py` for contributors as an example. I was not familiar with PyTest and lost most time trying to figure out and understand that (_initially, I was not aware of the existing project docs about it either, but that's on me_ :sweat_smile: )

One of the main challenges I had was multiple `nginx-proxy` instances, and not knowing how some of the magic PyTest was setup to do was working (_some errors weren't helpful_). I mostly didn't want to write many individual test files/configs to split out each `nginx-proxy` instance.

I recall trying to figure out how to interact with the container, use the assert logs call to specific containers (which I modified and could be extracted to a more common utility), access container ENV, etc.

I only used the PyTest docker image for running the tests, and for whatever reason that seemed to have network issues preventing some of the tests passing locally. I did not try running the tests without the test container setup, so I'm not sure if it's an issue due to that test container config or with my system (ubuntu 21.04 linux VPS with latest docker engine).

One other issue was the `.yml` configs. These don't mention `version` or `services`, I've only used `docker-compose` not `docker compose`, not sure if that has something to do with it, or if these configs are similar to compose but parsed by something else. When I tried to use some features it seemed they were not supported, I think it was being treated as a v1 docker-compose config, but adding a `version` property also didn't play well with the tests.

PR #1819, opened 2021-11-04
Clarify grammar in DH-GROUP section

HPPinata:
Since the second option got removed, the "either" makes no sense any more and may lead to confusion about non existent alternatives


buchdag:
Thanks !

PR #1888, opened 2022-02-14
Fix path to default.conf in README.md

nathanweeks:
`/etc/nginx/conf.d/default.conf` is accidentally listed as `/etc/nginx/conf.d/default` a couple times in README.md


buchdag:
Nice catch, thank you !

PR #1934, opened 2022-03-28
feat: Add support for HTTP keep-alive between the proxy and upstream

### LINKED ISSUE #609: How Do I Enable Keepalive connections
Please, what's the recommended way to enable `keepalive` connections.
I'm supposed to add something like
`keepalive 100;`
in the upstream directive, and 
`proxy_http_version 1.1;`
`proxy_set_header Connection "";`
in that of `location` correct?
But since the confs are generated, I can't seem to figure out a way to add these without breaking anything.

rhansen:
Fixes #609

Note: This depends on PR #2127, and will be marked as draft until that PR is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,6 +373,13 @@ docker run -d -p 80:80 -p 443:443 \
 
 You'll need apache2-utils on the machine where you plan to create the htpasswd file. Follow these [instructions](http://httpd.apache.org/docs/2.2/programs/htpasswd.html)
 
+### Upstream (Backend) Server HTTP Keep-Alive Support
+
+> **Warning**
+> This feature is experimental.  The behavior may change (or the feature may be removed entirely) without warning in a future release, even if the release is not a new major version.  If you use this feature, or if you would like to use this feature but you require changes to it first, please [provide feedback in #2194](https://github.com/nginx-proxy/nginx-proxy/discussions/2194).  Once we have collected enough feedback we will promote this feature to officially supported.
+
+To enable HTTP keep-alive between `nginx-proxy` and a backend server, set the `com.github.nginx-proxy.nginx-proxy.keepalive` label on the server's container to the desired maximum number of idle connections. See the [nginx keepalive documentation](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive) and the [Docker label documentation](https://docs.docker.com/config/labels-custom-metadata/) for details.
+
 ### Headers
 
 By default, `nginx-proxy` forwards all incoming request headers from the client to the backend server unmodified, with the following exceptions:
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -176,6 +176,7 @@
     {{- if exists $override }}
     include {{ $override }};
     {{- else }}
+        {{- $keepalive := first (keys (groupByLabel .Containers "com.github.nginx-proxy.nginx-proxy.keepalive")) }}
     location {{ .Path }} {
         {{- if eq .NetworkTag "internal" }}
         # Only allow traffic from internal clients
@@ -189,10 +190,14 @@
         root {{ trim .VhostRoot }};
         include fastcgi_params;
         fastcgi_pass {{ trim .Upstream }};
+            {{- if $keepalive }}
+        fastcgi_keep_conn on;
+            {{- end }}
         {{- else if eq .Proto "grpc" }}
         grpc_pass {{ trim .Proto }}://{{ trim .Upstream }};
         {{- else }}
         proxy_pass {{ trim .Proto }}://{{ trim .Upstream }}{{ trim .Dest }};
+        set $upstream_keepalive {{ if $keepalive }}true{{ else }}false{{ end }};
         {{- end }}
 
         {{- if (exists (printf "/etc/nginx/htpasswd/%s" .Host)) }}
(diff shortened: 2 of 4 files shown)


rhansen:
I'm not super satisfied with the tests I added, but I couldn't think of a good way to thoroughly test the feature (e.g., grepping the generated config file seems too fragile).

rhansen:
Friendly ping @buchdag

rhansen:
Marking as draft until #2127 is merged.

buchdag:
@rhansen I've got mixed feeling about the use of labels. While I too do prefer labels to environment variables (as they inherently avoid stuff like #2139), I think we should stick to environment variables unless we're prepared to migrate everything to labels, as we don't use labels for anything yet. I also believe that people are more familiar with environment variables than they are with labels.

rhansen:
> I think we should stick to environment variables unless we're prepared to migrate everything to labels, as we don't use labels for anything yet. I also believe that people are more familiar with environment variables than they are with labels.

Good points.  However, assuming we will eventually migrate the existing environment variables to labels (I think we should; they are designed for this purpose), using an environment variable for this feature makes more work for us now and later.  I'd rather live with the short-term inconsistency.

buchdag:
@rhansen the way I see it, in the end we should keep environment variables for configuration of the nginx-proxy (or docker-gen) container, and switch to labels on the proxied containers. Do you agree with this ?

rhansen:
I'm not sure.  There are pros and cons either way.  Advantages of environment variables for the `nginx-proxy` container:
  * Less work: we don't need to migrate them to labels like we do with the environment variables on proxied containers.
  * Having distinct ways of configuring `nginx-proxy` vs. the other containers might make configs more readable (environment variables are for the container's own use, labels are for use by a different container).

Advantages of labels:
  * Consistency: all configuration is done via labels (and include files).
  * Less risk of collision with environment variables used by other code in the `nginx-proxy` container (nginx itself, a custom nginx module added by the user, the shell, `docker-gen`, Docker API libraries, logging libraries, etc.).
  * Less pollution of the environment seen by various processes.
  * It might be easier to prevent undesired disclosure of privacy- or security-sensitive values:  Environment variables are normally copied to every spawned process, which makes it easier for an attacker to find a way to inspect the configuration.  It should be easier to lock down access to the Docker socket to prevent such disclosure.  (I haven't carefully studied the security implications so labels might actually increase the risk.)

rhansen:
I opened #2148 for further discussion about environment variables vs. labels.

buchdag:
@rhansen I'll review this again tonight or  tomorrow but it should be okay to merge before next release.

buchdag:
@rhansen I did not notice the experimental status of the feature. What would be needed to confirm that it work as intended ?

rhansen:
> What would be needed to confirm that it work as intended ?

Time, or discussions/bug reports.  We don't have a way of collecting usage statistics, so there's not much we can do other than eventually think to ourselves, "this feature has been around for a while and nobody has complained, so it's probably OK."

SchoNie:
keepalive is not experimental and is even [recommended to have enabled.](https://www.nginx.com/blog/avoiding-top-10-nginx-configuration-mistakes/#no-keepalives) I do like the way of this implementation by @rhansen (also I like the docker label concept)
Currently I manually overwrite the template file to have the keepalive directions in there. So having this merged would help me to run this container clean without any overrides.

rhansen:
> keepalive is not experimental

Agreed; the experimental thing here is the *ability to enable* keepalive in `nginx-proxy`, not keepalive itself. 

> I do like the way of this implementation by @rhansen (also I like the docker label concept)

Thanks for the feedback!

rhansen:
> I suspect there is a problem setting the `proxy_set_header Connection $proxy_connection_noclose;` in the location block.

Good catch, thank you @SchoNie!

> Maybe a better way would be to wrap the map in an if, and overwrite the first $proxy_connection (it will stay upgrade if received, else it clears the close with an empty string).

Thanks for the suggestion.  I went with a different approach that preserves per-upstream configurability (it can be turned on for some containers and left off for others).  I also updated the tests to check for the regression you found.  If you have the time, I would appreciate it if you looked at the new revision.

SchoNie:
Tested, LGTM.
Good you added a X-Real-IP test!

README.md
```diff
@@ -373,6 +373,13 @@ docker run -d -p 80:80 -p 443:443 \
 
 You'll need apache2-utils on the machine where you plan to create the htpasswd file. Follow these [instructions](http://httpd.apache.org/docs/2.2/programs/htpasswd.html)
 
+### Upstream (Backend) Server HTTP Keep-Alive Support
+
+> **Warning**
+> This feature is experimental.  The behavior may change (or the feature may be removed) without warning in a future release, even if the release is not a new major version.  If you use this feature, please provide feedback so we know whether it is working properly and can be promoted to officially supported.
```
buchdag, line 379: Could you open a discussion to provide feedback on this feature and put a link to this discussion here ? I think it can be merged once this is done.

buchdag (changes requested):

README.md
rhansen: Done.

buchdag (approved):

buchdag:
@rhansen could you just wait a few days before merging this ? I'd like to release a patch version of docker-gen and then of nginx-proxy before merging this feature and the load balancing one from @SchoNie

Issue #1504, opened 2020-10-01
Allow per-hostname ports


dajester2013:
The current template allows for separate hostnames, but they all use the same port for the upstream.

I want to proxy the sonatype/nexus3 container, which starts the docker registry on a separate port from the standard service (8082 for docker vs 8081 for maven).

I need to be able to configure two hostnames for the proxy, with two separate upstream ports.

I would think that you since you can do ` -e VIRTUAL_HOST=host1,host2`, you should also be able to do `-e VIRTUAL_PORT=8081,8082`

dajester2013:
I have a workaround - I modified the template and mounted it as a volume when starting the proxy container.

workaround template:
```
{{ $CurrentContainer := where $ "ID" .Docker.CurrentContainerID | first }}

{{ $external_http_port := coalesce $.Env.HTTP_PORT "80" }}
{{ $external_https_port := coalesce $.Env.HTTPS_PORT "443" }}

{{ define "upstream" }}
	{{ if .Address }}
		{{/* If we got the containers from swarm and this container's port is published to host, use host IP:PORT */}}
		{{ if and .Container.Node.ID .Address.HostPort }}
			# {{ .Container.Node.Name }}/{{ .Container.Name }}
			server {{ .Container.Node.Address.IP }}:{{ .Address.HostPort }};
		{{/* If there is no swarm node or the port is not published on host, use container's IP:PORT */}}
		{{ else if .Network }}
			# {{ .Container.Name }}
			server {{ .Network.IP }}:{{ .Address.Port }};
		{{ end }}
	{{ else if .Network }}
		# {{ .Container.Name }}
		{{ if .Network.IP }}
			server {{ .Network.IP }} down;
		{{ else }}
			server 127.0.0.1 down;
		{{ end }}
	{{ end }}

{{ end }}

{{ define "ssl_policy" }}
	{{ if eq .ssl_policy "Mozilla-Modern" }}
		ssl_protocols TLSv1.3;
		{{/* nginx currently lacks ability to choose ciphers in TLS 1.3 in configuration, see https://trac.nginx.org/nginx/ticket/1529 /*}}
		{{/* a possible workaround can be modify /etc/ssl/openssl.cnf to change it globally (see https://trac.nginx.org/nginx/ticket/1529#comment:12 ) /*}}
		{{/* explicitly set ngnix default value in order to allow single servers to override the global http value */}}
		ssl_ciphers HIGH:!aNULL:!MD5;
[...]

PR #1938, opened 2022-04-06
feat: Use container names for upstream names to avoid repeats

rhansen:
Note: This depends on PR #2127, and will be marked as draft until that PR is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -587,11 +587,73 @@ location / {
 #### Per-VIRTUAL_HOST `server_tokens` configuration
 Per virtual-host `servers_tokens` directive can be configured by passing appropriate value to the `SERVER_TOKENS` environment variable. Please see the [nginx http_core module configuration](https://nginx.org/en/docs/http/ngx_http_core_module.html#server_tokens) for more details.
 
-### Unhashed vs SHA1 upstream names
+### Upstream name style
 
-By default the nginx configuration `upstream` blocks will use this block's corresponding hostname as a predictable name. However, this can cause issues in some setups (see [this issue](https://github.com/nginx-proxy/nginx-proxy/issues/1162)). In those cases you might want to switch to SHA1 names for the `upstream` blocks by setting the `SHA1_UPSTREAM_NAME` environment variable to `true` on the nginx-proxy container.
+The generated nginx config will have one or more `upstream` blocks, each of which defines a [server group](https://nginx.org/en/docs/http/ngx_http_upstream_module.html) for the applicable containers. The name of the upstream is computed in one of three different ways (or "styles"), described below. The desired style is selected by setting the `UPSTREAM_NAME_STYLE` environment variable on the nginx-proxy container.
 
-Please note that using regular expressions in `VIRTUAL_HOST` will always result in a corresponding `upstream` block with an SHA1 name.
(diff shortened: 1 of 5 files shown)


buchdag:
> {{- /* TODO: Replace the next few lines with a call to `join` once nginx-proxy has a version of
>          docker-gen that includes https://github.com/nginx-proxy/docker-gen/pull/418. */}}

I'm on it: #1940

rhansen:
Thanks for updating docker-gen. I rebased and switched to the `join` function.

buchdag:
Thanks for the PR, unfortunately we can't make this behaviour a default one because a lot of people have custom configurations that rely on upstream names being equal to the hostnames.

See #1725, https://github.com/nginx-proxy/nginx-proxy/issues/1693#issuecomment-897917888, #1736 plus the release notes of version [`0.9.2`](https://github.com/nginx-proxy/nginx-proxy/releases/tag/0.9.2) and [`0.9.3`](https://github.com/nginx-proxy/nginx-proxy/releases/tag/0.9.3).

rhansen:
I updated the PR to keep the current behavior by default.

rhansen:
Friendly ping @buchdag

rhansen:
Marking as draft until #2127 is merged.

buchdag:
@rhansen sorry for the incredibly long delay on this one. I've tested the PR, and I'm still a bit uncertain about it.

Does it provide any advantage / behaviour change when not using `VIRTUAL_PATH` ?

What kind of issue could the "repeat" of upstreams when using `VIRTUAL_PATH` cause ?

I feel like I'm completely missing the point but I seem unable to grasp what issue this PR is attempting to solve or its benefits against it non trivial added complexity.

I also don't think it could realistically be made default as it would made upstream name somewhere between very hard to predict and unpredictable on systems or setups where container names are dynamically generated or containers replica dynamically started / stopped.

rhansen:
> Does it provide any advantage / behaviour change when not using `VIRTUAL_PATH` ?

Yes.  Currently, if a single container is used for multiple vhosts (e.g., `VIRTUAL_HOST: foo.example,bar.example`) then multiple identical `upstream` blocks are created, one per vhost.  With this change only one `upstream` is generated and shared with both vhosts.

> 
> What kind of issue could the "repeat" of upstreams when using `VIRTUAL_PATH` cause ?

`VIRTUAL_PATH` is irrelevant to this change.

I think the repeated upstreams could theoretically cause minor load balancing issues.

> 
> I feel like I'm completely missing the point but I seem unable to grasp what issue this PR is attempting to solve or its benefits against its non trivial added complexity. thinking

It's been a year so a lot has slipped away from memory, but IIRC my main motivation was to facilitate a fix for #1504.  Give me some time to recall what I was doing and I'll update the commit message with some rationale.

> 
> I also don't think it could realistically be made default as it would made upstream name somewhere between very hard to predict and unpredictable on systems or setups where container names are dynamically generated or containers replica dynamically started / stopped.

Hmm, that's a good point.  Let me think about this some more to see if there's a good general solution that checks all of these boxes:
  * avoids duplicates
  * avoids the need to encode a path in the upstream name
  * is stable
  * is predictable
  * is human-readable

PR #1939, opened 2022-04-06
chore: indent location, upstream in template

ntnj:


buchdag:
@ntnj thanks for the PR, could you change the commit message to something under 50 characters that begin with `chore:` ?

ntnj:
Done. Let me know if the new commit message looks good.

buchdag:
@ntml that's perfect, thank you !

PR #1985, opened 2022-05-11
Make sure networks order is the same

SilverFire:
Whenever **docker-gen** catches a new event, it tries to generate a new **nginx.conf**, even when an event is unrelated to the proxied containers.

Since **DockerClient** does not guarantee the order in a `Container.Networks` list, each config re-generation may result in a different order of debug messages in the **upstream** section. 

For example, the first run will produce the following config:

```nginx
upstream web.example.com {
        ## Can be connected with "nginx-proxy-network" network
        # web_nginx_1
        server 172.16.1.10:80;
        # Cannot connect to network 'net2' of this container
        # Cannot connect to network 'net3' of this container
}
```

While some other time the order will be different:

```nginx
upstream web.example.com {
        # Cannot connect to network 'net2' of this container
        ## Can be connected with "nginx-proxy-network" network
        # web_nginx_1
        server 172.16.1.10:80;
        # Cannot connect to network 'net3' of this container
}
```

When there are hundreds of containers starting/stopping in a Docker, it leads to a lot of useless Nginx reloads.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -105,7 +105,7 @@ upstream {{ .Upstream }} {
             {{ end }}
         {{ end }}
 		{{ range $knownNetwork := $networks }}
-			{{ range $containerNetwork := $container.Networks }}
+			{{ range $containerNetwork := sortObjectsByKeysAsc $container.Networks "Name" }}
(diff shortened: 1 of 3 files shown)


buchdag:
Thanks for the improvement @SilverFire 👍

PR #2002, opened 2022-06-20
Headers more filter module

dextervip:
### Background

By default nginx expose its name in the response header as `server: nginx`. Knowing type of a running web server allows testers to determine known vulnerabilities and the appropriate exploits to use during testing. Server Banners Exposed may be considered a vulnerability with low severity. However depending of level of required compliance, It can a reason of compliance rejection. 

### Severity Evaluation
Based on the National Vulnerabilities Database (NDV), the vulnerability severity ratings of CVSS v3.0 can be summarized as showed below:
CVSSv3 Vector: [AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:N/A:N](https://nvd.nist.gov/vuln-metrics/cvss/v3-calculator?vector=AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:N/A:N)
CVSSv3 Base Score: 3.7
Severity: Low (0.1 - 3.9)

Reference
[CWE-200: Information Exposure ](https://cwe.mitre.org/data/definitions/200.html)
https://owasp.org/www-project-web-security-testing-guide/stable/4-Web_Application_Security_Testing/01-Information_Gathering/02-Fingerprint_Web_Server 
https://owasp.org/www-project-web-security-testing-guide/stable/4-Web_Application_Security_Testing/01-Information_Gathering/08-Fingerprint_Web_Application_Framework 
Enel IT Security Guidelines - APPLICATIONS (Guideline no. 7, Version no. 2, dated 30/09/2017): paragraph 2.6. Control #32: Coding, Subparagraph 2.6.1 – Baseline

### Remediation
Remove or modify the banners included into the responses in order to do not provide any information about technologies. Since `server_tokens off` is not enough to remove the banner `server: nginx`. Compiling headers more filter module and adding to the image is a choice to allow easily to hide server header or simply to remove it as the example below:

```
RUN { \
      echo 'server_tokens off;'; \
      echo 'more_clear_headers Server;'; \
    } > /etc/nginx/conf.d/server_banner.conf
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,36 @@ RUN git clone https://github.com/nginx-proxy/forego/ \
    && cd - \
    && rm -rf /go/forego
 
+
+# Build headers more module from scratch
+FROM nginx:1.21.6 AS headers-more-nginx-module
+
+# nginx:alpine contains NGINX_VERSION environment variable, like so:
+# ENV NGINX_VERSION 1.15.0
+ARG HEADERS_MORE_VERSION=v0.33
+
+RUN apt-get update \
+    && apt-get install -y \
+        build-essential \
+        libpcre++-dev \
+        zlib1g-dev \
+        libgeoip-dev \
+        wget \
+        git
+# Download sources
+RUN wget "http://nginx.org/download/nginx-${NGINX_VERSION}.tar.gz" -O nginx.tar.gz
+
+RUN cd /opt \
+    && git clone --depth 1 -b $HEADERS_MORE_VERSION --single-branch https://github.com/openresty/headers-more-nginx-module.git \
+    && cd /opt/headers-more-nginx-module \
+    && git submodule update --init \
+    && cd /opt \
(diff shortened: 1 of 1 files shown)


buchdag:
While I understand the need here, this kind of addition to the Dockerfile is very hard to maintain on the long run as it basically means we have to additionally maintain the build of the headers more filter module. What about the Alpine based image ?

Also there are no test to validate that this actually works.

buchdag:
This issue https://github.com/openresty/headers-more-nginx-module/issues/138 seems to confirm that compatibility issues between `ngx_headers_more` and nginx latest stable versions might happen from time to time. I don't think that adding a module in order to fix something that might be considered a low severity exploit is worth the risk of being blocked on nginx updates. If you are in a case where nginx exposing itself in the response header might be a reason for compliance rejection, vulnerabilities in the base nginx image not being patched will pretty certainly cause compliance rejection well before that.

PR #2025, opened 2022-07-26
Support vouch-proxy as authentication for exposed services

rvaidya:
This PR adds support for authenticating to services exposed by nginx-proxy using [vouch-proxy](https://github.com/vouch/vouch-proxy) and the nginx auth_request module.

Authentication is a common enough use case to merit adding support for this over requiring users to configure something like this using custom nginx configuration and shared volumes.

vouch-proxy supports a wide range of login providers for SSO.


buchdag:
Hi. Thanks for the PR, unfortunately I feel this is too specific to a single platform to justify merging, and as you noted yourself this is already easily doable with a [custom config](https://github.com/nginx-proxy/nginx-proxy#per-virtual_host).

PR #2036, opened 2022-08-15
docs: add note on vhost.d files with acme-companion

remiberthoz:
I added a small note on the Per-VIRTUAL_HOST nginx configuration files, about the portion of file that is appended by the acme-companion container. I think the behavior is relatively clear to anyone using the acme-challenges, but I found it surprising to not see it mentioned in the doc.


buchdag:
@remiberthoz this has since been disabled by default in acme-companion.

PR #2078, opened 2022-10-28
Support TCP and UDP proxy

KagurazakaNyaa:
Add including to the root of nginx.conf to facilitate adding configuration outside the http module.

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -54,6 +54,7 @@ RUN apk add --no-cache --virtual .run-deps \
 
 # Configure Nginx
 RUN echo "daemon off;" >> /etc/nginx/nginx.conf \
+   && echo "include /etc/nginx/toplevel.conf.d/*.conf;" >> /etc/nginx/nginx.conf \
    && sed -i 's/worker_processes  1/worker_processes  auto/' /etc/nginx/nginx.conf \
    && sed -i 's/worker_connections  1024/worker_connections  10240/' /etc/nginx/nginx.conf \
    && mkdir -p '/etc/nginx/dhparam'
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,7 @@ RUN apt-get update \
 
 # Configure Nginx
 RUN echo "daemon off;" >> /etc/nginx/nginx.conf \
+   && echo "include /etc/nginx/toplevel.conf.d/*.conf;" >> /etc/nginx/nginx.conf \
    && sed -i 's/worker_processes  1/worker_processes  auto/' /etc/nginx/nginx.conf \
    && sed -i 's/worker_connections  1024/worker_connections  10240/' /etc/nginx/nginx.conf \
    && mkdir -p '/etc/nginx/dhparam'
--- a/README.md
+++ b/README.md
@@ -365,6 +365,35 @@ You'll need apache2-utils on the machine where you plan to create the htpasswd f
 
 If you need to configure Nginx beyond what is possible using environment variables, you can provide custom configuration files on either a proxy-wide or per-`VIRTUAL_HOST` basis.
 
+#### TCP and UDP stream support
+
+If you want to proxy non-HTTP traffic, you can use nginx's stream module. Write a configuration file and mount it to `/etc/nginx/toplevel.conf.d`.
+
+```Nginx
+# stream.conf
+stream {
+
+    server {
+        listen     12345;
+        #TCP traffic will be forwarded to the "stream_backend" upstream group
+        proxy_pass stream_backend;
+    }
+
+    server {
+        listen     12346;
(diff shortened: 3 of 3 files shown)


p12tic:
@buchdag Are there any clear requirements (e.g. tests) which prevent this PR from landing? Or is it that the benefit of this feature is simply not clear and thus polishing the PR does not make sense? I could do whatever is necessary to get this feature merged. Thanks a lot!

buchdag:
@p12tic a test would be nice but the PR seems rather straightforward to me, the main issue is rather that it conflicts with the current code ... and that I completely failed to notice it. I'll try to fix the merge conflicts this week.

p12tic:
>  I'll try to fix the merge conflicts this week.

I can create a separate PR with merge conflicts fixed (the authorship of the code remains with [KagurazakaNyaa](https://github.com/KagurazakaNyaa) of course).

buchdag:
No need to, I just rebased the branch and fixed the conflict, could you review the PR and check that I did not botch anything ?

buchdag (approved):

Dockerfile.alpine
```diff
@@ -17,8 +17,10 @@ ENV NGINX_PROXY_VERSION=${NGINX_PROXY_VERSION} \
 RUN apk add --no-cache --virtual .run-deps bash openssl
 
 # Configure Nginx
-RUN sed -i 's/worker_connections.*;$/worker_connections   10240;/' /etc/nginx/nginx.conf \
+RUN echo -e "\ninclude /etc/nginx/toplevel.conf.d/*.conf;" >> /etc/nginx/nginx.conf \
```
p12tic, line 20: Dockerfile.debian does not contain `-e` on echo. Not sure if we need it here - trailing newline is good thing to have (but maybe I don't know what you do).

p12tic:
Thanks. I will add a test in a subsequent PR.

Dockerfile.alpine
KagurazakaNyaa: The `\n` is added here to prevent problems that may occur if the default configuration file of nginx does not contain a newline at the end in future updates.

buchdag: @p12tic `echo` does not behave the same on Debian and Alpine, the latter need the `- e` flag for `\n` to be converted to a newline.

I'm not sure either command result in nginx.conf ending with a newline if that's the desired outcome, the `\n` at the start is rather to insert a blank line between our statement and the end of the existing `http` block.

KagurazakaNyaa: @buchdag In fact, in my tests, Debian's echo still needs `-e` to ensure that `\n` is escaped correctly

```bash
$ docker run --rm -it nginx:1.26.0 bash
root@e0184964fc67:/# echo "\ninclude /etc/nginx/toplevel.conf.d/*.conf;"
\ninclude /etc/nginx/toplevel.conf.d/*.conf;
root@e0184964fc67:/# echo -e "\ninclude /etc/nginx/toplevel.conf.d/*.conf;"

include /etc/nginx/toplevel.conf.d/*.conf;
root@e0184964fc67:/# 
exit

```

```bash
$ docker run --rm -it nginx:1.26.0-alpine sh  
/ # echo "\ninclude /etc/nginx/toplevel.conf.d/*.conf;"
\ninclude /etc/nginx/toplevel.conf.d/*.conf;
/ # echo -e "\ninclude /etc/nginx/toplevel.conf.d/*.conf;"

include /etc/nginx/toplevel.conf.d/*.conf;
/ # 

```

buchdag: @KagurazakaNyaa strangely that's what I initially committed (using the `-e` flag on both version) and it resulted on `/n` being printed to the nginx.conf and [all tests failing for the Debian version](https://github.com/nginx-proxy/nginx-proxy/actions/runs/8969610890).

KagurazakaNyaa: @buchdag I know why this strange problem occurs. During `docker build`, `/bin/sh` is used instead of `/bin/bash`, and their `echo` behaviors are different.

```sh
$ docker run --rm -it nginx:1.26.0 sh
# echo "\ninclude /etc/nginx/toplevel.conf.d/*.conf;"

include /etc/nginx/toplevel.conf.d/*.conf;
# echo -e "\ninclude /etc/nginx/toplevel.conf.d/*.conf;"
-e 
include /etc/nginx/toplevel.conf.d/*.conf;
# 
```

KagurazakaNyaa: So using `echo` without `-e` in debian is correct

buchdag: Is it okay for you if I merge the PR then ?

KagurazakaNyaa: LGTM, if there are no more issues, I think it is ready to merge.

PR #2080, opened 2022-11-02
feat: Providing env vars for dockergen and nginx

hypery2k:
Add possibility to tweak docker-gen and nginx via env vars.


buchdag:
Hi, thanks for the PR. The preferred and already supported way to do this is to use [separate docker-gen and nginx container](https://github.com/nginx-proxy/nginx-proxy#separate-containers), which can be tweaked at will.

PR #2091, opened 2022-11-08
Load Network Config depending on NETWORK_ACCESS env var

solrac200:
Instead of just checking if NETWORK_ACCESS is internal and then using the networks_internal config file, use the value of this env var and load the according config file.


buchdag:
Hi, thanks for the PR however it has no test, no documentation and break backward compatibility for people relying on the use of `network_internal.conf`. It won't be merged if those issues aren't addressed.

bitcoinfinancier (approved):

bitcoinfinancier (approved):

PR #2127, opened 2023-01-01
Various code health chores

rhansen:
No behavior changes, just cleanups.


buchdag (changes requested):
LGTM except for that last commit, I don't want to spark an endless debate but I find two space tabs make code harder to read, especially on GitHub.

rhansen:
> OK, I'll do 4 spaces.

Done.  I also pushed a new commit that re-wraps long template comments to keep those lines under 80 columns long.  (Unfortunately golang templates do not support line continuations inside `{{...}}` except for comments.)

buchdag (approved):

PR #2141, opened 2023-01-18
feat: Unconditionally produce debug comments

### LINKED ISSUE #2139: $DEBUG environment variable
(version: `nginx-proxy:1.1`)

Hey. One of my containers (specifically [wg-easy](https://github.com/WeeJeWel/wg-easy/blob/bbd19b7a6f390d95d80aab75271a24b92f6c591e/Dockerfile#L46)) uses the `$DEBUG` variable from inside and nginx-proxy picks it up [here](https://github.com/nginx-proxy/nginx-proxy/blob/5d22134ee6d37f5f5712aadde6ee2eda85e25956/nginx.tmpl#L113) and spams the entire log with errors while not generating the config. Stopping wg-easy helps, but then I lose access to my home so that's not a good solution. Also, the variable should be named differently, like `NGINX_PROXY_DEBUG`, because other apps may just read it and trigger some internal debugging too.

Thanks

Edit: sorry if this is a duplicate, but the word "DEBUG" doesn't help much in the github's search engine

rhansen:
Rationale for eliminating the check to see if the `DEBUG` environment variable holds a true value:
  * The `DEBUG` environment variable might be set on a container (for purposes specific to that container, not `nginx-proxy`) to a value that cannot be parsed as a bool, which would break `nginx-proxy`.
  * It simplifies the template.
  * It eliminates a cold code path.
  * It avoids heisenbugs.
  * It makes debugging easier for users.

Also delete the debug info tests, as they are fragile (e.g., they break if IPv6 is enabled) and they provide limited value.

Alternatively, we could avoid collision with the container's use of the `DEBUG` environment variable by using a [container label](https://docs.docker.com/engine/reference/commandline/run/#-set-metadata-on-container--l---label---label-file) such as `com.google.nginx-proxy.nginx-proxy.debug`.  I think doing so has dubious value, especially if we want to attempt backwards compatibility with the `DEBUG` environment variable.

Fixes #2139

cc @GGORG0


README.md
```diff
@@ -490,12 +490,13 @@ Please note that using regular expressions in `VIRTUAL_HOST` will always result
 
 ### Troubleshooting
 
-In case you can't access your VIRTUAL_HOST, set `DEBUG=true` in the client container's environment and have a look at the generated nginx configuration file `/etc/nginx/conf.d/default.conf`:
+If you can't access your `VIRTUAL_HOST`, inspect the generated `/etc/nginx/conf.d/default.conf` file:
 
 ```console
 docker exec <nginx-proxy-instance> cat /etc/nginx/conf.d/default.conf
 ```
```
buchdag, line 497: While we're at it, we should suggest to look a the full configuration and use `nginx -T` to get it rather than doing `cat` on a single file. That's what I end up asking most of the time on issues.
```suggestion
If you can't access your `VIRTUAL_HOST`, inspect the generated nginx configuration:

```console
docker exec <nginx-proxy-instance> nginx -T
```
```

buchdag:
@rhansen I completely agree on removing the per proxied container `DEBUG` variable, but I'm not certain about removing the global one. Should we not at least provide a way to have a debug free nginx config ?

rhansen:
> Should we not at least provide a way to have a debug free nginx config ?

No, I don't think it's necessary.  The extra comments are harmless.

rhansen:
@GGORG0 Does this PR solve your problem?

buchdag (approved):

buchdag:
> The extra comments are harmless.

I had more in mind users that would like a "clean" config generated, in the same way as you don't always want debug informations on your application logs. I leave that up to you, this can be merged as is so that @GGORG0 can test it more easily.

Issue #886, opened 2017-07-26
Allow upstream options


italomaia:
In some scenarios, upstream options are needed, like when you have to always route the same user to the same host, like socketio (with ip_hash).

Issue #871, opened 2017-07-04
USE_IP_HASH seems to have zero impact to the default.conf


redskate:
Hi,

My docker-compose definition for your nginx is:

```
nginx:
    image: xxx-nginx
    restart: always
    environment:
      DEFAULT_HOST: xxx.local #mandatory to load balance
      USE_IP_HASH: 1
    ports:
      - 90:80
    volumes:
      - "/var/run/docker.sock:/tmp/docker.sock:ro"
      - "./nginxlogs:/var/log/nginx"
    links:
      - xxx_tomcat_node_1
      - xxx_tomcat_node_2
```

When I bash log in into the corrensponding nginx container, and look into `/etc/nginx/conf.d/default.conf`, I do not find ip_hash" in it - why?

redskate:
Hi again - probably the docker-compose config should have also a `VIRTUAL_HOST: xxx.local` to trigger adding "ip_hash;" to nginx.tmpl.

HOWEVER I do have some general doubts, nginx (free) is a usable idea to load-balance in an upstream of sticky sessions bounded servers.

`ip_hash;` seems to be a quick-trick to allow something closed to session stickyness, but is defacto more an "IP balancing" than a load balancing. It does not balance the load at all!!!!

So it seems nginx be a "door" to spend some time in working with it, then opening to some non-free real systems - for cash - which ensure you can still have round-robin with session stickyness.

For this reason I will now distanciate myself from nginx!

Nice integration work anyway 👍

Pimmetje:
USE_IP_HASH is something i can only find in a pull request that is still open. Did i miss something?

JamBalaya56562:
`USE_IP_HASH` was only ever part of an unmerged PR (as @Pimmetje noted above), so it never had any effect in the released image.

To choose a non-default upstream load-balancing method such as `ip_hash` (e.g. for sticky behaviour across multiple replicas), use the `com.github.nginx-proxy.nginx-proxy.loadbalance` label on your container — it injects the directive (`ip_hash;`, `hash $remote_addr;`, etc.) into the generated `upstream{}` block. It's documented under "Upstream Server HTTP Load Balancing Support" (including the `$`-escaping note) and covered by the `test_loadbalancing` tests.

PR #299, opened 2015-11-21
Added mechanism to specify ip_hash in the upstream directive

tpcwang:
This should address https://github.com/jwilder/nginx-proxy/issues/221

Issue #221, opened 2015-08-20
session affinity


timhuff:
Does this proxy enable nginx session affinity? From looking at the `nginx.tmpl` file, it would appear not. It seems that this would be beneficial to include. When I release a new version of our website, I spin up containers with the new version and then stop containers with the old version. During this process, it's possible for a user to grab, for instance, the new javascript file but the old css file. By enabling session affinity, it would ensure that users that started out on the old version stay on the old version until the switch is complete. Is there anything that makes this possibility unrealistic?

md5:
The fact that it involves the proxy setting a cookie would mean that the proxy would no longer be transparent, which I believe is one of @jwilder's goals for this image. I could see this being an optional behavior, but not a default.

See https://github.com/jwilder/nginx-proxy/issues/153 for some previous discussion of sticky sessions.

DavidSichau:
It would be possible without any Sessions via the `ip_hash;` See http://nginx.org/en/docs/http/ngx_http_upstream_module.html#ip_hash. This would not add any state to the container and still allow sticky sessions.

redskate:
Hi, THANK YOU for having defined the possibility to use "ip_hash;"

In docker-compose, I am using `USE_IP_HASH: 1 ` in the environment section of your nginx, but this seems to have zero impact ...

PR #2173, opened 2023-02-14
feat: Add support for HTTP load balancing between the proxy and upstream server groups

SchoNie:
Currently all connections are forwarded upstream with Round Robin algorithm. (default Nginx)
This pull request adds a container labels to specify the load balancing algorithm. Which can be used to have session persistence / affinity.
Some web applications which use the HTTP long-polling transport like websocket require sticky sessions.

Related #871, #299, #221, #886

Note if #1934 is merged: When using load balancing methods other than the default round-robin method, it is necessary to activate them before the keepalive directive.

If you think this PR is usefull I also plan to add the 'weight' directive.

I am trying to write a test. I have no pytest experience and having a little hard time to setup the test suite.
```
def test_web_loadbalance(docker_compose, nginxproxy):
    conf = nginxproxy.get_conf().decode('ASCII')
    r1 = nginxproxy.get("http://web.nginx-proxy.tld")
    r2 = nginxproxy.get("http://web.nginx-proxy.tld")
    assert r1.status_code == 200
    assert r2.text == r1.text
```
If I am correct the test webserver should output the container hostname.
But this always passes. In my first test when still using round robin I would expect this to fail.
Could someone point me in the right direction?

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,6 +373,42 @@ docker run -d -p 80:80 -p 443:443 \
 
 You'll need apache2-utils on the machine where you plan to create the htpasswd file. Follow these [instructions](http://httpd.apache.org/docs/2.2/programs/htpasswd.html)
 
+### Upstream (Backend) Server HTTP Load Balancing Support
+
+> **Warning**
+> This feature is experimental.  The behavior may change (or the feature may be removed entirely) without warning in a future release, even if the release is not a new major version.  If you use this feature, or if you would like to use this feature but you require changes to it first, please [provide feedback in #2195](https://github.com/nginx-proxy/nginx-proxy/discussions/2195).  Once we have collected enough feedback we will promote this feature to officially supported.
+
+If you have multiple containers with the same `VIRTUAL_HOST` and `VIRTUAL_PATH` settings, nginx will spread the load across all of them.  To change the load balancing algorithm from nginx's default (round-robin), set the `com.github.nginx-proxy.nginx-proxy.loadbalance` label on one or more of your application containers to the desired load balancing directive.  See the [`ngx_http_upstream_module` documentation](https://nginx.org/en/docs/http/ngx_http_upstream_module.html) for available directives.
+
+> **Note**
+> * Don't forget the terminating semicolon (`;`).
+> * If you are using Docker Compose, remember to escape any dollar sign (`$`) characters (`$` becomes `$$`).
+
+Docker Compose example:
+
+```yaml
+services:
+  nginx-proxy:
+    image: nginxproxy/nginx-proxy
+    ports:
+      - "80:80"
+    volumes:
+      - /var/run/docker.sock:/tmp/docker.sock:ro
+    environment:
+      HTTPS_METHOD: nohttps
+  myapp:
+    image: jwilder/whoami
+    expose:
+      - "8000"
+    environment:
+      VIRTUAL_HOST: myapp.example
+      VIRTUAL_PORT: "8000"
+    labels:
+      com.github.nginx-proxy.nginx-proxy.loadbalance: "hash $$remote_addr;"
+    deploy:
+      replicas: 4
+```
+
 ### Headers
 
 By default, `nginx-proxy` forwards all incoming request headers from the client to the backend server unmodified, with the following exceptions:
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -214,6 +214,11 @@
 {{- define "upstream" }}
 upstream {{ .Upstream }} {
     {{- $server_found := false }}
+    {{- $loadbalance := first (keys (groupByLabel .Containers "com.github.nginx-proxy.nginx-proxy.loadbalance")) }}
+    {{- if $loadbalance }}
+    # From the container's loadbalance label:
+    {{ $loadbalance }}
+    {{- end }}
     {{- range $container := .Containers }}
     # Container: {{ $container.Name }}
         {{- $args := dict "globals" $.globals "container" $container }}
--- a/test/test_loadbalancing.yml
+++ b/test/test_loadbalancing.yml
@@ -1,0 +1,27 @@
+services:
+  loadbalance-hash:
+    image: web
+    expose:
+      - "81"
+    environment:
+      WEB_PORTS: 81
+      VIRTUAL_HOST: loadbalance-enabled.nginx-proxy.tld
+    labels:
+      com.github.nginx-proxy.nginx-proxy.loadbalance: "hash $$remote_addr;"
+    deploy:
+      replicas: 2
+
+  loadbalance-roundrobin:
+    image: web
+    expose:
+      - "82"
+    environment:
+      WEB_PORTS: 82
+      VIRTUAL_HOST: loadbalance-disabled.nginx-proxy.tld
+    deploy:
+      replicas: 2
+
+  sut:
+    image: nginxproxy/nginx-proxy:test
+    volumes:
+      - /var/run/docker.sock:/tmp/docker.sock:ro
--- a/test/test_loadbalancing.py
+++ b/test/test_loadbalancing.py
@@ -1,0 +1,16 @@
+import pytest
+import re
+
+def test_loadbalance_hash(docker_compose, nginxproxy):
+    conf = nginxproxy.get_conf().decode('ASCII')
+    r1 = nginxproxy.get("http://loadbalance-enabled.nginx-proxy.tld")
+    r2 = nginxproxy.get("http://loadbalance-enabled.nginx-proxy.tld")
+    assert re.search(r"hash \$remote_addr\;", conf)
+    assert r1.status_code == 200
+    assert r2.text == r1.text
+
+def test_loadbalance_roundrobin(docker_compose, nginxproxy):
+    r1 = nginxproxy.get("http://loadbalance-disabled.nginx-proxy.tld")
+    r2 = nginxproxy.get("http://loadbalance-disabled.nginx-proxy.tld")
+    assert r1.status_code == 200
+    assert r2.text != r1.text


rhansen:
> If you think this PR is useful

I do!  Thank you for the PR!

> If I am correct the test webserver should output the container hostname.

Yes.

> But this always passes. In my first test when still using round robin I would expect this to fail.

I too would expect that test to fail.  Some questions:

  * Could this be caused by `Connection: keep-alive` from the test client to nginx?
      * Does nginx use the same upstream server to handle all requests transmitted in the same persistent connection?
      * Does the HTTP client in the `nginxproxy` fixture use persistent connections?
  * What does the associated `.yml` file contain?
  * What do you get when you manually query? Like this:
    ```shell
    docker-compose -f test/test_load_balancing.yml up -d
    ip=$(docker container inspect -f '{{.NetworkSettings.Networks.bridge.IPAddress}}' test_sut_1)
    curl --connect-to ::${ip}: http://web.nginx-proxy.tld/
    ```
  * Does your nginx-proxy container have any extra config files that enable content caching?  (run `docker exec test_sut_1 nginx -T` to see nginx's full config)

rhansen:
Given that the user provides the full line including the semicolon, maybe it would be better to just add an `include` directive if a certain file exists.  For example, look for `/etc/nginx/upstream.d/${upstream_name}.conf` first, and if that doesn't exist, look for `/etc/nginx/upstream.d/default.conf`.  If either exists, that file will be included.  Then users can put the load balancing directive in the include file.  What do you think?

SchoNie:
Thank you Richard! Your hints pointed me in the right direction and I had a small error in my compose file.
Added some initial tests. All feedback is appreciated.

>  

> Given that the user provides the full line including the semicolon, maybe it would be better to just add an `include` directive if a certain file exists. For example, look for `/etc/nginx/upstream.d/${upstream_name}.conf` first, and if that doesn't exist, look for `/etc/nginx/upstream.d/default.conf`. If either exists, that file will be included. Then users can put the load balancing directive in the include file.
> 
> Two annoying things about my suggestion:
> 
> * The user has to be able to determine the upstream name that nginx-proxy generates, which I plan on changing ([feat: Use container names for upstream names to avoid repeats #1938](https://github.com/nginx-proxy/nginx-proxy/pull/1938)).
> * The user has to create a separate file and bind mount it into the container, rather than simply define a label.
> 
> Maybe a compromise would be to rename the label in this PR from `*.loadbalance` to something like `*.upstream_directives` and adjust the documentation to say that any directives suitable in an `upstream` block can be placed here, such as load balancing directives. One problem with this: what if multiple containers in the same upstream set this label? This PR picks an arbitrary winner; if the label is renamed, maybe it would be better to concatenate the values. There are usability downsides either way.

I was at first thinking to make the loadbalancing a true/false option. But then because of all flexible loadbalance options I decided to keep it clean and just insert the directive as-is from the label to prevent from a lot if/else stuff.
And focused more on writing the documentation with good examples.

I do like your suggestions. They gave me new ideas. My thoughts:
Including a upstream.d file is smart, but indeed your predicable name PR needs to be in there first. And maybe not so easy for the user to find out.
More and more apps use websockets and now we can provide a documented and unit-tested approach to handle this.

Renaming to `*.upstream_directives` is a possibility but for me it sounds a little to global. I don't see much more options in the [upstream module](http://nginx.org/en/docs/http/ngx_http_upstream_module.html) an average user would use. (The most interesting are related to the keepalive timeouts) and like you say can be dangerous if set different on multiple containers.

I rather would expand this PR to include the server weight and make a separate PR to include a upstream.d file. Which we can benefit from in more complex scenarios.

README.md
```diff
@@ -373,6 +373,50 @@ docker run -d -p 80:80 -p 443:443 \
 
 You'll need apache2-utils on the machine where you plan to create the htpasswd file. Follow these [instructions](http://httpd.apache.org/docs/2.2/programs/htpasswd.html)
 
+### Upstream (Backend) Server HTTP Load Balancing Support
+
+> **Warning**
+> This feature is experimental.  The behavior may change (or the feature may be removed) without warning in a future release, even if the release is not a new major version.  If you use this feature, please provide feedback so we know whether it is working properly and can be promoted to officially supported.
+To enable HTTP load balancing between `nginx-proxy` and a group of backend servers, set the `com.github.nginx-proxy.nginx-proxy.loadbalance` label on the server's container to the load balancing algorithm which should be used. See the [nginx load balancing documentation](https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/) and the [Docker label documentation](https://docs.docker.com/config/labels-custom-metadata/) for details.
+The load balancing algorithm is enabled for the containers sharing the same `VIRTUAL_HOST`
+#### Choosing a Load-Balancing Method
```
rhansen, line 382: I don't think it's necessary to go into detail about choosing a load balancing method.  I think it is sufficient to just link to the nginx documentation and explain how to set up load balancing in nginx-proxy.  Something like this:

> ### Upstream (Backend) Server HTTP Load Balancing Support
>
>> **Warning**
>> *(experimental warning goes here)*
>
> If you have multiple containers with the same `VIRTUAL_HOST` and `VIRTUAL_PATH` settings, nginx will spread the load across all of them.  To change the load balancing algorithm from nginx's default (round-robin), set the `com.github.nginx-proxy.nginx-proxy.loadbalance` label on one or more of your application containers to the desired load balancing directive.  See the [`ngx_http_upstream_module` documentation](https://nginx.org/en/docs/http/ngx_http_upstream_module.html) for available directives.
>
>> **Note**
>>   * Don't forget the terminating semicolon (`;`).
>>   * If you are using Docker Compose, remember to escape any dollar sign (`$`) characters (`$` becomes `$$`).
>
> Docker Compose example:
>
> ```yaml
> services:
>   nginx-proxy:
>     image: nginxproxy/nginx-proxy
>     ports:
>       - "80:80"
>     volumes:
>       - /var/run/docker.sock:/tmp/docker.sock:ro
>     environment:
>       HTTPS_METHOD: nohttps
>   myapp:
>     image: jwilder/whoami
>     expose:
>       - "8000"
>     environment:
>       VIRTUAL_HOST: myapp.example
>       VIRTUAL_PORT: "8000"
>     labels:
>       com.github.nginx-proxy.nginx-proxy.loadbalance: "hash $$remote_addr;"
>     deploy:
>       replicas: 4
> ```

nginx.tmpl
```diff
@@ -198,6 +198,11 @@
 {{- define "upstream" }}
 upstream {{ .Upstream }} {
     {{- $server_found := false }}
+    {{- $loadbalance := first (keys (groupByLabel .Containers "com.github.nginx-proxy.nginx-proxy.loadbalance")) }}
+    {{- if $loadbalance }}
+    # Loadbalancing method: {{ $loadbalance }}
```
rhansen, line 203: I don't think it's useful to repeat the value of the `$loadbalance` variable here.

```suggestion
    # From the container's loadbalance label:
```

test/test_loadbalancing.yml
```diff
@@ -0,0 +1,38 @@
+  loadbalance-hash-1:
```
rhansen, line 1: Missing `services:` line?

```suggestion
services:
  loadbalance-hash-1:
```

test/test_loadbalancing.yml
```diff
@@ -0,0 +1,38 @@
...
+    labels:
+      com.github.nginx-proxy.nginx-proxy.loadbalance: "hash $$remote_addr;"
+
+  loadbalance-hash-2:
+    image: web
+    expose:
+      - "82"
+    environment:
+      WEB_PORTS: 82
+      VIRTUAL_HOST: loadbalance-enabled.nginx-proxy.tld
```
rhansen, line 17: Rather than repeat the container, use `deploy.replicas`:

```suggestion
    deploy:
      replicas: 2
```

rhansen (changes requested):

test/test_loadbalancing.yml
SchoNie: It's missing in more tests. For example: [test_keepalive.yml](https://github.com/rhansen/nginx-proxy/blob/keepalive/test/test_keepalive.yml) and [test_log_format.yml](https://github.com/nginx-proxy/nginx-proxy/blob/main/test/test_log_format.yml) and more.
So I was not sure what is prefered.

rhansen: I should fix those.  Omitting it means use the old v1 semantics, which doesn't matter for those tests, but they should be updated anyway for consistency.  It is needed here in order to use `deploy.replicas`.

rhansen (approved):
LGTM.  What do you think @buchdag?

We should remember to squash the commits when merging (or you can squash them yourself @SchoNie and force push before we merge).

README.md
```diff
@@ -373,6 +373,42 @@ docker run -d -p 80:80 -p 443:443 \
 
 You'll need apache2-utils on the machine where you plan to create the htpasswd file. Follow these [instructions](http://httpd.apache.org/docs/2.2/programs/htpasswd.html)
 
+### Upstream (Backend) Server HTTP Load Balancing Support
+
+> **Warning**
+> This feature is experimental.  The behavior may change (or the feature may be removed) without warning in a future release, even if the release is not a new major version.  If you use this feature, please provide feedback so we know whether it is working properly and can be promoted to officially supported.
```
buchdag, line 379: I'll have the same request as on https://github.com/nginx-proxy/nginx-proxy/pull/1934 : please open a discussion for feedbacks on this feature first, then include a link to it in the documentation.

buchdag (changes requested):

buchdag:
@rhansen apart from the request change related to the feedback discussion and it inclusion in the docs, this LGTM (and yes we should squash the commits).

README.md
SchoNie: Done. #2195 
And also squashed.

rhansen (dismissed):

buchdag (approved):

PR #2183, opened 2023-02-20
fix: Sort networks and ports before iterating

### LINKED ISSUE #2182: order of exposed ports is not sorted
The order of exposed ports in the comment on line 84 in the `nginx.tmpl` is not sorted.

```
{{- /* If only 1 port exposed, use that as a default, else 80. */}}
#     exposed ports:{{ range $.container.Addresses }} {{ .Port }}/{{ .Proto }}{{ else }} (none){{ end }}
```

This leads to different `/etc/nginx/conf.d/default.conf` on each run of `docker-gen` and therefor to a restart of the `nginx`.

https://github.com/nginx-proxy/nginx-proxy/blob/main/nginx.tmpl#L83-L84

I've copied the generated conf every 5 seconds while having some GitLab runner containers started and stopped.

```
/srv/nginx-proxy$ for i in {1..60}; do docker compose cp dockergen:/etc/nginx/conf.d/default.conf conf-$i.conf ; sleep 5; done
```

Comparing the generated configs exposed the issue:
```/srv/nginx-proxy$ diff conf-1.conf conf-6.conf
432c432
<     #     exposed ports: 9252/tcp 8093/tcp
---
>     #     exposed ports: 8093/tcp 9252/tcp
485c485
<     #     exposed ports: 22/tcp 443/tcp 5000/tcp 80/tcp
---
>     #     exposed ports: 5000/tcp 80/tcp 22/tcp 443/tcp
```

The same has been in previous versions with the networks https://github.com/nginx-proxy/docker-gen/issues/431

JanMalte:
Please review and test, as I do not have any knowledge of the used template language

This should fix #2182 and the old issue of `docker-gen` (https://github.com/nginx-proxy/docker-gen/issues/431)

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -21,7 +21,7 @@
 {{- $_ := set $globals "ssl_policy" (or ($globals.Env.SSL_POLICY) "Mozilla-Intermediate") }}
 {{- $_ := set $globals "networks" (dict) }}
 # networks available to nginx-proxy:
-{{- range $globals.CurrentContainer.Networks }}
+{{- range sortObjectsByKeysAsc $globals.CurrentContainer.Networks "Name" }}
     {{- $_ := set $globals.networks .Name . }}
 #     {{ .Name }}
 {{- end }}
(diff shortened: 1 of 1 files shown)


rhansen (approved):
Thank you!

I fixed a small mistake and changed the commit message to indicate that this is a fix (nginx should not be restarted due to comment churn).

Issue #1537, opened 2020-12-01
Bad Gateway --net=host


Hamsterman:
I am trying to use the reverseproxy in "-net host" mode but it seems like I keep getting a "Bad gateway".
Is this a bug, or what am I doing wrong?

I first reported it at homebridge, but they suggested it was a ReverseProxy issue.
https://github.com/homebridge/homebridge/issues/2731

Hamsterman:
I have now tried this with another container - the exact same issue.

docker run -d --name ReverseProxy -p 80:80 -p 443:443 --net=host -v //etc/DockerStorage/ReverseProxy:/etc/nginx/certs -v /var/run/docker.sock:/tmp/docker.sock:ro --restart=unless-stopped jwilder/nginx-proxy

docker run -d --name HomeAssistant -e VIRTUAL_HOST=DOMAIN.COM -e VIRTUAL_PORT=8123 --net=host --expose=8123 -v "//etc/DockerStorage/HomeAssistant:/config" -v /etc/localtime:/etc/localtime:ro --restart=unless-stopped homeassistant/home-assistant:stable

Hamsterman:
Anyone ?

pmow:
Same.  Hosts are marked as being offline unless in bridge mode.

DatAres37:
> docker run -d --name HomeAssistant -e VIRTUAL_HOST=DOMAIN.COM -e VIRTUAL_PORT=8123 --net=host --expose=8123 -v "//etc/DockerStorage/HomeAssistant:/config" -v /etc/localtime:/etc/localtime:ro --restart=unless-stopped homeassistant/home-assistant:stable

I'm not a docker expert, but I don't think putting both containers on the host network works, since the expose option probably doesn't work on the host (same as publishing ports).

Typepluto:
Maybe you should try port forwarding.
In this case:
docker run -d --name HomeAssistant -e VIRTUAL_HOST=DOMAIN.COM -e VIRTUAL_PORT=8123 --net=host --expose=8123 -v "//etc/DockerStorage/HomeAssistant:/config" -v /etc/localtime:/etc/localtime:ro --restart=unless-stopped homeassistant/home-assistant:stable

> I have now tried this with another container - the exact same issue.
> 
[...]

PR #2222, opened 2023-04-19
Support containers running --net=host (#1537)

### LINKED ISSUE #832: Manually setting upstream
I have a docker container with --net=host as it is a service that sends magic packets for Wake on LAN, and I can't (as far as I understand) from within the docker network.

However, when starting my container, it is generated with an empty upstream block, causing my nginx instance to cease working.

Manually adding
    server 192.168.1.228:10440;

Resolved the issue, but it is obviously not a good soluton as it will be overwritten eventually.

Am I doing something wrong here, is it not supported, or is there any way around it?

Container is run with:

    docker run
         --name wol
         --net=host
         -e "VIRTUAL_HOST=wol.example.com"
         -e "VIRTUAL_PORT=10440"
         -e "LETSENCRYPT_HOST=wol.example.com"
         -e "LETSENCRYPT_EMAIL=someone@example.com"

### LINKED ISSUE #1270: VIRTUAL_PORT is not filled in default.conf when container uses host networking
## Description

I have a container with this config

```yaml
version: '3.6'
services:
  jellyfin:
    image: jellyfin
    network_mode: 'host'
    volumes:
      - ./config:/config
      - ./cache:/cache
      - ./netdisk:/media
    restart: always
    environment:
      - VIRTUAL_HOST=abc.example.com
      - VIRTUAL_IP=127.0.0.1
      - VIRTUAL_PORT=8096
      - SSL_POLICY=Mozilla-Modern
      - HTTPS_METHOD=noredirect
    privileged: true
```

It uses host networking, so I provide `VIRTUAL_IP` to specify the exact upstream address.

However, this service container cannot be reached via [https://abc.example.com](https://abc.example.com). It returns 502 Bad Gateway.

The log of `nginx-proxy` shows

```
proxy_1  | nginx.1    | abc.example.com 10.246.8.238 - - [01/May/2019:03:07:25 +0000] "GET /favicon.ico HTTP/2.0" 502 575 "https://abc.example.com/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/74.0.3729.108 Safari/537.36"
proxy_1  | nginx.1    | 2019/05/01 03:07:25 [error] 2627#2627: *15370 no live upstreams while connecting to upstream, client: 10.246.8.238, server: abc.example.com, request: "GET / HTTP/2.0", upstream: "http://abc.example.com/", host: "abc.example.com"
```

I think maybe there is something wrong with the generated `default.conf`. So I got inside of `nginx-proxy` container and saw the relevant section in `/etc/nginx/conf.d/default.conf` is:

```
upstream abc.example.com {                                                                                                                                
                                ## Can be connected with "host" network                                                                                                                                                           
                # jellyfindoc_jellyfin_1                    
                        server 127.0.0.1 down;                                                     
}
```

The upstream server does not fill in the value of `VIRTUAL_PORT` as I provided.

I changed it to `server 127.0.0.1:8086 down;` and performed `nginx -s reload`, I still cannot reach the service container via [https://abc.example.com](https://abc.example.com).

Then I changed it to `server 127.0.0.1:8086;` and performed `nginx -s reload`, this issue was solved. My service container now publically reachable via [https://abc.example.com](https://abc.example.com).

So this is an issue with upstream auto-generation.

The expected behavior is to produce a config similar to

```
upstream VIRTUAL_HOST {                                                                                                                                
                                ## Can be connected with "host" network                                                                                                                                                           
                # jellyfindoc_jellyfin_1                    
                        server VIRTUAL_IP:VIRTUAL_PORT;                                                     
}
```

when the user's container uses host networking.

## Background Info

Configuration of `nginx-proxy` service

```yaml
version: '3.6'
services:
  proxy:
    image: jwilder/nginx-proxy:alpine
    restart: always
    network_mode: host
    volumes:
      - ./custom.conf:/etc/nginx/conf.d/custom.conf
      - ./ssl/certs:/etc/nginx/certs/:ro
```

content of custom.conf

```
client_max_body_size 0;
proxy_request_buffering off;
proxy_read_timeout  36000s;
proxy_send_timeout  36000s;
```

rot169:
Containers which run host networking do not have their own IP address, which means the template doesn't know what to use as the upstream server. This change detects if a target container is running host networking, and if so, uses the IP address of the bridge net gateway as the upstream target.

Fixes #1270 
Fixes #832

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ In this example, the `my-nginx-proxy` container will be connected to `my-network
 
 `nginx-proxy` is compatible with containers using Docker's [host networking](https://docs.docker.com/network/host/), both with the proxy connected to one or more [bridge network](https://docs.docker.com/network/bridge/) (default or user created) or running in host network mode itself.
 
-Proxyed containers running in host network mode **must** use the [`VIRTUAL_PORT](#virtual-ports) environment variable, as this is the only way for `nginx-proxy` to get the correct port (or a port at all) for those containers running.
+Proxyed containers running in host network mode **must** use the [`VIRTUAL_PORT](#virtual-ports) environment variable, as this is the only way for `nginx-proxy` to get the correct port (or a port at all) for those containers.
 
 ### Custom external HTTP/HTTPS ports
 
--- a/README.md
+++ b/README.md
@@ -182,6 +182,12 @@ docker network connect my-other-network my-nginx-proxy
 
 In this example, the `my-nginx-proxy` container will be connected to `my-network` and `my-other-network` and will be able to proxy to other containers attached to those networks.
 
+### Host networking
+
+`nginx-proxy` is compatible with containers using Docker's [host networking](https://docs.docker.com/network/host/), both with the proxy connected to one or more [bridge network](https://docs.docker.com/network/bridge/) (default or user created) or running in host network mode itself.
+
+Proxyed containers running in host network mode **must** use the [`VIRTUAL_PORT](#virtual-ports) environment variable, as this is the only way for `nginx-proxy` to get the correct port (or a port at all) for those containers running.
+
 ### Custom external HTTP/HTTPS ports
 
 If you want to use `nginx-proxy` with different external ports that the default ones of `80` for `HTTP` traffic and `443` for `HTTPS` traffic, you'll have to use the environment variable(s) `HTTP_PORT` and/or `HTTPS_PORT` in addition to the changes to the Docker port mapping. If you change the `HTTPS` port, the redirect for `HTTPS` traffic will also be configured to redirect to the custom port. Typical usage, here with the custom ports `1080` and `10443`:
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -67,6 +67,11 @@
     #         {{ .Name }} (ignored)
             {{- continue }}
         {{- end }}
+        {{- if eq .Name "host" }}
+    #         {{ .Name }} (host network; using {{ (index $.globals.CurrentContainer.Networks 0).Name }} gateway)
+            {{- $ip = (index $.globals.CurrentContainer.Networks 0).Gateway }}
+            {{- continue }}
+        {{- end }}
         {{- if and (not (index $.globals.networks .Name)) (not $.globals.networks.host) }}
     #         {{ .Name }} (unreachable)
             {{- continue }}
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -68,6 +68,12 @@
             {{- continue }}
         {{- end }}
         {{- if eq .Name "host" }}
+            {{- /* Handle containers in host nework mode */}}
+            {{- if (index $.globals.networks "host") }}
+    #         both container and proxy are in host network mode, using localhost IP
+                {{- $ip = "127.0.0.1" }}
+                {{- continue }}
+            {{- end }}
             {{- range sortObjectsByKeysAsc $.globals.CurrentContainer.Networks "Name" }}
                 {{- if and . .Gateway }}
     #         container is in host network mode, using {{ .Name }} gateway IP
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -68,9 +68,16 @@
             {{- continue }}
         {{- end }}
         {{- if eq .Name "host" }}
-    #         {{ .Name }} (host network; using {{ (index $.globals.CurrentContainer.Networks 0).Name }} gateway)
-            {{- $ip = (index $.globals.CurrentContainer.Networks 0).Gateway }}
-            {{- continue }}
+            {{- range sortObjectsByKeysAsc $.globals.CurrentContainer.Networks "Name" }}
+                {{- if and . .Gateway }}
+    #         container is in host network mode, using {{ .Name }} gateway IP
+                    {{- $ip = .Gateway }}
+                    {{- break }}
+                {{- end }}
+            {{- end }}
+            {{- if $ip }}
+                {{- continue }}
+            {{- end }}
         {{- end }}
         {{- if and (not (index $.globals.networks .Name)) (not $.globals.networks.host) }}
     #         {{ .Name }} (unreachable)
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -160,6 +160,10 @@ def container_ip(container: Container):
         net_info = container.attrs["NetworkSettings"]["Networks"]
         if "bridge" in net_info:
             return net_info["bridge"]["IPAddress"]
+        
+        # container is running in host network mode
+        if "host" in net_info:
+            return "127.0.0.1"
 
         # not default bridge network, fallback on first network defined
         network_name = list(net_info.keys())[0]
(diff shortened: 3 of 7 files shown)


buchdag:
@rot169 thanks for the PR.

Unfortunately this change appears to break the tests entirely.

rot169:
Dang it, got tripped-up by a previous iteration of this change and didn't realise this one is broken, my bad... 🤦🏽‍♂️  Will try reworking this...

buchdag:
@rot169 I just tested it on a local dev machine and it does not work.

Compose file:

```yaml
version: '2'

services:
  nginx:
    image: nginx:1.23.4-alpine
    container_name: nginx-proxy
    volumes:
      - conf:/etc/nginx/conf.d
      - vhost:/etc/nginx/vhost.d
      - html:/usr/share/nginx/html
      - certs:/etc/nginx/certs:ro
    labels:                                                                                                                                                         
      - "com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_proxy"                   
    network_mode: "host"                                                                                                                                                                                                                        
                                                                                                                                                                    
  dockergen:                                                                                                                                                        
    image: nginxproxy/docker-gen:0.10.4
    container_name: nginx-proxy-gen                                                                                                                                                                                                                                                                                   
    command: -notify-sighup nginx-proxy -watch -wait 5s:30s /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf                                     
    volumes:                                                                                                                                                        
      - conf:/etc/nginx/conf.d                                                                                                                                      
      - vhost:/etc/nginx/vhost.d                                                                                                                                    
      - certs:/etc/nginx/certs:ro                                                                                                                                   
      - ./nginx.tmpl:/etc/docker-gen/templates/nginx.tmpl:ro
    labels:
      - "com.github.jrcs.letsencrypt_nginx_proxy_companion.docker_gen"
    network_mode: "host"

  letsencrypt:
    image: nginxproxy/acme-companion:2.2.8
    container_name: nginx-proxy-acme
    volumes:
      - vhost:/etc/nginx/vhost.d
      - html:/usr/share/nginx/html
      - certs:/etc/nginx/certs:rw
    network_mode: "host"

  foo:
    image: nginx:1.23.4-alpine-local
    build: .
    container_name: foo
    environment:
      VIRTUAL_HOST: foo.domain.tld
      LETSENCRYPT_HOST: foo.domain.tld
    network_mode: "host"

volumes:
  conf:
  vhost:
  html:
  certs:
```

Custom nginx Dockerfile for the proxyed container:
```Dockerfile
FROM nginx:1.23.4-alpine
COPY nginx.conf /etc/nginx/
EXPOSE 9080
CMD ["nginx", "-g", "daemon off;"]
```

Custom nginx config to have the proxyed nginx container listen on port 9080:
```nginx
user                            nginx;
worker_processes                auto;

error_log                       /var/log/nginx/error.log notice;

events {
    worker_connections          1024;
}

http {
    include                     /etc/nginx/mime.types;
    default_type                application/octet-stream;
    sendfile                    off;
    access_log                  off;
    keepalive_timeout           3000;
    server {
        listen                  9080;
        root                    /usr/share/nginx/html;
        index                   index.html;
        server_name             localhost;
        client_max_body_size    16m;
    }
}
```

Generated upstream block:
```nginx
# foo.domain.tld/
upstream foo.domain.tld {
    # Container: foo
    #     networks:
    #         host (host network; using host gateway)
    #     IP address: (none usable)
    #     exposed ports: (none)
    #     default port: 80
    #     using port: 80
    # Fallback entry
    server 127.0.0.1 down;
}
```

I can directly reach the proxyed nginx container on the host's port 9080 without issue.

buchdag:
`(index $.globals.CurrentContainer.Networks 0).Gateway` -> The host network does not have a Gateway.

rot169:
It looks like you're running the proxy container with host mode networking? The model that works is to run the proxy/dockergen/acme-companion with bridge networking (as normal) and it's only the upstream target that is configured to run in host mode.

rot169:
Not sure if they particularlly _wanted_ to run the proxy in host mode, or if it was just a way of trying to get it working with homeassistant (which has to run in a host mode container if you want it to do a load of network discovery). This is my exact use case too - I want to proxy to several containers running in a 'normal' bridge mode, and also to homeassistant running in host mode. I feel like most folks would want to keep bridge networking wherever possible, as it's part of the segregation benefits of containerisation, but I have no evidence to back up this assertion!

buchdag:
I added a minimal test.

@rhansen I'd like your advice on this, should we work a bit more on this feature (support proxy + container in host network mode, warn if `(index $.globals.CurrentContainer.Networks 0).Gateway` is empty because the network is private, iterate over `$.globals.CurrentContainer.Networks`, etc) or do you think the fix could be released in its current state with documented limitations ?

rhansen:
I'll take a look when I have some time next week.

I'm OK with marking this as experimental while we continue to evaluate.

buchdag:
@rot169 just to be sure: for a given container, host network mode and the other network modes are all mutually exclusive ?

buchdag:
I added the following to the PR:
- iteration over the proxy networks to find a suitable one rather than picking the first
- compatibility with the proxy running in host network mode itself
- tests for base PR scenario + proxy in host network mode
- documentation

@rhansen I'd still like your review on this, the template code might be sub optimal. Other than that I'm okay with the PR in its current state.

@rot169 could you test the PR in its current state ?

buchdag (approved):

buchdag:
Additional related issues:
- #1270
- #832

rot169:
> @rot169 just to be sure: for a given container, host network mode and the other network modes are all mutually exclusive ?

That's correct; as far as I understand it is not possible to have a container with both host-based network and a bridge network.

> @rot169 could you test the PR in its current state ?

Sure, will probably take me a few days, but will do so and report back.

buchdag:
I'm tempted to use `host.docker.internal` / `host.containers.internal` like [Traefik does](https://doc.traefik.io/traefik/providers/docker/#host-networking) with the current method as a fall back because that seems more straightforward and robust but that would require installing a DNS resolution util in the image, and the lightest I could find (`bind9-host`) adds ~40MB to the image.

prashker:
For my workaround with Home Assistant container being host networked, I used [qoomon/docker-host](https://github.com/qoomon/docker-host). The same solution is what I suspect people are using to make nginx-proxy handle non-dockerized applications. I don't like the approach, but it has worked for me.

The proposed solution looks great, and while I figure out how to build my own image, I will give a friendly ping to @rot169 if they've done their testing yet.

rot169:
Thanks for the ping - and apologies this'd dropped off my radar. Yes, the latest code seems to work great for me, and is clearly more robust than my initial proposal.

buchdag:
I spotted a typo in the doc I added, I'll fix this then probably merge this PR by the end of the week.

buchdag:
Fixed, thank you @prashker

prashker:
Thank you - sorry for the LONG delay. I can confirm this functionality works.

My `--net=host` Home Assistant instance has `VIRTUAL_HOST` `LETSENCRYPT_HOST` and `VIRTUAL_PORT` assigned and functioning as expected.

Good riddance to my workaround!

PR #2231, opened 2023-04-30
build: don't install wget inside images

buchdag:
This was only used a (long) while ago to get docker-gen and/or forego binaries.


buchdag:
To be honest I'm not certain we need `ca-certificates` either.

rhansen (approved):

Issue #1937, opened 2022-04-02
No upstreams generated when container has it's own IP


exander77:
I have containers with their own IPS assigned (so they are not using the host network), the IP is not detected and VIRTUAL_PORT is ignored. Would be great to at least set upstream by an environment variable.

exander77:
I solved this by this patch to support `VIRTUAL_ADDRESS`:
```
diff --git a/nginx.tmpl b/nginx.tmpl
index 610edf1..919c05d 100644
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -135,6 +135,12 @@ location {{ .Path }} {
                        {{ end }}
                {{ end }}
        {{ end }}
+       {{ range $container := .Containers }}
+    {{ if $container.Env.VIRTUAL_ADDRESS }}
+    {{ $server_found = "true" }}
+    server {{ $container.Env.VIRTUAL_ADDRESS }}:{{ $container.Env.VIRTUAL_PORT }};
+    {{ end }}
+    {{ end }}
        {{/* nginx-proxy/nginx-proxy#1105 */}}
        {{ if (eq $server_found "false") }}
         # Fallback entry
```

Issue #1474, opened 2020-07-21
Required VIRTUAL_UPSTREAM: support the web service not in docker. 


ccpaging:
My nginx-proxy is working in enterprise NAS based OpenMediaVault. There is a web service hosted in another windows machine based Apache/MySQL/Php. I can create a api.conf in /etc/nginx/conf.d to manage it now.

Is it possible to support VIRTUAL_UPSTREAM=192,168.8.10:8080,192.168.8.11:8080?I have tried it by modified nginx.tmpl. It's working.

I saw VIRTUAL_PATH in docker_gen. It is useful in enterprise server also.

I can help for testing.

minakmostoles:
I want to do similar something. I join the question. I tried to make a fake container with environment variables.
My docker compose have this form  

services:
  nginx-proxy:
    image: jwilder/nginx-proxy
    container_name: nginxproxy
    restart: always
    ports:
      - "80:80"
      - "443:443"
    volumes:
      - /var/run/docker.sock:/tmp/docker.sock:ro
      - /app/nginx/conf.d:/etc/nginx/conf.d
      - /app/nginx/certs:/etc/nginx/certs:ro
      - /app/nginx/vhostd:/etc/nginx/vhost.d
      - /app/nginx/html:/usr/share/nginx/html
      #- /app/nginx/logs:/var/log/nginx
    labels:
      - com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_proxy=true 
    network_mode: "host"
    logging:
      driver: "json-file"
      options:
        max-size: "25m"
...
  jellyfin:
    image: cwempe/docker-dummy
    container_name: jellyfin
    depends_on:
      - nginx-proxy
    environment:
      - VIRTUAL_HOST=subdomain.domain.com
      - VIRTUAL_PORT= 32400
      - UPSTREAM_NAME= <IP_SERVER_EXTERNAL>
      - NETWORK_ACCESS=external
      - HTTPS_METHOD=nohttp
    restart: always
    logging:
      driver: "json-file"
      options:
        max-size: "1m"

I have tried many settings but without success. 
I've looked at the issues #1191 and #1100 , but finaly i have 500 Internal Server Error because the nginx conf uses the dummy container address.

Thank you very much

ccpaging:
Modify nginx.tmpl as below:
```
# {{ $host }}
upstream {{ $upstream_name }} {

{{ range $container := $containers }}
[...]

PR #2234, opened 2023-05-05
Add support for remote backends via VIRTUAL_IP

f403:
This adds support for non-container upstreams (e.g. somewhere on LAN).
Some time ago I used a patch from #1100 to solve this issue, but current template has changed a lot.
Although running a dummy container for each custom upstream  is a bit hacky, but it keeps same approach for all deployments, implements a feature requested many times, and does this with minimal changes in the template.

This PR may potentially close:
- #832 
- #1937 
- #1191 
- #1474
- #1270
- #1537

Related PRs (some may be hard to merge now, if they are old):
- #1100 
- #1219 
- #1477
- #2222

### DIFF
--- a/README.md
+++ b/README.md
@@ -397,6 +397,38 @@ docker run -d -p 80:80 -p 443:443 \
 
 You'll need apache2-utils on the machine where you plan to create the htpasswd file. Follow these [instructions](http://httpd.apache.org/docs/2.2/programs/htpasswd.html)
 
+### Custom Upstream (Backend) Server Support
+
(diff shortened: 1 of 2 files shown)


buchdag:
Hi.

Thanks for the PR.

Unfortunately while I do understand that's a wanted feature, [my answer from mid 2021](https://github.com/nginx-proxy/nginx-proxy/pull/1100#issuecomment-864572115) still stands: this is out of the original stated scope of this project, and I don't see it implemented using zombie containers.

PR #2238, opened 2023-05-08
build: get forego from pre-built image

buchdag:
Pretty much the same as #2230 but for forego, which now have a pre built image (`nginxproxy/forego`).

The [diff between version `0.17.0` and `0.17.1`](https://github.com/nginx-proxy/forego/compare/v0.17.0...v0.17.1) is pretty much just dependencies maintenance and the new Dockerfiles + GitHub Action workflow required for the images building and publishing.

I don't plan on doing any work on forego beside this and the occasional dependencies bump, so this PR aim is mostly to further simplify nginx-proxy Dockerfiles and get yet another small bump in build speed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,6 @@
-# setup build arguments for version of dependencies to use
-ARG FOREGO_VERSION=v0.17.0
-
 FROM nginxproxy/docker-gen:0.10.4-debian AS docker-gen
 
-# Build forego from scratch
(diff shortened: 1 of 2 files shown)


rhansen (approved):

buchdag:
Just FYI we went from ~16 minutes build time to ~2 minutes build time by using pre built docker-gen and forego.

PR #2186, opened 2023-03-01
fix: Improve handling of unknown hosts and missing certs

rhansen:
* fix: Don't create cert error https server if https is not enabled
  * fix: Emit TLS error if there are no certs available
  * fix: Don't create fallback http(s) server when http(s) disabled

plus a couple of cleanups

PR #2239, opened 2023-05-08
fix: always on session cache on HTTPS fallback listener

buchdag:
This PR should fix an issue with Secure Renegociation (RFC 5746) [identified](https://github.com/nginx-proxy/nginx-proxy/pull/2186#issuecomment-1533863923) by @SchoNie and introduced in #2186 

That's counterintuitive but it seems like the `ssl_session_cache` must be enabled as soon as there is an active fallback HTTPS listener if we want secure renegociation to work, even if we're going to return a TLS error from the fallback listener.

@SchoNie I tested the PR with both SSL Labs and OpenSSL and it appears to fix the issue, could you confirm ?

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -422,11 +422,11 @@ server {
             {{- if $globals.enable_ipv6 }}
     listen [::]:{{ $globals.external_https_port }} ssl http2; {{- /* Do not add `default_server` (see comment above). */}}
             {{- end }}
+    ssl_session_cache shared:SSL:50m;
+    ssl_session_tickets off;
         {{- end }}
     {{ $globals.access_log }}
         {{- if $globals.default_cert_ok }}
-    ssl_session_cache shared:SSL:50m;
-    ssl_session_tickets off;
     ssl_certificate /etc/nginx/certs/default.crt;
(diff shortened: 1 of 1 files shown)


SchoNie:
Yes, thanks that works for me.
testssl.sh still fails `sudo docker run --rm -ti -v "./tmp":/tmp drwetter/testssl.sh --debug 3 -R domain.ltd`  (debug logs are written to /tmp) but I suspect that is because the TLS handshake is hitting the fallback directive. (its giving me the year 2050 cert) But I don't understand why because I use a valid servername, so it should hit the server vhost directive where certs are valid.

Not sure if that can be solved, but just to tell my finding.
The test is around [line 16890](https://github.com/drwetter/testssl.sh/blob/3.1dev/testssl.sh#L16890) from testsll.sh

buchdag:
Regarding the test with `drwetter/testssl.sh`, are you testing using a publicly reachable domain or a local test domain ?

I remember seeing issues on the project's GitHub with the former.

SchoNie:
> I remember seeing issues on the project's GitHub with the later.

Seems so yes, now tested with a public domain and all tests pass also the `drwetter/testssl.sh`
Thanks

rhansen (approved):

PR #2250, opened 2023-05-22
fix: Replace mDNS .local domain with .internal domain

SaturnIC:
Issue: https://github.com/nginx-proxy/nginx-proxy/issues/2249

.local domains are special use domains for mDNS and must not be used for other routing purposes,
use .internal instead in documentation.

See RFC 6762 for details: https://datatracker.ietf.org/doc/html/rfc6762

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,13 +68,13 @@ services:
     expose:
       - "8000"
     environment:
-      - VIRTUAL_HOST=whoami.local
+      - VIRTUAL_HOST=whoami.example
       - VIRTUAL_PORT=8000
 ```
 
 ```console
 docker-compose up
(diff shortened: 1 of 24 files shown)


SaturnIC:
Another alternative to .internal would be .example domain according to RFC 6761

buchdag:
Hi.

Thanks for the PR.

Do you think it would be possible to use `.example` rather than `.internal` as you suggested above ? I think it might be less confusing for the docs sections.

SaturnIC:
Thank you for the example suggestion, i agree it makes even more sense to use that domain.

I changed the commit as suggested.

buchdag (approved):

PR #2274, opened 2023-07-12
New AWS SSL policies

buchdag:
New SSL policies from #1644 by @patrickdk77.


nginx.tmpl
```diff
@@ -161,6 +160,30 @@
     ssl_protocols TLSv1 TLSv1.1 TLSv1.2 TLSv1.3;
     ssl_ciphers 'ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:DHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES256-SHA256:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA';
     ssl_prefer_server_ciphers on;
+    {{- else if eq .ssl_policy "AWS-FS-1-2-Res-2020-10" }}
+    ssl_protocols TLSv1.2 TLSv1.3;
```
rhansen, line 164: If I'm reading the [AWS docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html#tls-security-policies) correctly, none of these newly added policies support TLSv1.3.  If the addition of TLSv1.3 is intentional, please add a comment explaining why it differs from the AWS policy.

nginx.tmpl
```diff
@@ -144,12 +144,11 @@
     ssl_protocols TLSv1.3;
         {{- /*
              * nginx currently lacks ability to choose ciphers in TLS 1.3 in
-             * configuration; see https://trac.nginx.org/nginx/ticket/1529.  A
-             * possible workaround can be modify /etc/ssl/openssl.cnf to change
-             * it globally (see
-             * https://trac.nginx.org/nginx/ticket/1529#comment:12).  Explicitly
-             * set ngnix default value in order to allow single servers to
-             * override the global http value.
+             * configuration, see https://trac.nginx.org/nginx/ticket/1529.
```
rhansen, line 147: The linked bug is closed as fixed. Is this comment still valid?

rhansen (changes requested):

nginx.tmpl
buchdag: It's not, I lifted the code from #1644 without even thinking to double check.

Furthermore AWS now have a grand total of 18 SSL policies, and I'm wondering is those should be added too.

nginx.tmpl
buchdag: I asked myself the same question a few months ago, I did not have time to properly test yet.

nginx.tmpl
rhansen: I just learned that [the v1.3 and <=v1.2 cipher suites are disjoint sets](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4).  None of these new policies mention any TLSv1.3 cipher suites.  It's unclear to me how nginx/OpenSSL behaves in this case.  I see three possibilities:
  * TLSv1.3 is effectively disabled, as if it was excluded from `ssl_protocols`.
  * TLSv1.3 negotiations always fail due to lack of acceptable cipher suites. (Perhaps causing clients to retry with TLS v1.2?)
  * The default set of TLSv1.3 cipher suites is used.

Regardless, I think that TLSv1.3 should be removed from these new policies.  I'm hesitant to remove it from the existing policies without testing because that might break compatibility.

> 
> Furthermore AWS now have a grand total of 18 SSL policies, and I'm wondering is those should be added too.

That would be nice, although that can be done in a future PR if you don't feel like doing it now.

PR #2278, opened 2023-07-18
feat: experimental HTTP/3 support + optional HTTP/2 disabling

### LINKED ISSUE #1055: Disable HTTP/2 http2 option
Some of the applications behind `nginx-proxy` do not handle SPDY protocol errors well, for example by resetting the DOM and requiring a force reload of the pages. When these pages are supposed to be automatically refreshed so they can be monitored, this is a nuisance.

I cannot find much detail on how SPDY protocol errors occur but in my use case, I will not miss the usage of HTTP/2 anyway - can we get an ENVAR that allows us to disable HTTP/2 on the nginx installation?

### LINKED ISSUE #2170: QUIC and HTTP/3 Support
Can you support it? Possible in different docker branch while not in mainline
https://www.nginx.com/blog/our-roadmap-quic-http-3-support-nginx/

buchdag:
@SchoNie I went the extra mile and opened a draft PR with what I got so far. It seems mostly functional to me, let me know if you think something is missing / should be done differently.

It indeed work when replacing `proxy_set_header Host $http_host;` with `proxy_set_header Host $host;`, should we also replace the other uses of `$http_host` beside this one ?

The PR still lacks documentation at this point, and I don't think it can be automatically tested beside parsing the rendered template (Python's `request` is not even capable of doing HTTP/2).

New labels on the **proxied** containers:
- `com.github.nginx-proxy.nginx-proxy.http2.enable`:
enable HTTP/2 support when set to `true` (default: `true`)
- `com.github.nginx-proxy.nginx-proxy.http3.enable`:
enable HTTP/3 support when set to `true` (default: `false`)

New environment variables on the **proxy** container:
- `ENABLE_HTTP2`: globally enable HTTP/2 when set to `true` (default: `true`)
- `ENABLE_HTTP3`: globally enable HTTP/3 when set to `true` (default: `false`)

closes #2170
closes #1055

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -203,10 +203,6 @@
         include /etc/nginx/network_internal.conf;
         {{- end }}
 
-        {{- if .http3 }}
-        add_header alt-svc 'h3=":{{ $.globals.external_https_port }}"; ma=86400;';
-        {{- end }}
-
         {{- if eq .Proto "uwsgi" }}
         include uwsgi_params;
         uwsgi_pass {{ trim .Proto }}://{{ trim .Upstream }};
@@ -281,8 +277,8 @@ map $http_x_forwarded_proto $proxy_x_forwarded_proto {
 }
 
 map $http_x_forwarded_host $proxy_x_forwarded_host {
-    default {{ if $globals.trust_downstream_proxy }}$http_x_forwarded_host{{ else }}$http_host{{ end }};
-    '' $http_host;
+    default {{ if $globals.trust_downstream_proxy }}$http_x_forwarded_host{{ else }}$host{{ end }};
+    '' $host;
 }
 
 # If we receive X-Forwarded-Port, pass it through; otherwise, pass along the
@@ -337,7 +333,7 @@ map $proxy_x_forwarded_proto $proxy_x_forwarded_ssl {
 
 gzip_types text/plain text/css application/javascript application/json application/x-javascript text/xml application/xml application/xml+rss text/javascript;
 
-log_format vhost '{{ or $globals.Env.LOG_FORMAT "$host $remote_addr - $remote_user [$time_local] \"$request\" $status $body_bytes_sent \"$http_referer\" \"$http_user_agent\" \"$upstream_addr\" $http3" }}';
+log_format vhost '{{ or $globals.Env.LOG_FORMAT "$host $remote_addr - $remote_user [$time_local] \"$request\" $status $body_bytes_sent \"$http_referer\" \"$http_user_agent\" \"$upstream_addr\"" }}';
 
 access_log off;
 
@@ -443,6 +439,7 @@ proxy_set_header Proxy "";
 server {
     server_name _; # This is just an invalid value which will never trigger on a real hostname.
     server_tokens off;
+    {{ $globals.access_log }}
     http2 on;
         {{- if $fallback_http }}
     listen {{ $globals.external_http_port }}; {{- /* Do not add `default_server` (see comment above). */}}
@@ -452,19 +449,19 @@ server {
         {{- end }}
         {{- if $fallback_https }}
     listen {{ $globals.external_https_port }} ssl; {{- /* Do not add `default_server` (see comment above). */}}
-            {{- if $http3 }}
-    listen {{ $globals.external_https_port }} quic reuseport; {{- /* Do not add `default_server` (see comment above). */}}
-            {{- end }}
             {{- if $globals.enable_ipv6 }}
     listen [::]:{{ $globals.external_https_port }} ssl; {{- /* Do not add `default_server` (see comment above). */}}
-                {{- if $http3 }}
+            {{- end }}
+            {{- if $http3 }}
+    http3 on;
+    listen {{ $globals.external_https_port }} quic reuseport; {{- /* Do not add `default_server` (see comment above). */}}
+                {{- if $globals.enable_ipv6 }}
     listen [::]:{{ $globals.external_https_port }} quic reuseport; {{- /* Do not add `default_server` (see comment above). */}}
                 {{- end }}
             {{- end }}
     ssl_session_cache shared:SSL:50m;
     ssl_session_tickets off;
         {{- end }}
-    {{ $globals.access_log }}
         {{- if $globals.default_cert_ok }}
     ssl_certificate /etc/nginx/certs/default.crt;
     ssl_certificate_key /etc/nginx/certs/default.key;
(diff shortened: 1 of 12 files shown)


nginx.tmpl
```diff
@@ -333,7 +337,7 @@ map $proxy_x_forwarded_proto $proxy_x_forwarded_ssl {
 
 gzip_types text/plain text/css application/javascript application/json application/x-javascript text/xml application/xml application/xml+rss text/javascript;
 
-log_format vhost '{{ or $globals.Env.LOG_FORMAT "$host $remote_addr - $remote_user [$time_local] \"$request\" $status $body_bytes_sent \"$http_referer\" \"$http_user_agent\" \"$upstream_addr\"" }}';
+log_format vhost '{{ or $globals.Env.LOG_FORMAT "$host $remote_addr - $remote_user [$time_local] \"$request\" $status $body_bytes_sent \"$http_referer\" \"$http_user_agent\" \"$upstream_addr\" $http3" }}';
```
SchoNie, line 340: Also add `$http2`
It's empty if not negotiated , just like `$http3`

nginx.tmpl
```diff
@@ -549,7 +571,9 @@ server {
     {{- if $server_tokens }}
     server_tokens {{ $server_tokens }};
     {{- end }}
+    {{- if $http2 }}
```
SchoNie, line 574: http2 disabling will not work because it's enabled in the fallback block.
Nginx creates listen processes for for each ip:port combination and then sums the capabilities of every host. So the `http2 on` from the fallback does globally turn it on.

So the label on the **proxied** container will never work. If 1 vhost is http2 enabled, it will be for all. I think that can be removed.
The environment on the **proxy** container could work. But then [line #446](https://github.com/nginx-proxy/nginx-proxy/blob/12653debb5562c227a5f1ac5703bb942c6167173/nginx.tmpl#L446) also needs to be behind the if statement.

nginx.tmpl
```diff
@@ -549,7 +571,9 @@ server {
...
     http2 on;
+    {{- end }}
```
SchoNie, line 576: `http3 on` is default nginx behaviour.
Which seems unlogical for me, because http2 is still disabled by default. I know if you do not set the quic listener it is kind of disabled.
But for reading purposes would it be an idea to add `http3 on`? I think for debugging purposes it is more clear if you see that directive instead of assuming. And maybe the default would change if nginx thinks it's more mature.

SchoNie (changes requested):
[I think it is safe to change to `$host`](https://github.com/nginx-proxy/nginx-proxy/pull/2278/files#diff-a2a77c71854a4489d896f49c59a5f864e5b1e290f6b1ba16f9f39c1a32829c57R284) there is only 1 instance left in the `$http_x_forwarded_proto` map.
I ran the tests with change and do not encounter any difference.

nginx.tmpl
buchdag: Agreed, and thanks for clarifying that it is on by default, I was under the impression that `http3 on` did nothing. But why on earth did they made it default to on if it's considered an experimental feature ?

nginx.tmpl
buchdag: Thanks for those information, I'll keep the global variable with the needed modification and remove the proxied container label.

SchoNie: I was thinking some more and maybe its better to not have the option to disable http2 at all.
All currently supported QUIC browsers seem to connect over http1/http2 first.
When the alt-svc header is provided, the browser upgrades to QUIC and caches that for the time specified in alt-svc.
After that cache period a tcp (http2) request is done to negotiate this again.

And for the next foreseeable future you probably want http2 as fallback and for now adds unnecessary overhead.

80 > 443 (http2) > alt-svc (http2) > QUIC (http3)

buchdag: The HTTP/2 disabling feature is a separate request that people made specifically (see https://github.com/nginx-proxy/nginx-proxy/pull/1644#issuecomment-851297376), it was added to this PR because it kind of made sense but it does not imply in any way that you have to or even should disable HTTP/2 when using HTTP/3.

buchdag: Additional request for HTTP/2 disabling feature : https://github.com/nginx-proxy/nginx-proxy/issues/1055

buchdag: After testing per virtual host HTTP/2 disabling, it does appear to work as intended. Requests made with `curl --http2`, www.foo.buz has HTTP/2 enabled (default), www.foo.bar has HTTP/2 disabled via the label:
```console
nginx-proxy          | nginx.1     | www.foo.buz 172.22.0.1 - - [22/Jul/2023:08:35:17 +0000] "GET / HTTP/2.0" 200 615 "-" "curl/8.2.0-DEV" "172.22.0.2:80"
nginx-proxy          | nginx.1     | www.foo.bar 172.22.0.1 - - [22/Jul/2023:08:35:22 +0000] "GET / HTTP/1.1" 200 615 "-" "curl/8.2.0-DEV" "172.22.0.3:80"
```

nginx.tmpl
buchdag: I'll end up removing `$http3` because we already have the protocol through `$request`.

nginx.tmpl
buchdag: Done ✅

buchdag:
I re-organised the template a bit so that HTTP/3 related stuff is a bit more grouped together.

More importantly I moved the `add-header` statement for Alt-Svc from the location template to the server block to avoid overriding the HSTS header (thanks @knapoc for spotting this in #2271).

I also removed the `$http3` variable from logs because as explained in discussion above we already have the protocol logged through the `$request` variable.

And finally I replaced the last `$http_host` instance with `$host`.

buchdag:
@SchoNie @Knapoc is it okay if I add you as co-authors on relevant commits ?

SchoNie:
👍
I did create a few tests. I adjusted the hsts test to check for overridden headers.
And made soms basic http3 tests. Mostly to check for config syntax as there is not really a method to test by requests.

Will try to push later when at office and then you can improve or discard😅

cool this PR is getting along! Hopefully it helps more people to enable QUIC and maybe it encourages to use of this project as a early adopter.

SchoNie:
HSTS override tests, http2 & http3 tests: https://github.com/SchoNie/nginx-proxy/commit/618da602bc3614e17863095200d5db4898e216a8
Feel free to use, improve or discard.

buchdag:
@SchoNie sorry for the lack of activity on august, the tests looked interesting, do you still have them ?

SchoNie:
@buchdag Yes. Still have the tests: https://github.com/nginx-proxy/nginx-proxy/commit/27cc611bb7b62fe1fb55cefa42fe4b76a4a7eca0

PR #2279, opened 2023-07-24
feat: allow nginx / docker-gen network segregation

Knapoc:
This PR adds the ability to allow for network segregation of the nginx and docker-gen containers in a **separate container setup**.

As of now this is not possible as it is assumed, that docker-gen and nginx are attached to the same networks.

In case

1. docker-gen and nginx do not share a network,
2. docker-gen cannot reach the backend,
3. and nginx **would** be able to connect to the backends,

docker-gen would set the server to a fallback-entry within the upstream directive.

This behavior is changed with this PR when the nginx-container is known to docker-gen (by having a label applied). In this case the fallback-entry is only set, when the backend is not available to nginx.


Knapoc:
@buchdag sorry to ping you, but is there a chance that this gets merged?

buchdag:
@Knapoc yes I'd like to see this merged, but I know worked on it on a different computer last december and it looks like I lost my work. I can't remember everything I did but I think I made the label configurable.

Knapoc:
@buchdag your changes make total sense and provide a cleaner solution. I'm totally fine with them. Also thank you for considering this feature/pull requests.

buchdag (approved):

buchdag:
@Knapoc thank you for the PR and sorry for the long wait 😅

SchoNie:
The tests seems to be a bit flaky. Failing for me most of the time.
Also in: [Merge pull request #2602 from p12tic/acme-unknown-virtual-host](https://github.com/nginx-proxy/nginx-proxy/actions/runs/15124095460/job/42512768167)

buchdag:
@SchoNie indeed the tests for the feature I added (custom label) seems flaky, I'll remove the feature and test for now until I can understand what's happening / reproduce what's going on during tests.

PR #2281, opened 2023-07-28
Upgrade create_server_certificate.sh to nginx:latest for less maintenance

SchoNie:
Simple version bump of nginx used in `create_server_certificate.sh` because dependabot will not scan in shell scripts.
Using the latest tag for less maintenance.

### DIFF
--- a/test/certs/create_server_certificate.sh
+++ b/test/certs/create_server_certificate.sh
@@ -24,7 +24,7 @@ fi
 # Create a nginx container (which conveniently provides the `openssl` command)
 ###############################################################################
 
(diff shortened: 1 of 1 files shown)


buchdag:
I'm more inclined to keep a fixed version here, if we use a `latest` tag the utils that depends on the openssl shipping with the nginx container might suddenly break, and this might be a pain to diagnose because it'll be hard to determine the previous working version. It's the same thing for [Boulder](https://github.com/letsencrypt/boulder) in the nginx-proxy/acme-companion tests, we used to test against the `latest` image but that caused way too much unexpected test breakage for my taste. It's now a fixed version upgraded from time to time.

I'm not against a version bump though 👍

SchoNie:
Yes, agreed!
#2283

PR #2321, opened 2023-10-21
Add support for grpcs protocol

theFra985:


buchdag:
Hi. I'm not sure I understand the point of this PR. The added code is already there under the `grpc` proto.

theFra985:
Hi, it doesn't, really. My issue was the lack of `grpcs://` protocol support as the rule added in the grpc PR doesn't include the grpcs protocol while matching the specified protocol and while deploying PMM under nginx-proxy it was required to have grpcs as the protocol instead of grpc for it to work properly.
An alternative would be to match both of them in the same condition but the grpc PR lacks a grpcs option

buchdag:
@theFra985 my bad, I misunderstood the change.

buchdag (approved):

PR #2349, opened 2023-12-19
Tests: support custom 'docker compose' command

pini-gh:
Enable overriding default 'docker compose' command with environment variable 'DOCKER_COMPOSE'. This way docker compose v1 is still supported with:

  $ DOCKER_COMPOSE=docker-compose pytest

This is important because people using the Debian packaged docker compose are stuck to v1.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -28,6 +28,9 @@ need more verbosity ?
 
     pytest -s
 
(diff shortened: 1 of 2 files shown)


buchdag:
@pini-gh thanks for the PR, do you think you could document the environment variable in `test/README.md` ?

pini-gh:
Sure.

buchdag (approved):

PR #2355, opened 2023-12-22
Fix xpass and xfails test_restart_while_missing_cert.py

SchoNie:
Tests in `stress_tests\test_restart_while_missing_cert.py` were being marked as xfailed and some others are partial. But I do not see a special reason. (nor could find in the commit history).
So I removed the xfail and incremental markers. And fixed 2 tests.

Now CI should have 100% passed score.

If there was a reason they should be xfail / xpass where I was not aware off just delete this PR.


buchdag:
I think I also stumbled upon this about a year ago, can't remember why I did not remove those at the time (might be that the test where still occasionally failing). I'll go ahead and merge this PR, if those tests start failing again we'll revert and investigate.

buchdag (approved):

PR #2377, opened 2024-01-22
add http include

azlux:
for https://github.com/nginx-proxy/nginx-proxy/discussions/2376


azlux:
Not useful for me anymore because don't use the app with this need of config.
Maybe for other usecase.

PR #2393, opened 2024-02-13
tests: non standard port Host header check

SchoNie:
Tests for #2388: Make sure the port number is included in the Host header for non standard ports.


buchdag (approved):

buchdag:
Thanks @SchoNie 👍🏻

PR #2402, opened 2024-02-19
tests: htpasswd

SchoNie:
Tests for `VIRTUAL_HOST` and `VIRTUAL_PATH` htpasswd behaviour
Draft until #2401 is merged (assertion on path without double slash)


buchdag:
Thanks @SchoNie 👍

PR #2410, opened 2024-03-09
refactor: add 'ports' dict to the data structure

pini-gh:
Nothing changes for the legacy syntax, but adding this level to the data structure enables advanced port configuration.


buchdag (approved):

buchdag:
Pretty much what I had in mind next, thank you 👍

PR #2420, opened 2024-04-08
correctly handle redirect logic for path-base routing

chenqiaoanying:
Now path-base routing can not correctly handle redirect logic for path-base routing
The example is as below, this is a demo docker compose file to reproduce the issue.
```docker-compose.yaml
version: '3.8'

services:
  nginx-proxy:
    image: nginxproxy/nginx-proxy
    restart: always
    environment:
      - ENABLE_IPV6=true
      - DEFAULT_HOST=localhost
    volumes:
    ports:
      - 80:80
      - 443:443
    networks:
      - nat-bridge

  jellyfin:
    image: linuxserver/jellyfin
    restart: always
    environment:
      - VIRTUAL_HOST=localhost
      - VIRTUAL_PORT=8096
      - VIRTUAL_PATH=/jellyfin/
      - VIRTUAL_DEST=/
    ports:
      - 8096:8096
    networks:
      - nat-bridge
networks:
  nat-bridge:
```

start it and access http://localhost/jellyfin/ from browser, and you can see the issue as below:
The redirect response do not go jellyfin app.
image

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -276,6 +276,9 @@
         grpc_pass {{ trim $proto }}://{{ trim $upstream }};
         {{- else }}
         proxy_pass {{ trim $proto }}://{{ trim $upstream }}{{ trim $dest }};
+            {{- if not (eq .Path "/") }}
+        proxy_redirect ~^/(.*)$ {{ trimSuffix `/` .Path }}/$1;
+            {{- end }}
         set $upstream_keepalive {{ if ne $keepalive "disabled" }}true{{ else }}false{{ end }};
         {{- end }}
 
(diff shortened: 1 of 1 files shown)


SchoNie:
This is more something you should configure in your application (Jellyfin in your case): [Base URL configuration for Jellyfin](https://jellyfin.org/docs/general/networking/#base-url)

> The Base URL setting in the Networking page is an advanced setting used to specify the URL prefix that your Jellyfin instance can be accessed at. In effect, it adds this URL fragment to the start of any URL path. For instance, if you have a Jellyfin server at http://myserver and access its main page http://myserver/web/index.html, setting a Base URL of /jellyfin will alter this main page to http://myserver/jellyfin/web/index.html. This can be useful if administrators want to access multiple Jellyfin instances under a single domain name, or if the Jellyfin instance lives only at a subpath to another domain with other services listening on /.

Issue #1325, opened 2019-09-05
Custom Nginx Error Messages?


dm17:
It seems like most people use CDNs for this type of functionality these days, but is there some way to have "site will be back momentarily" sign until all of the nginx-proxy-fed docker services finish starting? This seems like a pretty common use case, so I figured someone here would have a good suggestion or extensions they've written to do this. So if visiting the domain while the services are still coming up - instead of seeing an nginx error (meaningless to most people) - it shows a "please check back in a few moments" message.

I had previous posted this in the wrong project:
https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/issues/574

hasnat:
in vhost.d nginx config for your container have something like
```
        error_page 503 = @custom_error;
        location @custom_error {
                internal;
                return 503 'please check back in a few moments';
        }
```
not 100% if its 503 or some other error you have to cover.

dm17:
Great, very useful... But how do I get it so the error appears immediately in the browser. If the site isn't up and this gets triggered, then it seems the browser will always display this first:

Attackers might be trying to steal your information from [domain] (for example, passwords, messages, or credit cards). Learn more
NET::ERR_CERT_AUTHORITY_INVALID

Then you have to click twice to proceed.

iNviNho:
@dm17 I dont think you can do anything with that. Browser remembers that site is being loaded over HTTPS and if your site is down/certificate is invalid, user will get this error from browser.

sgabe:
Note that you might want to use the `proxy_intercept_errors on;` directive to intercept and redirect proxied error responses to nginx for processing with the `error_page` directive in order to prevent information exposure through error messages.

PR #2430, opened 2024-05-02
feat: custom default error page

buchdag:
This PR restore the default error page provided by the nginx container and allows to customise it by mounting another HTML file in its place.

Solve #1325

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -799,7 +799,7 @@ Per virtual-host `servers_tokens` directive can be configured by passing appropr
 
 ### Custom error page
 
-To override the default (nginx provided) error page displayed on 50x errors, mount your custom HTML error page inside the container at `/usr/share/nginx/html/errors/50x.html`:
+To override the default error page displayed on 50x errors, mount your custom HTML error page inside the container at `/usr/share/nginx/html/errors/50x.html`:
 
 ```console
 docker run --detach \
--- a/docs/README.md
+++ b/docs/README.md
@@ -797,6 +797,21 @@ location / {
 
 Per virtual-host `servers_tokens` directive can be configured by passing appropriate value to the `SERVER_TOKENS` environment variable. Please see the [nginx http_core module configuration](https://nginx.org/en/docs/http/ngx_http_core_module.html#server_tokens) for more details.
 
+### Custom error page
+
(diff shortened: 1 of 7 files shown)


nginx.tmpl
```diff
@@ -655,7 +655,17 @@ server {
...
-    return 503;
+
+        {{- if (exists "/usr/share/nginx/html/50x.html") }}
+    error_page 500 502 503 504 /50x.html;
+    location /50x.html {
+        root /usr/share/nginx/html;
+        internal;
+    }
+        {{- end }}
+    location ^~ / {
```
Prime1Code, line 666: In case the user navigates to "example.com" or "www.example.com" he gets a 503, right?

If so: how to serve a custom root-Page without the need of a separate nginx serving just that root path?

buchdag: I'm not certain I understand the question. 🤔 

This `server` block (the one which is modified) is a catch all for when you're requesting a hostname that does not exist on a particular instance of nginx-proxy (It's never used when real proxying takes place). The purpose of the PR is to allow customisation of the default error page you get when this happen.

The `location` block is required because having only `return 503` does not trigger the `error_page`.

Prime1Code (approved):

buchdag:
I'm not certain this PR should restore the nginx image's 50x page by default.

This would switch from this, which have been there since the beginning :

79203896-fd88e980-7e3b-11ea-9147-44908aa3e665

To this:

image

That's arguably easier to the eye but that might not be what users want as default either.

@Prime1Code any opinion on this ?

Prime1Code:
Yes, I think so too. This error page should not be enabled by default. 

But the option to add one is still a better customization option than without it.

Prime1Code (approved):

docs/README.md
```diff
@@ -797,6 +797,21 @@ location / {
 
 Per virtual-host `servers_tokens` directive can be configured by passing appropriate value to the `SERVER_TOKENS` environment variable. Please see the [nginx http_core module configuration](https://nginx.org/en/docs/http/ngx_http_core_module.html#server_tokens) for more details.
 
+### Custom error page
+
+To override the default (nginx provided) error page displayed on 50x errors, mount your custom HTML error page inside the container at `/usr/share/nginx/html/errors/50x.html`:
```
buchdag, line 802:
```suggestion
To override the default error page displayed on 50x errors, mount your custom HTML error page inside the container at `/usr/share/nginx/html/errors/50x.html`:
```

PR #2432, opened 2024-05-03
fix: tests webserver syntaxwarning

SchoNie:
> Changed in Python version 3.12: Unrecognized escape sequences produce a SyntaxWarning. In a future Python version they will be eventually a SyntaxError.
[2.4.1.1. Escape sequences](https://docs.python.org/3/reference/lexical_analysis.html#index-23)

Change to raw string literal to fix:
```
/webserver.py:17: SyntaxWarning: invalid escape sequence '\d'
  elif re.match("/status/(\d+)", self.path):
/webserver.py:18: SyntaxWarning: invalid escape sequence '\d'
  result = re.match("/status/(\d+)", self.path)
```


buchdag (approved):

buchdag:
Thanks !

PR #2011, opened 2022-07-10
fix: include a complete fastcgi_params config

rodrigoaguilera:
Fixes https://github.com/nginx-proxy/nginx-proxy/issues/1466

PR #2449, opened 2024-05-16
fix: include a complete fastcgi_params config

### LINKED ISSUE #1466: fastcgi does not work
I am unable to get this to work via fastcgi with a simple php-fpm container. Here's my `docker-compose.yml` file:

```yaml
version: "3"

services:
  proxy:
    image: jwilder/nginx-proxy
    ports:
      - "80:80"
    volumes:
    environment:
      - DEFAULT_HOST=test.local

  fpm:
    image: php:fpm
    environment:
      - VIRTUAL_HOST=test.local
      - VIRTUAL_PROTO=fastcgi
```

I then drop in a simple `index.php` file by running:

```sh
 docker container exec -it web_fpm_1 /bin/bash -c 'echo "<?php phpinfo(); ?>" > /var/www/html/index.php'
```

(It puts `web_` in front because this project is in a directory named `web/`.)

I also modify my `hosts` file to point `test.local` to `127.0.0.1`, so I can test it.
However, every attempt to browse to `test.local` results in a blank white page.

The logs for the `web_fpm_1` container show that nothing gets sent except a 200 response:
```
[06-Jul-2020 15:40:39] NOTICE: fpm is running, pid 1
[06-Jul-2020 15:40:39] NOTICE: ready to handle connections
172.19.0.3 -  06/Jul/2020:15:41:02 +0000 "- " 200
172.19.0.3 -  06/Jul/2020:15:41:03 +0000 "- " 200
```

What am I doing wrong?

buchdag:
Rebase of #2011 by @rodrigoaguilera 

Should fix #1466

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -265,7 +265,7 @@
         uwsgi_pass {{ trim $proto }}://{{ trim $upstream }};
         {{- else if eq $proto "fastcgi" }}
         root {{ trim .VhostRoot }};
-        include fastcgi_params;
+        include fastcgi.conf;
         fastcgi_pass {{ trim $upstream }};
             {{- if ne $keepalive "disabled" }}
         fastcgi_keep_conn on;


buchdag:
@rodrigoaguilera @tkw1536  I'm really unfamiliar with PHP-FPM, before I merge this could you explain to me what is the effect of

```nginx
fastcgi_param  SCRIPT_FILENAME    $document_root$fastcgi_script_name;
```

because this additional line is the only difference between the `fastcgi_params` and the `fastcgi.conf` files.

By merging this, isn't it possible that we break setup that worked with `fastcgi_params` but won't with `fastcgi.conf` ?

I've also seen this mentioned as a possible fix instead of the former line in #1466:

```nginx
fastcgi_param  SCRIPT_FILENAME    $request_filename;
```

Is it the same ?

rodrigoaguilera (approved):

rodrigoaguilera:
The SCRIPT_FILENAME un fastcgi.conf is more complete and suites for more complex apps such as Drupal, WordPress or Laravel.

The variation that only uses $request_filename might be enough for apps that work with a single index.php without request query arguments.
http://nginx.org/en/docs/http/ngx_http_core_module.html#variables


Using $fastcgi_script_name
https://nginx.org/en/docs/http/ngx_http_fastcgi_module.html

rodrigoaguilera (approved):
Looks good to me.

buchdag:
Thanks for the explanation, seems safe to merge to me 👍

PR #2450, opened 2024-05-16
fix: enforce nohhtp on missing cert as well

pini-gh:
There is no reason to silently switch from nohhtp to noredirect on missing cert.


pini-gh:
Superseded by #2452 .

PR #2452, opened 2024-05-16
fix: enforce HTTPS_METHOD on missing cert as well

pini-gh:
There is no reason to silently switch to noredirect on missing cert.

The default being `HTTPS_METHOD=redirect` most of the test cases need to set `HTTPS_METHOD=nohttps` or `HTTPS_METHOD=noredirect`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -582,6 +582,7 @@ If no matching certificate is found for a given virtual host, nginx-proxy will:
 
 - configure nginx to use the default certificate (`default.crt` with `default.key`) and return a 500 error for HTTPS,
 - force enable HTTP; i.e. `HTTPS_METHOD` will switch to `noredirect` if it was set to `nohttp` or `redirect`.
+  If this switch to HTTP is not wanted set `ENABLE_HTTP_ON_MISSING_CERT=false` (default is `true`).
 
 If the default certificate is also missing, nginx-proxy will configure nginx to accept HTTPS connections but fail the TLS negotiation. Client browsers will render a TLS error page. As of March 2023, web browsers display the following error messages:
 
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -590,8 +590,9 @@ proxy_set_header Proxy "";
 
     {{- $default := eq $globals.Env.DEFAULT_HOST $hostname }}
     {{- $https_method := or (first (groupByKeys $vhost_containers "Env.HTTPS_METHOD")) $globals.Env.HTTPS_METHOD "redirect" }}
+    {{- $enable_http_on_missing_cert := parseBool (or (first (groupByKeys $vhost_containers "Env.ENABLE_HTTP_ON_MISSING_CERT")) $globals.Env.ENABLE_HTTP_ON_MISSING_CERT "true") }}
     {{- /* When the certificate is missing we want to ensure that HTTP is enabled; hence switching from 'nohttp' or 'redirect' to 'noredirect' */}}
-    {{- if (and (not $cert_ok) (or (eq $https_method "nohttp") (eq $https_method "redirect"))) }}
+    {{- if (and $enable_http_on_missing_cert (not $cert_ok) (or (eq $https_method "nohttp") (eq $https_method "redirect"))) }}
         {{- $https_method = "noredirect" }}
     {{- end }}
     {{- $http2_enabled := parseBool (or (first (keys (groupByLabel $vhost_containers "com.github.nginx-proxy.nginx-proxy.http2.enable"))) $globals.Env.ENABLE_HTTP2 "true")}}
--- a/test/test_enable_http_on_missing_cert.py
+++ b/test/test_enable_http_on_missing_cert.py
@@ -1,0 +1,18 @@
+import pytest
+
+
+def test_nohttp_missing_cert_disabled(docker_compose, nginxproxy):
+    r = nginxproxy.get("http://nohttp-missing-cert-disabled.nginx-proxy.tld/", allow_redirects=False)
+    assert r.status_code == 503
+
+def test_nohttp_missing_cert_enabled(docker_compose, nginxproxy):
+    r = nginxproxy.get("http://nohttp-missing-cert-enabled.nginx-proxy.tld/", allow_redirects=False)
+    assert r.status_code == 200
+
+def test_redirect_missing_cert_disabled(docker_compose, nginxproxy):
+    r = nginxproxy.get("http://redirect-missing-cert-disabled.nginx-proxy.tld/", allow_redirects=False)
+    assert r.status_code == 301
+
+def test_redirect_missing_cert_enabled(docker_compose, nginxproxy):
+    r = nginxproxy.get("http://redirect-missing-cert-enabled.nginx-proxy.tld/", allow_redirects=False)
+    assert r.status_code == 200
--- a/test/test_enable_http_on_missing_cert.yml
+++ b/test/test_enable_http_on_missing_cert.yml
@@ -1,0 +1,46 @@
+version: "2"
+
+services:
+  sut:
+    image: nginxproxy/nginx-proxy:test
+    volumes:
+      - /var/run/docker.sock:/tmp/docker.sock:ro
+      - ./withdefault.certs:/etc/nginx/certs:ro
+    environment:
+      ENABLE_HTTP_ON_MISSING_CERT: "false"
+
+  nohttp-missing-cert-disabled:
+    image: web
+    expose:
+      - "81"
+    environment:
+      WEB_PORTS: "81"
+      VIRTUAL_HOST: nohttp-missing-cert-disabled.nginx-proxy.tld
+      HTTPS_METHOD: nohttp
+
+  nohttp-missing-cert-enabled:
+    image: web
+    expose:
+      - "82"
+    environment:
+      WEB_PORTS: "82"
+      VIRTUAL_HOST: nohttp-missing-cert-enabled.nginx-proxy.tld
+      HTTPS_METHOD: nohttp
+      ENABLE_HTTP_ON_MISSING_CERT: "true"
+
+  redirect-missing-cert-disabled:
+    image: web
+    expose:
+      - "83"
+    environment:
+      WEB_PORTS: "83"
+      VIRTUAL_HOST: redirect-missing-cert-disabled.nginx-proxy.tld
+
+  redirect-missing-cert-enabled:
+    image: web
+    expose:
+      - "84"
+    environment:
+      WEB_PORTS: "84"
+      VIRTUAL_HOST: redirect-missing-cert-enabled.nginx-proxy.tld
+      ENABLE_HTTP_ON_MISSING_CERT: "true"


pini-gh:
> `test_dockergen/test_dockergen_v3.py::test_forwards_to_whoami FAILED      [  9%]`

@buchdag I don't understand this failure. All the tests are successful on my box, and this very same test passed with the alpine version. Can we try a relaunch?

buchdag:
Test restarted.

I got and identical unexplained failure of the same test yesterday : https://github.com/nginx-proxy/nginx-proxy/actions/runs/9101618380/job/25019246589 😒

buchdag:
@pini-gh If we're updating a behaviour of nginx-proxy, shouldn't we update the test assertions rather than the test setups ?

pini-gh:
I don't think so. The related tests use HTTP URLs, while default `HTTPS_METHOD` is `redirect`, i.e. HTTP cannot be tested.

To explicitly enable HTTP mode we need to set `HTTPS_METHOD=nohttps` ou `HTTPS_METHOD=noredirect`.

The other way around would be to configure valid certificates for all these tests, which seems cumbersome to me.

buchdag:
Okay, got it. I'd prefer `HTTPS_METHOD=noredirect` where possible because it's kinda "less specific" than `HTTPS_METHOD=nohttps`, as in we're not specifically testing the absence of an HTTPS listener here. Does that make sense ?

pini-gh:
To me `HTTPS_METHOD=nohttps` is more appropriate as it states explicitly that the current test doesn't involve HTTPS at all. Moreover, these tests don't have valid certificates, so HTTPS wouldn't work anyway.

But I won't object setting `noredirect` if you stick to it. You tell me.

pini-gh:
Because this PR may change the behavior of existing setups I could add an environment variable `ENFORCE_HTTPS_METHOD` with default value `false`. What do you think?

buchdag:
After carefully re-reading and testing this PR I'm afraid we're going in the wrong direction, we actually _do_ want to silently switch to `noredirect` when there is no certificate present for a given vhost, because `HTTPS_METHOD` should only affect vhost that have HTTPS enabled (ie that have a certificate):

- HTTPS not enabled -> `noredirect` and `HTTPS_METHOD` has no effect
- HTTPS enabled -> `redirect` by default, overridden by `HTTPS_METHOD`

As you remarked in your previous message, changing this to make the proxy enforce `HTTPS_METHOD` on HTTP only vhost would break pretty much every HTTP only config. I think I misled you in  #2446 and drew an invalid parallel between the behaviour of `nohttp` and the breaking changes we made to `VIRTUAL_HOST` a few years ago.

To get a better view of the proxy behaviour and see if things need to be fixed, I tried to test it a bit more exhaustively and in an easier to understand way than in the `fallback` test:

<details>
<summary>Unknown vhosts</summary>

`SSL_ERROR_SYSCALL` is what I got testing with curl on my host, I'm not certain how to interpret it or why I get this when at least one HTTP / HTTPS vhost is running.

| default certificate | at least one HTTP vhost | at least one HTTPS vhost | HTTPS Method (on proxy) | HTTP response | HTTPS response    | Certificate served |
| :-----------------: | :---------------------: | :----------------------: | ----------------------- | ------------- | ----------------- | ------------------ |
|         ❌          |           ❌            |            ❌            | not set                 | `503`         | `444` (TLS Error) | _n/a_              |
|         ❌          |           ❌            |            ❌            | `redirect`              | `503`         | `444` (TLS Error) | _n/a_              |
|         ❌          |           ❌            |            ❌            | `noredirect`            | `503`         | `444` (TLS Error) | _n/a_              |
|         ❌          |           ❌            |            ❌            | `nohttp`                | `503`         | `444` (TLS Error) | _n/a_              |
|         ❌          |           ❌            |            ❌            | `nohttps`               | `503`         | `444` (TLS Error) | _n/a_              |
|                     |                         |                          |                         |               |                   |                    |
|         🟢          |           ❌            |            ❌            | not set                 | `503`         | `503`             | default            |
|         🟢          |           ❌            |            ❌            | `redirect`              | `503`         | `503`             | default            |
|         🟢          |           ❌            |            ❌            | `noredirect`            | `503`         | `503`             | default            |
|         🟢          |           ❌            |            ❌            | `nohttp`                | `503`         | `503`             | default            |
|         🟢          |           ❌            |            ❌            | `nohttps`               | `503`         | `503`             | default            |
|                     |                         |                          |                         |               |                   |                    |
|         ❌          |           🟢            |            ❌            | not set                 | `503`         | `444` (TLS Error) | _n/a_              |
|         ❌          |           🟢            |            ❌            | `redirect`              | `503`         | `444` (TLS Error) | _n/a_              |
|         ❌          |           🟢            |            ❌            | `noredirect`            | `503`         | `444` (TLS Error) | _n/a_              |
|         ❌          |           🟢            |            ❌            | `nohttp`                | `503`         | `444` (TLS Error) | _n/a_              |
|         ❌          |           🟢            |            ❌            | `nohttps`               | `503`         | SSL_ERROR_SYSCALL | _n/a_              |
|                     |                         |                          |                         |               |                   |                    |
|         🟢          |           🟢            |            ❌            | not set                 | `503`         | `503`             | default            |
|         🟢          |           🟢            |            ❌            | `redirect`              | `503`         | `503`             | default            |
|         🟢          |           🟢            |            ❌            | `noredirect`            | `503`         | `503`             | default            |
|         🟢          |           🟢            |            ❌            | `nohttp`                | `503`         | `503`             | default            |
|         🟢          |           🟢            |            ❌            | `nohttps`               | `503`         | SSL_ERROR_SYSCALL | _n/a_              |
|                     |                         |                          |                         |               |                   |                    |
|         ❌          |           🟢            |            🟢            | not set                 | `503`         | `444` (TLS Error) | _n/a_              |
|         ❌          |           🟢            |            🟢            | `redirect`              | `503`         | `444` (TLS Error) | _n/a_              |
|         ❌          |           🟢            |            🟢            | `noredirect`            | `503`         | `444` (TLS Error) | _n/a_              |
|         ❌          |           🟢            |            🟢            | `nohttp`                | `503`         | `444` (TLS Error) | _n/a_              |
|         ❌          |           🟢            |            🟢            | `nohttps`               | `503`         | SSL_ERROR_SYSCALL | _n/a_              |
|                     |                         |                          |                         |               |                   |                    |
|         🟢          |           🟢            |            🟢            | not set                 | `503`         | `503`             | default            |
|         🟢          |           🟢            |            🟢            | `redirect`              | `503`         | `503`             | default            |
|         🟢          |           🟢            |            🟢            | `noredirect`            | `503`         | `503`             | default            |
|         🟢          |           🟢            |            🟢            | `nohttp`                | `503`         | `503`             | default            |
|         🟢          |           🟢            |            🟢            | `nohttps`               | `503`         | SSL_ERROR_SYSCALL | _n/a_              |
</details>

<details>
<summary>Known vhosts</summary>

The only "issue" I see here is `nohttps` behaving differently in the absence and presence of a default certificate.

| default certificate | vhost certificate | HTTPS Method | HTTP response | HTTPS response | Certificate served |
| :-----------------: | :---------------: | ------------ | ------------- | -------------- | ------------------ |
|         ❌          |        ❌         | not set      | `200` ✅      | `444` ⛔️      | _n/a_              |
|         ❌          |        ❌         | `redirect`   | `200` ✅      | `444` ⛔️      | _n/a_              |
|         ❌          |        ❌         | `noredirect` | `200` ✅      | `444` ⛔️      | _n/a_              |
|         ❌          |        ❌         | `nohttp`     | `200` ✅      | `444` ⛔️      | _n/a_              |
|         ❌          |        ❌         | `nohttps`    | `200` ✅      | `444` ⛔️      | _n/a_              |
|                     |                   |              |               |                |                    |
|         🟢          |        ❌         | not set      | `200` ✅      | `500` ⛔️      | default            |
|         🟢          |        ❌         | `redirect`   | `200` ✅      | `500` ⛔️      | default            |
|         🟢          |        ❌         | `noredirect` | `200` ✅      | `500` ⛔️      | default            |
|         🟢          |        ❌         | `nohttp`     | `200` ✅      | `500` ⛔️      | default            |
|         🟢          |        ❌         | `nohttps`    | `200` ✅      | `503` ⛔️      | default            |
|                     |                   |              |               |                |                    |
|         ❌          |        🟢         | not set      | `301` ➡️      | `200` ✅       | vhost              |
|         ❌          |        🟢         | `redirect`   | `301` ➡️      | `200` ✅       | vhost              |
|         ❌          |        🟢         | `noredirect` | `200` ✅      | `200` ✅       | vhost              |
|         ❌          |        🟢         | `nohttp`     | `503` ⛔️     | `200` ✅       | vhost              |
|         ❌          |        🟢         | `nohttps`    | `200` ✅      | `444` ⛔️      | _n/a_              |
|                     |                   |              |               |                |                    |
|         🟢          |        🟢         | not set      | `301` ➡️      | `200` ✅       | vhost              |
|         🟢          |        🟢         | `redirect`   | `301` ➡️      | `200` ✅       | vhost              |
|         🟢          |        🟢         | `noredirect` | `200` ✅      | `200` ✅       | vhost              |
|         🟢          |        🟢         | `nohttp`     | `503` ⛔️     | `200` ✅       | vhost              |
|         🟢          |        🟢         | `nohttps`    | `200` ✅      | `503` ⛔️      | default            |
</details>

Let me know if you see other behaviours that seem incorrect or if I forgot some cases.

pini-gh:
> `HTTPS_METHOD` should only affect vhost that have HTTPS enabled (ie that have a certificate)

That's where I'd want a different behavior. A certificate going missing could be a bug or a mistake. For my sites I don't want HTTP silently enabled on such cases. Then I need a way to enforce `noredirect` or `nohttp`.

There are several ways to achieve this. If we don't want to change the current default behavior I see two ways:
1. Guard the change with environment variable `ENFORCE_HTTPS_METHOD`
2. Add new values for `HTTPS_METHOD`:
    * `noredirect_enforced`
    * `nohttp_enforced`

Would you agree on one of these?

pini-gh:
I've created PR #2465 to make the current behavior explicit and documented.

buchdag:
> That's where I'd want a different behavior. A certificate going missing could be a bug or a mistake. For my sites I don't want HTTP silently enabled on such cases.

Ok I totally get this use case 👍

Ideally I'd prefer if we did not introduce an entirely new environment variable (to avoid adding one dimension to the previous behaviour tables), so `redirect_enforced` / `nohttp_enforced`.

pini-gh:
May I insist on a using a new variable? Say `ENABLE_HTTP_ON_MISSING_CERT` (default `true`)? Because it precisely conveys what it is for. While `redirect_enforced` and `nohttp_enforced` look like poor workarounds compared to it.

The new variable doesn't add a full new dimension. In both cases we just need a couple of tests where we check that HTTP is not enabled on missing cert when `HTTPS_METHOD=redirect` or `HTTPS_METHOD=nohttp`.

buchdag:
I guess I really liked my tables above ^^

Let's roll with `ENABLE_HTTP_ON_MISSING_CERT`, defaulting to `true`.

docs/README.md
```diff
@@ -579,7 +579,7 @@ _WARNING_: HSTS will force your users to visit the HTTPS version of your site fo
 
 If no matching certificate is found for a given virtual host, nginx-proxy will:
 * configure nginx to use the default certificate (`default.crt` with `default.key`) and return a 500 error for HTTPS,
-* force enable HTTP; i.e. `HTTPS_METHOD` will switch to `noredirect` if it was set to `nohttp` or `redirect`.
+* if `ENABLE_HTTP_ON_MISSING_CERT=true` (default) force enable HTTP; i.e. `HTTPS_METHOD` will switch to `noredirect` if it was set to `nohttp` or `redirect`.
```
buchdag, line 582: I'm not certain formulating it this way really highlight the use case for the new environment variable.

If you're not in a hurry I can try to come up with something.

buchdag (dismissed):

buchdag (approved):

PR #2453, opened 2024-05-17
Draft: feat: var BACKUP_SERVER to tag server as backup

pini-gh:
Related to nginx-proxy/acme-companion#1062


SchoNie:
Would it be an idea to use a more generic variable to set parameters of a server?
For example in my setup I need to specify `weight=5;`
And somebody in #2403 needed `fail_timeout=30s;`

Something like `SERVER_PARAMETERS` which can contain any parameter like `backup, weight, fail_timeout` (or a combination) from https://nginx.org/en/docs/http/ngx_http_upstream_module.html#server

PR #2465, opened 2024-05-27
chore/doc: explicit policy on missing certificate

pini-gh:
This doesn't change the current nginx-proxy behavior, but makes explicit the current HTTPS_METHOD policy on missing certificate.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -572,7 +572,9 @@ _WARNING_: HSTS will force your users to visit the HTTPS version of your site fo
 
 ### Missing Certificate
 
-If HTTPS is enabled for a virtual host but its certificate is missing, nginx-proxy will configure nginx to use the default certificate (`default.crt` with `default.key`) and return a 500 error.
+If HTTPS is enabled for a virtual host but its certificate is missing, nginx-proxy will:
+* configure nginx to use the default certificate (`default.crt` with `default.key`) and return a 500 error,
+* force enable HTTP; i.e. `HTTPS_METHOD` will switch to `noredirect` if it was set to `nohttp` or `redirect`.
 
 If the default certificate is also missing, nginx-proxy will configure nginx to accept HTTPS connections but fail the TLS negotiation. Client browsers will render a TLS error page. As of March 2023, web browsers display the following error messages:
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -572,7 +572,7 @@ _WARNING_: HSTS will force your users to visit the HTTPS version of your site fo
 
 ### Missing Certificate
 
-If no matching certificate is found for a given virtual, nginx-proxy will:
+If no matching certificate is found for a given virtual host, nginx-proxy will:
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -572,7 +572,9 @@ _WARNING_: HSTS will force your users to visit the HTTPS version of your site fo
 
 ### Missing Certificate
 
-If HTTPS is enabled for a virtual host but its certificate is missing, nginx-proxy will configure nginx to use the default certificate (`default.crt` with `default.key`) and return a 500 error.
+If no matching certificate is found for a given virtual, nginx-proxy will:
```
buchdag, line 575: I managed to suggest a typo 👏 
```suggestion
If no matching certificate is found for a given virtual host, nginx-proxy will:
```

buchdag (approved):

ConnorNelson:
I use nginx-proxy, and acme-companion. In my production setup, both run successfully. In my development setup, I allow acme-companion to fail to generate a certificate (DNS isn't pointed at CI). This worked great, my development setup would just stick to HTTP, and I didn't need to support any special-case logic.

Now, I get 503s. This seems to be coming from:
```
# nginx -s reload
2024/05/28 23:40:30 [emerg] 305#305: duplicate location "/.well-known/acme-challenge/" in /etc/nginx/vhost.d/default:2
nginx: [emerg] duplicate location "/.well-known/acme-challenge/" in /etc/nginx/vhost.d/default:2
```

Inside `/etc/nginx/conf.d/default.conf`:
```
...
server {
    server_name localhost.pwn.college;
    http2 on;
    listen 80 default_server;
    location /.well-known/acme-challenge/ {
        auth_basic off;
        allow all;
        root /usr/share/nginx/html;
        try_files $uri =404;
        break;
    }
    # No certificate found for this vhost, so use the default certificate and
    # return an error code if the user connects via https.
    ssl_certificate /etc/nginx/certs/default.crt;
    ssl_certificate_key /etc/nginx/certs/default.key;
    if ($https) {
        return 500;
    }
    location / {
        proxy_pass http://localhost.pwn.college;
        set $upstream_keepalive false;
    }
}
```

Inside `/etc/nginx/vhost.d/default`:
```
## Start of configuration add by letsencrypt container
location ^~ /.well-known/acme-challenge/ {
    auth_basic off;
    auth_request off;
    allow all;
    root /usr/share/nginx/html;
    try_files $uri =404;
    break;
}
## End of configuration add by letsencrypt container
```

It seems like this (coupled with #2446) has broken my use-case.
Pinning from `:latest` to `1.5` has resolved this issue.

pini-gh:
> It seems like this (coupled with #2446) has broken my use-case.

Now that nginx-proxy handles ACME HTTP challenge correctly I guess acme-companion should be updated so it doesn't add this configuration when it already exists.

In the meantime - as it seems to be a [requested feature](https://github.com/nginx-proxy/nginx-proxy/pull/2446#discussion_r1595904704) - I'd like to guard nginx-proxy ACME HTTP Challenge configuration by an environment variable. I'll propose a PR on this topic soon.

PR #2468, opened 2024-05-29
feat: boolean variable ACME_HTTP_CHALLENGE

pini-gh:
Enable / disable ACME HTTP Challenge blocks generation by nginx-proxy.

Default: true.

This feature is currently needed because acme-companion may generate the HTTP Challenge configuration while it was done already by nginx-proxy (see #2465#issuecomment-2136361373).

Also sometimes a hardcoded ACME challenge location is not wanted because the challenge validation is not done with acme-companion / Let's Encrypt, and with a challenge location setup differently.


buchdag:
@pini-gh the variable should be set to `false` by default, and the condition on line 746 and 756 (on the HTTP redirect server) removed. The idea is to disable the new behaviour introduced in #2446 by default, and it alone.

The priority for now is to avoid breaking existing setups (even if they insist on using `latest`) without forcing people to use a new environment variable, we'll revisit this soon when the ability to disable automatic location config will be added to acme-companion and we can release a version of both that are compatible with each other out of the box.

nginx.tmpl
```diff
@@ -596,6 +596,7 @@ proxy_set_header Proxy "";
     {{- end }}
     {{- $http2_enabled := parseBool (or (first (keys (groupByLabel $vhost_containers "com.github.nginx-proxy.nginx-proxy.http2.enable"))) $globals.Env.ENABLE_HTTP2 "true")}}
     {{- $http3_enabled := parseBool (or (first (keys (groupByLabel $vhost_containers "com.github.nginx-proxy.nginx-proxy.http3.enable"))) $globals.Env.ENABLE_HTTP3 "false")}}
+    {{- $acme_http_challenge_enabled := parseBool (or (first (groupByKeys $vhost_containers "Env.ACME_HTTP_CHALLENGE")) $globals.Env.ACME_HTTP_CHALLENGE "true") }}
```
buchdag, line 599:
```suggestion
    {{- $acme_http_challenge_enabled := parseBool (or (first (groupByKeys $vhost_containers "Env.ACME_HTTP_CHALLENGE")) $globals.Env.ACME_HTTP_CHALLENGE "false") }}
```

nginx.tmpl
```diff
@@ -741,6 +743,7 @@ server {
     listen [::]:{{ $globals.external_http_port }} {{ $default_server }};
         {{- end }}
 
+        {{- if $vhost.acme_http_challenge_enabled }}
```
buchdag, line 746:
```suggestion
```

nginx.tmpl
```diff
@@ -750,6 +753,7 @@ server {
         try_files $uri =404;
         break;
     }
+        {{- end }}
```
buchdag, line 756:
```suggestion
```

docs/README.md
```diff
@@ -421,6 +421,8 @@ If you are running the container in a virtualized environment (Hyper-V, VirtualB
 
 [acme-companion](https://github.com/nginx-proxy/acme-companion) is a lightweight companion container for the nginx-proxy. It allows the automated creation/renewal of SSL certificates using the ACME protocol.
 
+By default nginx-proxy generates location blocks to handle ACME HTTP Challenge. To disable this behavior set `ACME_HTTP_CHALLENGE=false` (default `true`) at the nginx-proxy container level or at the app container one.
+
```
buchdag, line 425: Either we choose to not document this feature for now because it won't be complete, or we alter the text, for example like this:
```suggestion
By default nginx-proxy generates location blocks to avoid redirecting ACME HTTP challenge to HTTPS when using `HTTPS_METHOD=redirect` (the default value). To additionally generate location blocks for HTTPS, set `ACME_HTTP_CHALLENGE=true` (default `false`) at the nginx-proxy container level or at the app container one.
```

buchdag (changes requested):

docs/README.md
pini-gh: There are broken use cases with both defaults:

**true**
acme_companion configures an extraneous HTTP Challenge location block which is not valid.

**false**
Use cases where certificates are not generated with acme_companion loose HTTP Challenge configuration and may fail to renew their certificates.

Why not fixing acme_companion instead?

nginx.tmpl
pini-gh: There are broken use cases with both defaults:

**true**
acme_companion configures an extraneous HTTP Challenge location block which is not valid.

**false**
Use cases where certificates are not generated with acme_companion loose HTTP Challenge configuration and may fail to renew their certificates.

Why not fixing acme_companion instead?

EDIT: The easiest way would be to keep the default as `true` and update acme_companion so it wouldn't generate the HTTP Challenge configuration by default (but for standalone certificates).

EDIT1: moving this comment with was on the wrong thread.

pini-gh: How about having 3 possible values for `ACME_HTTP_CHALLENGE`?
* `legacy`: previous incomplete behavior
* `true`: generate ACME HTTP Challenge location blocks when `HTTPS_METHOD` is in [ `redirect`, `noredirect` ] (current default behavior) 
* `false`: do not generate ACME HTTP Challenge location blocks

And we set the new default to `legacy`.

buchdag: The `legacy` / `true` /  `false` solution with `legacy` being the default (until we change acme-companion default behaviour, which I plan on doing soon) seems perfect to me, if it's not too much additional work for you.

> Why not fixing acme_companion instead?

Because I'm a bit short in time this week 😬 but that's planned, yes.

(by the way regarding acme-companion and it's location tinkering, I tried to automatically skip location configuration when not required a few years ago, I could not come up with anything else than very hacky and fragile solutions that did not even work in the end. Manually enabling or disabling location configuration as you suggest seem to be the only realistic way).

buchdag: Minor request: could we use `ACME_HTTP_CHALLENGE_LOCATION` instead of `ACME_HTTP_CHALLENGE` ?

pini-gh: Done.

buchdag (approved):

Montana:
Hi @pini-gh,

Just an idea here for the location block: 

Perhaps set `ACME_ENABLED` to `false` to include the ACME challenge location block. Then set the Production Environment: `ACME_ENABLED` to `true` to exclude the ACME challenge location block.

So you would have a `default.conf` that would check for an `env var` assigned `ACME_ENABLED` this would determine to either include the ACME challenge block or not. This is a `default.conf` I made:

```nginx
server {
    server_name localhost.pwn.college;
    http2 on;
    listen 80 default_server;

    {{ if eq (or (index $globals.Env "ACME_ENABLED") "false") "false" }}
]    location /.well-known/acme-challenge/ {
        auth_basic off;
        allow all;
        root /usr/share/nginx/html;
        try_files $uri =404;
        break;
    }
    {{ end }}

    ssl_certificate /etc/nginx/certs/default.crt;
    ssl_certificate_key /etc/nginx/certs/default.key;
    if ($https) {
        return 500;
    }

    location / {
        proxy_pass http://localhost.pwn.college;
        set $upstream_keepalive false;
    }
}
```

Setting these environment variables, I'm obviously talking about Docker Compose:

```yaml
environment:
      - ACME_ENABLED=false # set to true in production
```
This is only an idea, maybe it helps or gets something stirring. Have a wonderful Thursday.

pini-gh:
> Perhaps set `ACME_ENABLED` to `false` to include the ACME challenge location block. Then set the Production Environment: `ACME_ENABLED` to `true` to exclude the ACME challenge location block.
> 
> So you would have a `default.conf` that would check for an `env var` assigned `ACME_ENABLED` this would determine to either include the ACME challenge block or not.

All the uses cases are enabled by the current PR already:
* legacy behavior
* generate ACME HTTP Challenge location blocks
* do not generate ACME HTTP Challenge location blocks

PR #2494, opened 2024-07-28
Limit exposed ports in template to 10

mikfar:
This fixes the problem described in [this issue](https://github.com/nginx-proxy/nginx-proxy/issues/2492) where containers that expose a large number of ports will cause the template to render a line exceeding the max line length of Nginx.

### DIFF
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -125,7 +125,7 @@
      */}}
 {{- define "container_port" }}
     {{- /* If only 1 port exposed, use that as a default, else 80. */}}
-    #     exposed ports:{{ range sortObjectsByKeysAsc $.container.Addresses "Port" }} {{ .Port }}/{{ .Proto }}{{ else }} (none){{ end }}
+    #     exposed ports:{{ $maxPorts := 10 }}{{ $count := 0 }}{{ range sortObjectsByKeysAsc $.container.Addresses "Port" }}{{ if lt $count $maxPorts }} {{ .Port }}/{{ .Proto }}{{ $count = add $count 1 }}{{ end }}{{ else }} (none){{ end }}
     {{- $default_port := when (eq (len $.container.Addresses) 1) (first $.container.Addresses).Port "80" }}
     #     default port: {{ $default_port }}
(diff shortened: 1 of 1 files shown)


nginx.tmpl
```diff
@@ -125,7 +125,7 @@
      */}}
 {{- define "container_port" }}
     {{- /* If only 1 port exposed, use that as a default, else 80. */}}
-    #     exposed ports:{{ range sortObjectsByKeysAsc $.container.Addresses "Port" }} {{ .Port }}/{{ .Proto }}{{ else }} (none){{ end }}
+    #     exposed ports:{{ $maxPorts := 10 }}{{ $count := 0 }}{{ range sortObjectsByKeysAsc $.container.Addresses "Port" }}{{ if lt $count $maxPorts }} {{ .Port }}/{{ .Proto }}{{ $count = add $count 1 }}{{ end }}{{ else }} (none){{ end }}
```
buchdag, line 128: I think it can be expressed a bit more simply, this should work:

```suggestion
    #     exposed ports:{{ range $index, $address := (sortObjectsByKeysAsc $.container.Addresses "Port") }}{{ if lt $index 10 }} {{ $address.Port }}/{{ $address.Proto }}{{ end }}{{ else }} (none){{ end }}
```

buchdag (changes requested):

nginx.tmpl
```diff
@@ -125,7 +125,7 @@
      */}}
 {{- define "container_port" }}
     {{- /* If only 1 port exposed, use that as a default, else 80. */}}
-    #     exposed ports:{{ range sortObjectsByKeysAsc $.container.Addresses "Port" }} {{ .Port }}/{{ .Proto }}{{ else }} (none){{ end }}
+    #     exposed ports:{{ range $index, $address := (sortObjectsByKeysAsc $.container.Addresses "Port") }}{{ if lt $index 10 }} {{ $address.Port }}/{{ $address.Proto }}{{ end }}{{ else }} (none){{ end }}
```
buchdag, line 128:
```suggestion
    #     exposed ports (first ten):{{ range $index, $address := (sortObjectsByKeysAsc $.container.Addresses "Port") }}{{ if lt $index 10 }} {{ $address.Port }}/{{ $address.Proto }}{{ end }}{{ else }} (none){{ end }}
```

buchdag (approved):

PR #2510, opened 2024-09-13
fix: correctly enable TLSv1 and TLSv1.1

liuxiaoy:


buchdag:
@liuxiaoy hi, could you provide a bit more context on the PR and explanations on the fix ?

liuxiaoy:
Hi, I want my site support TLSv1 TLSv1.1 TLSv1.2 TLSv1.3.
Then I use environment SSL_POLICY=Mozilla-Old. 
But the openssl disabled TLSv1 TLSv1.1 default, so we need append :@SECLEVEL=0' to ssl_ciphers in nginx to support TLSv1 really.

liuxiaoy:
https://github.com/nginxinc/docker-nginx/issues/858#issuecomment-1894574478

buchdag:
Thanks for the links 👍 

However, I'm not able to confirm that this work locally using either `nmap --script ssl-enum-ciphers`, `openssl s_client` or `curl`.

Have you used it and confirmed that it does work, and if so how ?

liuxiaoy:
I have test it with the shell below:
```shell
curl --tls-max 1.0 https://your.com -Ik
# or
openssl s_client -connect your.com:443 -tls1
```

or test on the site online like  `https://www.ssllabs.com/ssltest/analyze.html?d=your.com&hideResults=on`

liuxiaoy:
I used to use a lower version of nginx to solve the tlsv1 problem, which was v1.12. Later, in order to automate the certificate renewal, I found some discussions on this topic online. After experiments, I found that a higher version could also solve the problem, so I proposed this PR

buchdag:
I think I identified why it can't validate it, see https://github.com/nginxinc/docker-nginx/issues/858#issuecomment-2379190004 for details, I'll think of a way we could handle this.

buchdag:
@liuxiaoy regarding your message on https://github.com/nginxinc/docker-nginx/issues/858 : nginx-proxy is built upon the upstream nginx Docker images, so if you can't enable TLSv1 / TLSv1.1 in the base nginx image, you will have the same issue in nginx-proxy, that's not something specific to the later.

buchdag:
The issue comes from the [`ssl_protocols`](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_protocols) directive : when used in the `http` context, it will limit what protocols are available in the `server` context.

Currently [we set `ssl_protocols` and `ssl_ciphers` in the `http` context](https://github.com/nginx-proxy/nginx-proxy/blob/main/nginx.tmpl#L439) with the Mozilla-Intermediate SSL policy values.

We can either permanently replace this with the Mozilla-Old values, or dynamically use the Mozilla-Old values when at least one container use an `SSL_POLICY` that has TLSv1 and/or TLSv1.1 enabled. I'm not sure the later is required, since we're always manually setting `ssl_ciphers` in the `server` context, those "default" values are actually never used for anything else than the fallback listener.

buchdag:
@liuxiaoy could try image `nginxproxy/nginx-proxy:2510` ?

[It's based on this change](https://github.com/nginx-proxy/nginx-proxy/compare/main...tlsv1) and should correctly enable TLSv1 / TLSv1.1

buchdag:
@liuxiaoy I took the liberty to to update your PR.

buchdag (approved):

liuxiaoy:
@buchdag Sorry for forgetting this. I have tried the ```nginxproxy/nginx-proxy:2510``` and after deployment, I tested it successfully using curl --tls-max 1.0. The online website also tested that This server supports TLS 1.0 and TLS 1.1. Grade capped to B. 

If you can't test it there, we can communicate so that you can test it successfully.

buchdag:
I was able to confirm that it work on my side, but I wanted to be sure that it worked for you too.

The only limitation is the fact that it won't work with the Alpine variant of the image as I indicated in the docs (I could not manage to get it to work, if you have any idea on how to do it that would be welcome).

This isn't realistically testable at the moment so I'll merge it as-is.

Thanks for bringing the issue to light @liuxiaoy 👍

liuxiaoy:
@buchdag I have try the same on ` nginxproxy/nginx-proxy:1.6-alpine`，it also works；Just append  `:@SECLEVEL=0` using `vi /etc/nginx/conf.d/default.conf` then run `nginx -s reload` on the contaner.

I also try it ok on `nginx:1.27-alpine` with config below:
```
    server {
        listen 443 ssl;
        ssl_certificate /etc/ssl/localhost.crt;
        ssl_certificate_key /etc/ssl/localhost.key;
        ssl_ciphers DEFAULT:@SECLEVEL=0;
        location / { return 200 'OK - $ssl_protocol - $ssl_cipher\n'; }
    }
```

Test record
```
curl --tls-max 1.0 https://127.0.0.1:443/ -k                                                                                                                                                                                                                       
OK - TLSv1 - ECDHE-RSA-AES256-SHA
```

buchdag:
@liuxiaoy it does indeed work with the alpine version too, I probably mixed up something while testing.

Thanks for the heads up, I'll update the docs 👍

PR #2535, opened 2024-11-03
feat: trust default certificate

buchdag:
See https://github.com/nginx-proxy/nginx-proxy/discussions/2501

~~Making the default certificate unusable by virtual hosts and considered kinda like an "always bad" certificate was an unintended mistake, probably induced by the fact that acme-companion used to auto generate a self signed default certificate ([this has been disabled by default](https://github.com/nginx-proxy/acme-companion/pull/1157) recently).~~

~~The default certificate is and always was intended to be a legit fallback certificate used by vhosts when there is no vhost matching certificate.~~

Edit: that's not true, see below message.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 **/__pycache__/
 **/.cache/
 .idea/
+wip
--- a/docs/README.md
+++ b/docs/README.md
@@ -568,8 +568,8 @@ Complete list of policies available through the `SSL_POLICY` environment variabl
 The default behavior for the proxy when port 80 and 443 are exposed is as follows:
 
 - If a virtual host has a usable cert, port 80 will redirect to 443 for that virtual host so that HTTPS is always preferred when available.
-- If the virtual host does not have a usable cert, but `default.crt` and `default.key` exist, those will be used as the virtual host's certificate and the client browser will receive a 500 error.
-- If the virtual host does not have a usable cert, and `default.crt` and `default.key` do not exist, SSL handshake will be rejected (see [Missing Certificate](#missing-certificate) below).
+- If the virtual host does not have a usable cert, but `default.crt` and `default.key` exist, those will be used as the virtual host's certificate.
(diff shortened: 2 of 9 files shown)


buchdag:
Well looking back at the docs and template history, the intent for the default certificate might not be as straightforward as I thought. The doc always stated that a vhost using the default certificate would respond with an error on HTTPS, first 503 then later 500, and the code is consistent with that.

Contrary to what I thought and stated (both in the first message and in #2501) this is not something that we broke and the default certificate was seemingly never considered a legit, useable certificate.

This make very little sense to me, given that the certificate has to be manually provided to nginx-proxy by the user like any other certificate, nginx-proxy should certainly not make the assumption that the certificate is invalid for any given vhost. But since this is the legacy behaviour, we probably need a way to keep it and optionally "trust" the default certificate.

buchdag:
@hardillb would you be interested in / have time to test this before we merge it ?

hardillb:
Will try and find some time later, but can't promise I'll get to it this week.

buchdag:
I've pushed two Docker images that contain this PR's changes:
- `nginxproxy/nginx-proxy:2535`
- `nginxproxy/nginx-proxy:2535-alpine`

#### Changes summary:
- `defautl.crt` is now trusted by default and used as a fallback certificate if no virtual host specific certificate was found.
- this behaviour can be reverted to the previous one (not trusting `default.crt`) either globally with the `TRUST_DEFAULT_CERT ` environment variable or per virtual host with the `com.github.nginx-proxy.nginx-proxy.trust-default-cert` label (both must be set to `false`).
- the `500` error when `default.crt` is not trusted and no virtual host specific certificate was found has been replaced with an SSL handshake fail, exactly like when there is no `default.crt` nor virtual host specific certificate.

hardillb:
Managed to throw together a quick test rig.

Looks good, works as described

buchdag:
Thanks @hardillb, I'll merge this by the end of the week end.

I know that we're changing a legacy behaviour here but I think that's for the better, as you pointed out in #2501 the feature was rather pointless in this state.

PR #2547, opened 2024-11-26
docs: remove experimental label for http loadbalance

SchoNie:
I think we can remove the experimental warning to clean the docs. No issues came up in #2195

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -281,13 +281,9 @@ If you use fastcgi,you can set `VIRTUAL_ROOT=xxx` for your root directory
 
 ### Upstream Server HTTP Load Balancing Support
 
-> **Warning**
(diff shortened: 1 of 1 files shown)


buchdag (approved):

buchdag:
Errr what's going on with the tests ?

SchoNie:
> Errr what's going on with the tests ?

I think because of this: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-including-paths

> If a workflow is skipped due to path filtering, [branch filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore), or a [commit message](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs), **then checks associated with that workflow will remain in a "Pending" state.** A pull request that requires those checks to be successful will be blocked from merging.

PR #2569, opened 2024-12-24
tests: fix, cleanup and restructure test code

buchdag:
This PR fixe various minor or code style issues in the test code base:
- do not remove all containers on host during test (I can't believe we did that, that was a very pleasant surprise ...)
- `report.longrepr` is now always a `ReprExceptionInfo`, see https://github.com/pytest-dev/pytest/issues/1316 and its fix
- `report.longrepr.addsection()` second parameter should be a `str`, but `container.logs()` returns `bytes`
- add `packaging` and `urllib3` as explicit dependencies
- `log.warn` (deprecated) -> `log.warning`
- type hints have been added to `conftest.py` where relevant
- remove unused imports from test files

The test files structure has also been changed so that we no longer have test files at the root of `test` and others in their own directory (all test files are in their own subdirectory now), and the name of the files themselves has been somewhat standardized.

@SchoNie if you happen to have the time (and since you're the only one regularly contributing test code), a review would be nice 👍


test/requirements/python-requirements.txt
```diff
@@ -1,4 +1,6 @@
 backoff==2.2.1
 docker==7.1.0
+packaging==24.2
 pytest==8.3.4
 requests==2.32.3
+urllib3==2.3.0
```
buchdag, line 6: `urllib3` is explicitly installed to use `HAS_IPV6` instead of doing the dubious `from requests.packages.urllib3.util.connection import HAS_IPV6`.

test/conftest.py
```diff
@@ -71,7 +75,7 @@ def __init__(self):
             self.session.verify = CA_ROOT_CERTIFICATE
 
     @staticmethod
-    def get_nginx_proxy_containers() -> List[Container]:
+    def get_nginx_proxy_container() -> Container:
```
buchdag, line 78: This method fails the test if the list's length is zero of more than one, so we're not really returning a list.

test/conftest.py
```diff
@@ -301,23 +298,23 @@ def get_nginx_conf_from_container(container):
         return conffile.read()
 
 
-def docker_compose_up(compose_file='docker-compose.yml'):
+def docker_compose_up(compose_file: str) -> None:
```
buchdag, line 301: The method is never called with a `None` parameter.

test/conftest.py
```diff
@@ -301,23 +298,23 @@ def get_nginx_conf_from_container(container):
...
     logging.info(f'{DOCKER_COMPOSE} -f {compose_file} up -d')
     try:
         subprocess.check_output(shlex.split(f'{DOCKER_COMPOSE} -f {compose_file} up -d'), stderr=subprocess.STDOUT)
     except subprocess.CalledProcessError as e:
-        pytest.fail(f"Error while runninng '{DOCKER_COMPOSE} -f {compose_file} up -d':\n{e.output}", pytrace=False)
+        pytest.fail(f"Error while running '{DOCKER_COMPOSE} -f {compose_file} up -d':\n{e.output}", pytrace=False)
 
 
-def docker_compose_down(compose_file='docker-compose.yml'):
+def docker_compose_down(compose_file: str) -> None:
```
buchdag, line 309: The method is never called with a `None` parameter.

test/conftest.py
```diff
@@ -59,10 +63,10 @@ def ipv6(force_ipv6=True):
     FORCE_CONTAINER_IPV6 = False
 
 
-class requests_for_docker(object):
+class RequestsForDocker:
```
buchdag, line 66: `class foo(object)` made sense in Python 2 only.

test/test_trust-downstream-proxy/test_default.py
```diff
@@ -1,6 +1,7 @@
-import pytest
 import re
```
SchoNie, line 2:
```suggestion

```

test/test_trust-downstream-proxy/test_disabled.py
```diff
@@ -1,6 +1,7 @@
-import pytest
 import re
```
SchoNie, line 2:
```suggestion

```

test/test_trust-downstream-proxy/test_enabled.py
```diff
@@ -1,6 +1,7 @@
-import pytest
 import re
```
SchoNie, line 2:
```suggestion

```

test/test_vhost-in-multiple-networks/test_vhost-in-multiple-networks.py
```diff
@@ -1,7 +1,9 @@
-import pytest
 import logging
 import time
```
SchoNie, line 2:
```suggestion
from time import sleep
```

test/test_vhost-in-multiple-networks/test_vhost-in-multiple-networks.py
```diff
@@ -1,7 +1,9 @@
...
```
SchoNie, line 3:
```suggestion

```

test/test_virtual-path/test_virtual-paths.py
```diff
@@ -3,6 +3,7 @@
 import pytest
```
SchoNie, line 3: Line 2 and maybe order

SchoNie (commented):
Nice work, lots of files so respect going trough!
Hope that `remove_all_containers` did not struck your machine. Good catch.
Just some minor styling comments, maybe you did that on purpose, then just discard my suggestions.

There is a [gitignore](https://github.com/nginx-proxy/nginx-proxy/blob/test/cleanup/test/test_dockergen/.gitignore) file which seems to have no purpose.

`WEB_PORTS` and `VIRTUAL_HOST` sometimes have quotations and other not. Maybe also a good moment to lint that?

buchdag:
@SchoNie thanks for the feedback 🙏  

Regarding the blank line between standard library and third party imports, that's what the PyCharm linter does by default, I think it's applying the [PEP8](https://peps.python.org/pep-0008/#imports) style guide:

> Imports should be grouped in the following order:
> 
> 1.  Standard library imports.
> 2.  Related third party imports.
> 3.  Local application/library specific imports.
>
> You should put a blank line between each group of imports.

> There is a [gitignore](https://github.com/nginx-proxy/nginx-proxy/blob/test/cleanup/test/test_dockergen/.gitignore) file which seems to have no purpose.

I've removed it.

> `WEB_PORTS` and `VIRTUAL_HOST` sometimes have quotations and other not. Maybe also a good moment to lint that?

I've cleaned `WEB_PORTS`, will do `VIRTUAL_HOST` later.

> Maybe also use pathlib here instead of os (`test/test_fallback/test_fallback.py`)

That's planned in the next PR 👍

SchoNie (approved):
Just 1 missed WEB_PORTS [test_debug-endpoint/test_global.yml Line 51](https://github.com/nginx-proxy/nginx-proxy/blob/ea99c1a6f900ceaad4c491e37ebc4cb090236ca1/test/test_debug-endpoint/test_global.yml#L51)

> Regarding the blank line between standard library and third party imports, that's what the PyCharm linter does by default, I think it's applying the [PEP8](https://peps.python.org/pep-0008/#imports) style guide:
> 
> > Imports should be grouped in the following order:
> > 
> > 1. Standard library imports.
> > 2. Related third party imports.
> > 3. Local application/library specific imports.
> > 
> > You should put a blank line between each group of imports.

Nice, didn't know that, learning everyday! 🧑‍🎓

PR #2570, opened 2024-12-30
tests: factor out base nginx-proxy config and enable local testing on macOS / Darwin

buchdag:
Okay, this one might be a bit hard to read.

Tests don't work on macOS / Darwin because [they all resolve the test domains to the IP of the nginx-proxy container](https://github.com/nginx-proxy/nginx-proxy/blob/4ccbc3edec423ca36364dd5bacbc4ea25b3d0be3/test/conftest.py#L248-L252) (thanks to `monkey_patch_urllib_dns_resolver()`), but [it's explicitly not possible to route traffic directly to containers with Docker Desktop](https://docs.docker.com/desktop/features/networking/#i-cannot-ping-my-containers).

The solution to this is "simply" [to use port forwarding](https://docs.docker.com/desktop/features/networking/#i-want-to-connect-to-a-container-from-the-host) and resolve the test domains to localhost's IP instead of the IP of the nginx-proxy container.

This in turn means port publishing must be added to the nginx-proxy container on almost every test compose file:
```yaml
[...]
    ports:
      - "80:80"
      - "443:443"
```

We're already duplicating the nginx-proxy base config on every compose file (with the container sometime named `sut`, sometime `nginx-proxy`, sometime something else ...) so I decided to change the way the `docker_compose` fixture work to reduce the quantity of boilerplate code.

By default it will use those files for every test module:
1. `test/compose.base.yml` (base nginx-proxy config reused for most tests)
2. `test/{test_dir}/compose.base.override.yml` if the file exist (override and/or containers for all tests in a given dir)
3. `test/{test_dir}/{test_module_name}.yml` (override and/or containers for a specific test module)

The [compose files are merged](https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/) in this order.

This automatic merge can be bypassed by using a file named `test/{test_dir}/{test_module_name}.base.yml`. When this file exist, it will be the only used by the test and no merge will automatically occur.

The `docker_compose` fixture also set the `PYTEST_MODULE_PATH` environment variable to the absolute path of the current test directory, so it can be used to mount files or directory relatives to the current test.

The support for test compose files with the `.yaml` extension has been removed because it did not serve any purpose.

The support for test compose files named `docker-compose.yml` that were treated as the default compose file for a whole test directory has also been removed because it wasn't really used.

I've never been fond of naming the nginx-proxy service [`sut`](http://xunitpatterns.com/SUT.html), so I've standardized to `nginx-proxy`.


buchdag:
Local run a MacBook Pro M1 with PyCharm:

pyCharm-tests

buchdag:
Ready to review if you have some spare time @SchoNie

test/test_fallback/test_fallback.data/custom-fallback.yml
```diff
@@ -1,6 +1,5 @@
 services:
-  sut:
-    image: nginxproxy/nginx-proxy:test
+  nginx-proxy:
     volumes:
       - /var/run/docker.sock:/tmp/docker.sock:ro
       - ./custom-fallback.conf:/etc/nginx/conf.d/zzz-custom-fallback.conf:ro
```
SchoNie, line 5:
```suggestion
      - ${PYTEST_MODULE_PATH}/custom-fallback.conf:/etc/nginx/conf.d/zzz-custom-fallback.conf:ro
```

test/test_fallback/test_fallback.data/nodefault.yml
```diff
@@ -1,6 +1,5 @@
 services:
-  sut:
-    image: nginxproxy/nginx-proxy:test
+  nginx-proxy:
     volumes:
       - /var/run/docker.sock:/tmp/docker.sock:ro
       - ./nodefault.certs:/etc/nginx/certs:ro
```
SchoNie, line 5:
```suggestion
      - ${PYTEST_MODULE_PATH}/nodefault.certs:/etc/nginx/certs:ro
```

test/test_fallback/test_fallback.data/nohttp-on-app.yml
```diff
@@ -1,6 +1,5 @@
 services:
-  sut:
-    image: nginxproxy/nginx-proxy:test
+  nginx-proxy:
     volumes:
       - /var/run/docker.sock:/tmp/docker.sock:ro
       - ./withdefault.certs:/etc/nginx/certs:ro
```
SchoNie, line 5:
```suggestion
      - ${PYTEST_MODULE_PATH}/withdefault.certs:/etc/nginx/certs:ro
```

test/test_fallback/test_fallback.data/nohttp-with-missing-cert.yml
```diff
@@ -1,6 +1,5 @@
 services:
-  sut:
-    image: nginxproxy/nginx-proxy:test
+  nginx-proxy:
     volumes:
       - /var/run/docker.sock:/tmp/docker.sock:ro
       - ./withdefault.certs:/etc/nginx/certs:ro
```
SchoNie, line 5:
```suggestion
      - ${PYTEST_MODULE_PATH}/withdefault.certs:/etc/nginx/certs:ro
```

test/test_fallback/test_fallback.data/nohttp.yml
```diff
@@ -1,6 +1,5 @@
 services:
-  sut:
-    image: nginxproxy/nginx-proxy:test
+  nginx-proxy:
     volumes:
       - /var/run/docker.sock:/tmp/docker.sock:ro
       - ./withdefault.certs:/etc/nginx/certs:ro
```
SchoNie, line 5:
```suggestion
      - ${PYTEST_MODULE_PATH}/withdefault.certs:/etc/nginx/certs:ro
```

test/test_fallback/test_fallback.data/untrusteddefault.yml
```diff
@@ -1,6 +1,5 @@
 services:
-  sut:
-    image: nginxproxy/nginx-proxy:test
+  nginx-proxy:
     volumes:
       - /var/run/docker.sock:/tmp/docker.sock:ro
       - ./withdefault.certs:/etc/nginx/certs:ro
```
SchoNie, line 5:
```suggestion
      - ${PYTEST_MODULE_PATH}/withdefault.certs:/etc/nginx/certs:ro
```

test/test_fallback/test_fallback.data/withdefault.yml
```diff
@@ -1,6 +1,5 @@
 services:
-  sut:
-    image: nginxproxy/nginx-proxy:test
+  nginx-proxy:
     volumes:
       - /var/run/docker.sock:/tmp/docker.sock:ro
       - ./withdefault.certs:/etc/nginx/certs:ro
```
SchoNie, line 5:
```suggestion
      - ${PYTEST_MODULE_PATH}/withdefault.certs:/etc/nginx/certs:ro
```

test/test_nominal/test_nominal.py
```diff
@@ -19,6 +21,10 @@ def test_forwards_to_web2(docker_compose, nginxproxy):
     assert r.text == "answer from port 82\n" 
 
 
+@pytest.mark.skipif(
+    platform.system() == "Darwin",
+    reason="This test rely on being able to directly contact the container's IP"
```
SchoNie, line 26:
```suggestion
    reason="This test depends on direct communication with the container's IP"
```

test/test_ipv6/test_ipv6.py
```diff
@@ -1,3 +1,13 @@
+import platform
+
+import pytest
+
+pytestmark = pytest.mark.skipif(
+    platform.system() == "Darwin",
+    reason="Those tests rely entirely on being able to directly contact container's IP"
```
SchoNie, line 7:
```suggestion
    reason="Those tests rely entirely on being able to directly contact the container's IP"
```

SchoNie (commented):
Nice work good improvements! Code looks good for me.
A few missed strings and 2 grammar suggestions.

Environment variables (`VIRTUAL_HOST` and some other) could use some linting. Good time to choose the consistent standard as there are already lot of changes in each file.

test/conftest.py
```diff
@@ -428,30 +472,33 @@ def connect_to_all_networks() -> List[Network]:
...
             return
         self._down()
-        if docker_compose_file is None:
+        if docker_compose_files is None or project_name is None:
             return
-        docker_compose_up(docker_compose_file)
+        docker_compose_up(docker_compose_files, project_name)
         self._networks = connect_to_all_networks()
         wait_for_nginxproxy_to_be_ready()
         time.sleep(3)  # give time to containers to be ready
```
SchoNie, line 499: Not really related to this PR but why we sleep 3 seconds here?
`wait_for_nginxproxy_to_be_ready()` checks already to have nginx-proxy ready (real-time stream)
So that sleeps just seems to be for the 'web' container. But I don't think that needs 3 seconds.
And as we do like 90 compose up's thats already 270 seconds of waiting (4,5 min).

A good win to reduce test time I think.

buchdag: That's something I tried to work on, one problem is that in its current state `wait_for_nginxproxy_to_be_ready()` does not seem to really tell us if the proxy is actually ready or not.

```python
def wait_for_nginxproxy_to_be_ready():
    """
    If one (and only one) container started from image nginxproxy/nginx-proxy:test is found,
    wait for its log to contain substring "Watching docker events"
    """
    containers = docker_client.containers.list(filters={"ancestor": "nginxproxy/nginx-proxy:test"})
    if len(containers) != 1:
        return
    container = containers[0]
    for line in container.logs(stream=True):
        if b"Watching docker events" in line:
            logging.debug("nginx-proxy ready")
            break
```

- If `docker_client.containers.list()` found zero or more than one container, the function will return without any further check.
- I'm not certain that `for line in container.logs(stream=True):` mean that the function will wait for the expected log line to appear, for me it will just iterate over the the log lines generated by the container up to that point, return early if it find `Watching docker events` or return anyway when it reach the end of the generator.
- the appearance of the `Watching docker events` log line mean that docker-gen is ready, not that nginx has loaded the docker-gen generated conf, which is what we're actually testing.

We're automatically doing a 9s backoff on 404 or 503 when using the  `nginxproxy` fixture's `get()` which kind of automatically wait for the containers to be ready, but not all tests are using this.

buchdag:
> Environment variables (`VIRTUAL_HOST` and some other) could use some linting. Good time to choose the consistent standard as there are already lot of changes in each file.

Eventually I'd like to change the use of environment variable on app containers:
- environment variables should only be used on the nginx-proxy container to configure the proxy itself
- labels should be used on the app containers to configure the reverse proxying to those containers

The rationale behind this is that environment variables should be used to store runtime configuration for a given process / container, not metadata for something else (ie nginx-proxy) to configure a third party service (ie a reverse proxy) on behalf of this container. nginx-proxy should not access other container's environment variables.

Using labels for this kind of purpose is actually how Kubernetes does it. 

Furthermore, since environment variables aren't namespaced, collision can (and already did) occur, ie an app container that use some environment variable with similar names as those used by nginx-proxy might be unusable with nginx-proxy.

App containers's environment variables such as `VIRTUAL_HOST` should ideally have a corresponding label created, replaced with the corresponding label on the docs and be marked as deprecated.

test/conftest.py
SchoNie: Ah you are right, that ready state is indeed based of docker-gen.
Changing the bytestring to something else makes the for loop block. So it will wait. But the `Watching docker events` is probably always there.

Adding `follow=True` in `for line in container.logs(stream=True, follow=True):` makes the generator [follow](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.logs) new events in the stream.

SchoNie (approved):
[- line 62](https://github.com/nginx-proxy/nginx-proxy/blob/aa8145b62dbb7e10fbe05358141381ee35241cf3/test/test_events/test_events.py#L62)

buchdag:
I think I'll do a linting pass on every Python files later 👍

Issue #2583, opened 2025-01-31
Please revert 142a159 - ci: bump python module docker version check


pini-gh:
Commit 142a159 breaks the test-suite on debian stable (12 - bookworm) which has python3-docker 5.0.3.

Thanks,

SchoNie:
Maybe try to update the [docker](https://pypi.org/project/docker/#history) pip requirement?
https://github.com/nginx-proxy/nginx-proxy/blob/1da623019f56182b6d6ace282715cdf092da551e/test/requirements/python-requirements.txt#L2

pini-gh:
This is not an option for me. I want to stick to the packages available in Debian stable.

buchdag:
@SchoNie I'm not sure there would be adverse effect to lowering the version check, as far as I know we're not using stuff that's only available on version 7 of the docker Python module. 🤔

SchoNie:
Yes, I think lowering will work fine. I just bumped it because 7 added [official Python 3.12 support](https://docker-py.readthedocs.io/en/7.0.0/change-log.html#features).
But I am interested what this docker pip module has to do with Debian stable packages. I would expect no external libraries are installed by default so it was installed manually. So why not upgrade to 7.1? Even on old python versions...

The changes in `packaging` and `version` don't seem breaking as they are supported even in really old Python versions.

pini-gh:
> But I am interested what this docker pip module has to do with Debian stable packages. I would expect no external libraries are installed by default so it was installed manually. So why not upgrade to 7.1? Even on old python versions...

On Debian many Python libraries are packaged and available as Debian packages. So is the docker Python library, at version 5.0.3 in Debian bookworm:
https://tracker.debian.org/pkg/python-docker

Many people prefer to stick to official Debian packages for Python libraries instead of installing manually from PIP.

buchdag:
You don't use the `requirement.txt` then ?

pini-gh:
Indeed. It worked fine 'til 1.7.0.

buchdag:
@SchoNie any remaining objection to the revert of the version check ?

SchoNie:
> [@SchoNie](https://github.com/SchoNie) any remaining objection to the revert of the version check ?

No, if it helps @pini-gh go ahead, that is most important!

PR #2584, opened 2025-01-31
Sopports docker-compose v1 and Debian bookwom (current stable)

### LINKED ISSUE #2582: Please keep the test suite compatible with docker-compose v1
Hi,

Since the recent test-suite refactoring it isn't possible to run it with docker-compose v1 anymore. I am aware that docker-compose v1 is deprecated in favor of v2, but Debian still has it. Could you please consider rolling back the refactoring to support docker-compose v1?

I can propose a patch if you're okay with that.

Many thanks in advance.

pini-gh:
Fix #2582 #2583 .


SchoNie (commented):
You are right there is something wrong with the cleanup logic if an exception occurs. I experienced the same thing no teardown happens and it leaves all ghost containers and networks. Especially in the host-network-mode tests like you also have. I had the same issue on KeyboardInterrupt but I had no time to dig into this. So thanks for also pointing this out.

My first investigation suspects is `pytest.fail` from [L328](https://github.com/nginx-proxy/nginx-proxy/blob/main/test/conftest.py#L328) and [L336](https://github.com/nginx-proxy/nginx-proxy/blob/main/test/conftest.py#L336)
I think this exits the [DockerComposer](https://github.com/nginx-proxy/nginx-proxy/blob/main/test/conftest.py#L472) ContextManager and no _down teardown is executed.

But instead of doing `docker_compose_down` if an exception happens I would recommend to fix the ContextManager.

Sorry I cannot test this right now but my recommendations are:
1. Change the pytest.fail from `docker_compose_up` and `docker_compose_down` to a `logging.error`
2. Improve the compose function in the ContextManager with a try except block to make sure teardown happens. And also move the self. declarations up so we are sure they are set even if an exception occurs.

```python
    def compose(self, docker_compose_files: List[str], project_name: str):
        if docker_compose_files == self._docker_compose_files and project_name == self._project_name:
            return
        self._down()
        if docker_compose_files is None or project_name is None:
            return
        #Move these states to early top, because we are sure it is set. It never reached it on exception.         
        self._docker_compose_files = docker_compose_files
        self._project_name = project_name
        
        try:
            docker_compose_up(docker_compose_files, project_name)
            self._networks = connect_to_all_networks()
            wait_for_nginxproxy_to_be_ready()
            time.sleep(3)  # give time to containers to be ready

        except KeyboardInterrupt:
            logging.warning("KeyboardInterrupt detected! Cleaning up...")
            self._down()  # Always teardown
            raise  # Re-raise to allow pytest to exit cleanly

        except:
            logging.error(f"Docker Compose setup failed")
            self._down()  # Always teardown
            pytest.fail("Docker Compose setup failed")
```

3. [L487](https://github.com/nginx-proxy/nginx-proxy/blob/main/test/conftest.py#L487) `self._docker_compose_file = None` >> `self._docker_compose_files = None` (file**s**)
4. Should we make sure `self._networks = []` is set in `_down` ?

pini-gh:
> Sorry I cannot test this right now but my recommendations are:

Well, take your time :)

As said before my Python skills are limited. I came up with my rudimentary fix, but I won-t be able to do this the _right way_ (tm) as easily as you would.

buchdag:
@SchoNie I had issues with the way things where setup in [test_fallback](https://github.com/nginx-proxy/nginx-proxy/blob/main/test/test_fallback/test_fallback.py) too, I almost refactored it to avoid using a self defined `docker_compose_files` as I'm not certain it adds much. Unless I'm missing something I don't see why this test module would necessitate running multiple checks over multiple Docker Compose setups in one gigantic parameterized test case.

@pini-gh  could we focus this PR on fixing the test suite on Debian / with Docker Compose v1 ? I think the error handling issue is a separate problem and should be handled in a separate PR, allowing this one to be merged.

buchdag (approved):

buchdag:
@pini-gh just to be sure, the PR in its current state fixes both #2582 and #2583 on your Debian running host ?

pini-gh:
> @pini-gh just to be sure, the PR in its current state fixes both #2582 and #2583 on your Debian running host ?

Yes

PR #2587, opened 2025-02-18
Proxy Protocol Support

antoniomika:
Similar to #1487, I took a look at adding proxy-protocol support to this project. 

The changes included here add a new `ENABLE_PROXY_PROTOCOL` env var, which will enable proxy-protocol support for all docker-gen created listeners. Because setting `proxy_protocol` on a `listen` enables the functionality for all `server`'s that use the same port, I opted to make this option global.

Unlike the other PR, I decided to map the `x-forwarded-for` (and `x-forwarded-port`) entry to conform with current use-cases. If setting the real ip module settings is desired, there already exists methods for adding other configurations to the template.

I also added tests to verify this functionality **DOES NOT** break or modify any of the current proxy header use cases. This feature is opt-in, as nginx will stop responding to HTTP requests that aren't made using the proxy protocol.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -805,6 +805,25 @@ For legacy compatibility reasons, `nginx-proxy` forwards any client-supplied `X-
 
 The default for `TRUST_DOWNSTREAM_PROXY` may change to `false` in a future version of `nginx-proxy`. If you require it to be enabled, you are encouraged to explicitly set it to `true` to avoid compatibility problems when upgrading.
 
+### Proxy Protocol Support
+
+`nginx-proxy` has support for the [Proxy Protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt). This allows a separate proxy to send requests to `nginx-proxy` and encode information about the client connection without relying on HTTP headers. This can be enabled by setting `ENABLE_PROXY_PROTOCOL=true` on the main `nginx-proxy` container. It's important to note that enabling the proxy protocol will require all connections to `nginx-proxy` to use the protocol.
+
+You can use this feature in conjunction with the `realip` module in nginx. This will allow for setting the `$remote_addr` and `$remote_port` nginx variables to the IP and port that are provided from the protocol message. Documenation for this functionality can be found in the [nginx documentation](https://nginx.org/en/docs/http/ngx_http_realip_module.html).
+
+A simple example is as follows:
+
+1. Create a configuration file for nginx, this can be global (in `conf.d`) or host specific (in `vhost.d`)
+2. Add your `realip` configuration:
+
+```nginx
+# Your proxy server ip address
+set_real_ip_from  192.168.1.0/24;
+
+# Where to replace `$remote_addr` and `$remote_port` from
+real_ip_header proxy_protocol;
+```
+
 ⬆️ [back to table of contents](#table-of-contents)
 
 ## Custom Nginx Configuration
@@ -1451,6 +1470,7 @@ curl -s -H "Host: test.nginx-proxy.tld" localhost/nginx-proxy-debug | jq
     "enable_debug_endpoint": "true",
     "enable_http2": "true",
     "enable_http3": "false",
+    "enable_proxy_protocol": "false",
     "enable_http_on_missing_cert": "true",
     "enable_ipv6": false,
     "enable_json_logs": false,
--- a/docs/README.md
+++ b/docs/README.md
@@ -1348,6 +1348,7 @@ Configuration available either on the nginx-proxy container, or the docker-gen c
 | [`ENABLE_HTTP2`](#http2-support) | `true` |
 | [`ENABLE_HTTP3`](#http3-support) | `false` |
 | [`ENABLE_IPV6`](#listening-on-ipv6) | `false` |
+| [`ENABLE_PROXY_PROTOCOL`](#proxy-protocol-support) | `false` |
 | [`HTTP_PORT`](#custom-external-httphttps-ports) | `80` |
 | [`HTTPS_PORT`](#custom-external-httphttps-ports) | `443` |
 | [`HTTPS_METHOD`](#how-ssl-support-works) | `redirect` |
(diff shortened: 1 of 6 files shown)


bvli:
I'd like to see this feature implemented as well.. Are there any reason for not merging this PR?

nginx.tmpl
```diff
@@ -436,6 +437,11 @@ upstream {{ $vpath.upstream }} {
     {{- when .Enable "access_log /var/log/nginx/access.log vhost;" "" }}
 {{- end }}
 
+map $proxy_add_x_forwarded_for $proxy_x_forwarded_for {
+    default {{ if $globals.config.trust_downstream_proxy }}{{ if $globals.config.enable_proxy_protocol }}$proxy_protocol_addr{{ else }}$proxy_add_x_forwarded_for{{ end }}{{ else }}$remote_addr{{ end }};
+    '' $remote_addr;
+}
```
buchdag, line 443: **suggestion**: that should be a bit easier to read
```suggestion
map $proxy_add_x_forwarded_for $proxy_x_forwarded_for {
    {{- if $globals.config.trust_downstream_proxy }}
        {{- if $globals.config.enable_proxy_protocol }}
    default $proxy_protocol_addr;
        {{- else }}
    default $proxy_add_x_forwarded_for;
        {{- end }}
    {{- else }}
    default $remote_addr;
    {{- end }}
    '' $remote_addr;
}
```

nginx.tmpl
```diff
@@ -887,9 +899,9 @@ server {
     server_tokens {{ $vhost.server_tokens }};
         {{- end }}
     {{ template "access_log" (dict "Enable" $globals.config.enable_access_log) }}
-    listen {{ $globals.config.external_http_port }} {{ $default_server }};
+    listen {{ $globals.config.external_http_port }} {{- $default_server }} {{- $proxy_protocol }};
```
buchdag, line 902: **question**: I'm not sure of what's get rendered here if `default_server` and `proxy_protocol` both have a non blank value, aren't the two strings concatenated because of `{{-` ? 🤔

antoniomika: Nope, everything looks correct because of this: https://github.com/nginx-proxy/nginx-proxy/pull/2587/files/6baf7bbf57785df273cd5ed5871937b9a0a1da2d#diff-a2a77c71854a4489d896f49c59a5f864e5b1e290f6b1ba16f9f39c1a32829c57R887

I made this change so the spacing doesn't look all wonky if neither are enabled.

Here's a test if you're curious:

1. Run a proxy-protocol test case like so:
```bash
~$ make build-nginx-proxy-test-alpine build-webserver && docker compose -f ./test/compose.base.yml -f ./test/test_proxy_protocol/test_proxy-protocol-global-enabled.yml up -d
```
2. Exec the container and check the proxy lines look okay:
```bash
~$ docker exec nginx-proxy nginx -T 2>&1 | grep 'listen.*proxy_protocol'
    listen 80 proxy_protocol;
    listen 443 ssl proxy_protocol;
    listen 80 proxy_protocol;
    listen 443 ssl proxy_protocol;
```
3. Add a default server by adding `DEFAULT_HOST: "proxy-protocol-global-enabled.nginx-proxy.tld"` to the `nginx-proxy` service' environment in `./test/test_proxy_protocol/test_proxy-protocol-global-enabled.yml`. Rerun the command in 1.
4. Re-grep to check the result:
```bash
~$ docker exec nginx-proxy nginx -T 2>&1 | grep 'listen.*proxy_protocol'
    listen 80 default_server proxy_protocol;
    listen 443 ssl default_server proxy_protocol;
```
5. Disable proxy protocol by setting the `ENABLE_PROXY_PROTOCOL` environment var to false in the same file above. Rerun the command in step 1 and grep for `default_server`.
```bash
~$ docker exec nginx-proxy nginx -T 2>&1 | grep 'listen.*default_server'
    listen 80 default_server;
```
6. Finally with both disabled:
```bash
    listen 80;
    listen 443 ssl;
    listen 80;
    listen 443 ssl;
```

Rerunning the steps above on main, default server disabled looks like this:
```
    listen 80;
    listen 443 ssl;
    listen 80 ;
    listen 443 ssl ;
```

And enabled like this:
```
    listen 80 default_server;
```

antoniomika:
Thanks for the review @buchdag! I addressed your comments and squashed/rebased the PR. Let me know if there's anything else I can help with.

docs/README.md
```diff
@@ -805,6 +805,25 @@ For legacy compatibility reasons, `nginx-proxy` forwards any client-supplied `X-
 
 The default for `TRUST_DOWNSTREAM_PROXY` may change to `false` in a future version of `nginx-proxy`. If you require it to be enabled, you are encouraged to explicitly set it to `true` to avoid compatibility problems when upgrading.
 
+### Proxy Protocol Support
+
+`nginx-proxy` has support for the [Proxy Protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt). This allows a separate proxy to send requests to `nginx-proxy` and encode information about the client connection without relying on HTTP headers. This can be enabled by setting `ENABLE_PROXY_PROTOCOL=true` on the main `nginx-proxy` container. It's important to note that enabling the proxy protocol will require all connections to `nginx-proxy` to use the protocol.
+
+You can use this feature in conjunction with the `realip` module in nginx. This will allow for setting the `$remote_addr` and `$remote_port` nginx variables to the IP and port that are provided from the protocol message. Documenation for this functionality can be found in the [nginx documentation](https://nginx.org/en/docs/http/ngx_http_realip_module.html).
```
SchoNie, line 812:
```suggestion
You can use this feature in conjunction with the `realip` module in nginx. This will allow for setting the `$remote_addr` and `$remote_port` nginx variables to the IP and port that are provided from the protocol message. Documentation for this functionality can be found in the [nginx documentation](https://nginx.org/en/docs/http/ngx_http_realip_module.html).
```

nginx.tmpl
buchdag: So you've actually fixed the pre existing spacing issue, thank you 👍

PR #2591, opened 2025-02-28
tests: DockerComposer contextmanager improvements 

SchoNie:
Improvements to the DockerComposer contextmanager as mentioned in https://github.com/nginx-proxy/nginx-proxy/pull/2584#pullrequestreview-2588289338

I run the test suite as container, and found that if errors occur or on keyboardinterrupt the containers and/or networks are not cleaned properly leaving ghost stuff around and preventing new test runs.
In Github CI not a problem because after tests the runner is terminated.

- Improve the DockerComposer contextmanager to make sure teardown happens by using try except block.
- Move the self. declarations up so we are sure they are set even if an exception occurs.
- set `self._networks`
- `self._docker_compose_file = None` >>` self._docker_compose_files = None` (files typo)
- Change pytest.fail to a logging error because `pytest.fail` makes DockerComposer ContextManager exit and no _down teardown is executed.
- Add KeyboardInterrupt logic
- skip test_proxy-host-network-mode when `PYTEST_RUNNING_IN_CONTAINER` as this does not work

A review from @pini-gh is highly appreciated to make sure Debian Stable is still OK.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -428,9 +428,15 @@ def connect_to_network(network: Network) -> Optional[Network]:
         # Make sure our container is connected to the nginx-proxy's network,
         # but avoid connecting to `none` network (not valid) with `test_server-down` tests
         if network.name not in my_networks and network.name != 'none':
-            logging.info(f"Connecting to docker network: {network.name}")
-            network.connect(my_container)
-            return network
+            try:
+                logging.info(f"Connecting to docker network: {network.name}")
+                network.connect(my_container)
+                return network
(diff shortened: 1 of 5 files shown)


pini-gh:
The tests run fine on Debian.

buchdag (approved):
I tried to run this on Darwin / macOS but somehow tests are broken again on my mac with PyCharm (not this branch specifically, they're broken on the main branch too).

I'll try again tomorrow, if I still can't get the tests to work I'll merge the PR anyway.

Issue #1046, opened 2018-01-29
Mutual auth between server and client


enric1994:
Hey!
I am trying to implement mutual authentication between my web and my client.
For the moment, I configured the server side auth, using: 
`-v /path/to/certs:/etc/nginx/certs `

It works properly with the server side.

To enable the user auth I amb trying to modify the /etc/nginx/conf.d/default.conf mounting my own .conf file, which includes the following lines:
```
  ssl_client_certificate /etc/nginx/certs/ca/ca.crt;
  ssl_verify_client on;
```

But the nginx-proxy container don't start when I load the .conf.
**How can I modify it to enable user authentication?**
 
(I created the user cert and imported it to the browser)
(Using Docker 17.12.0-ce , docker-compose 1.18.0 and the latest version of jwilder/nginx-proxy)

Thanks!

enric1994:
The answer is in: https://github.com/jwilder/nginx-proxy/issues/1027#issuecomment-362042435

PR #2592, opened 2025-02-28
feat: SSL client certificate validation

### LINKED ISSUE #137: Adding extra SSL options
I am looking for a sound way to add extra SSL related options, specifically:
- ssl_verify_client
- ssl_crl

But when I attempt to add any SSL related options using the "proxy-wide" or "per-VIRTUAL_HOST" methodologies described in the README, the proxy goes belly-up.

``` Bash
dockergen.1 | 2015/04/02 19:49:22 Error running notify command: nginx -s reload, exit status 1
```

Is there a sensible way to achieve this without resorting to using a custom Go template?

SchoNie:
This PR adds support for the `ssl_client_certificate`, `ssl_verify_client optional` and `ssl_crl` directives. It was requested and discussed multiple times over the years and I wanted to expirement with mTLS in my setup so I continued and combined the existing PR's enabling verification of client side certificates ([mutual TLS](https://www.cloudflare.com/learning/access-management/what-is-mutual-tls/)) based on if a CA file exists.

A [Certificate Revocation List (CRL)](https://medium.com/@arpanagupta10/understanding-certificate-revocation-list-using-openssl-c57f1b7dc43c) can be used.

Optional [ssl_verify_client](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_verify_client) can be activated by using the `com.github.nginx-proxy.nginx-proxy.ssl_verify_client: "optional"` label. This does not block access but instead stores the result  in the [$ssl_client_verify](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#var_ssl_client_verify) variable which gives user control which paths are controlled.

Tests for all kind of mTLS scenarios. (10 year expiry cert)

Resolves #137, #278,#1027, #1046, #1197, #1584, #1644, 
Discussed in: #2568

If you think this is something interesting to be merged I will add docs and write a wiki page how to generate all the client certificates. Please let me know.


buchdag:
> If you think this is something interesting to be merged

@SchoNie I do 👍

SchoNie:
> > If you think this is something interesting to be merged
> 
> @SchoNie I do 👍

[Wiki page](https://github.com/nginx-proxy/nginx-proxy/wiki/mTLS-client-side-certificate-authentication) and added docs.

buchdag (commented):
Incredible work @SchoNie , thanks 👍 

The only suggestion I have other than the style one would be to maybe make it a bit clearer in the docs that global CRL can only work with global CA, and same thing for per-`VIRTUAL_HOST`.

buchdag (approved):

PR #2601, opened 2025-03-13
Rename generated nginx config file to 50-default.conf to help ordering

p12tic:
Currently nginx-proxy puts its configuration into a file named default.conf which is included via include /etc/nginx/conf.d/*.conf; directive. The nginx configuration is order dependent, thus if users rely on anything in default.conf being ordered before or after their config file, they must choose the name for the config file accordingly. Currently the name must be something that comes either before or after default.conf in alphabetical order. This is confusing.

A simple fix for this is be to rename the generated config file from default.conf to have some number prepended, such as 50-default.conf. This commit implements exactly this.

Fixes https://github.com/nginx-proxy/nginx-proxy/issues/2593.


p12tic:
Closing per https://github.com/nginx-proxy/nginx-proxy/issues/2593#issuecomment-2747359671

Issue #2606, opened 2025-04-08
split nginx.tmpl to support stream module


lloydzhou:
i think, can split nginx.tmpl in two files:
`upstream.tmpl` --> `upstream.conf`
`server.tmpl` --> `default.conf`

then, can include upstream.conf in `http` block and `stream` block

related #792 #979 #1009 #759

lloydzhou:
the `stream_upstream` do not sopport `keepalive`, so just genrate new file (stream.conf.d/10-upstream.conf) for stream block

https://nginx.org/en/docs/stream/ngx_stream_upstream_module.html
https://nginx.org/en/docs/http/ngx_http_upstream_module.html

PR #2608, opened 2025-04-10
split nginx.tmpl to support stream module

lloydzhou:
# split nginx.tmpl to support stream module

`nginx.tmpl` with env `TARGET=streamupstream` --> `stream.conf.d/10-upstream.conf`
`nginx.tmpl` --> `default.conf`

then, can include upstream.conf in `stream` block

#2606

# usage
```
# stream.conf
stream {
    include stream.conf.d/10-upstream.conf;

    server {
        listen     8000;
        #TCP traffic will be forwarded to the specified server
        proxy_pass web1.nginx-proxy.example;
    }
}

docker run --detach \
    --name nginx-proxy \
    --publish 80:80 \
    --publish 8000:8000 \
    --volume /var/run/docker.sock:/tmp/docker.sock:ro \
    --volume ./stream.conf:/etc/nginx/toplevel.conf.d/stream.conf:ro \
    nginxproxy/nginx-proxy
```


buchdag:
Hi

Thanks for the PR.

Unfortunately I'm not fond of the approach of having the `nginx.tmpl` work in two different contexts with conditional logic, as this won't be maintainable long term.

Running multiple docker-gen processes in the container should also be avoided (a single process can render multiple templates).

PR #2614, opened 2025-05-03
Clear instruction for naming of htpasswd file

damascene:
Previous instructions could make the impression that user have to create a directory named as the virtual host and also a file named as the virutal host.


buchdag:
Thanks @damascene 👍

PR #2626, opened 2025-05-29
fix: set auth_request off for all acme challenge locations

srstsavage:
Adds missing `auth_request off;` to a few `.well-known/acme-challenge` location blocks.

This is needed to allow unrestricted access to `.well-known/acme-challenge` files on servers where `auth_request` is otherwise globally applied.

See #1409, nginx-proxy/acme-companion#570


buchdag:
Thanks @srstsavage 👍

PR #2650, opened 2025-08-19
docs: add precisions to multi container compose file

Bob-le-pirate:
Added precision for 3-container version

### DIFF
--- a/docker-compose-separate-containers.yml
+++ b/docker-compose-separate-containers.yml
@@ -9,15 +9,20 @@ services:
       - "80:80"
     volumes:
       - nginx_conf:/etc/nginx/conf.d:ro
+      - network_internal.conf:/etc/nginx/network_internal.conf:ro
 
   dockergen:
     image: nginxproxy/docker-gen
     command: -notify-sighup nginx -watch /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
+    # the name "nginx" between "-sighup" and "-watch" must exactly reflect the service name used for the nginx container above.
+    environment:
+      # Environment Variable (HTTP_PORT etc.) must be inserted as variables of dockergen container (not nginx container)  
     volumes:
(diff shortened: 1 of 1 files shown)


docker-compose-separate-containers.yml
```diff
@@ -9,15 +9,20 @@ services:
       - "80:80"
     volumes:
       - nginx_conf:/etc/nginx/conf.d:ro
+      - network_internal.conf:/etc/nginx/network_internal.conf:ro
 
   dockergen:
     image: nginxproxy/docker-gen
     command: -notify-sighup nginx -watch /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
+    # the name "nginx" between "-sighup" and "-watch" must exactly reflect the service name used for the nginx container above.
```
buchdag, line 17:
```suggestion
    # the name "nginx" passed to the "-notify-sighup" flag must exactly reflect the service name used for the nginx container above.
```

docker-compose-separate-containers.yml
```diff
@@ -9,15 +9,20 @@ services:
...
+    environment:
+      # Environment Variable (HTTP_PORT etc.) must be inserted as variables of dockergen container (not nginx container)
```
buchdag, line 19:
```suggestion
      # nginx-proxy environment variable (HTTP_PORT etc.) must be set on the docker-gen container (not the nginx container)  
```

buchdag (changes requested):
I'm not sure that the usage of `network_mode: "none"` and the internal network feature should be made default in the compose examples without any explanation. The usage of `network_mode` on a service [is also mutually exclusive with `networks` for the entire compose file](https://docs.docker.com/reference/compose-file/services/#network_mode) and I don't think it's a good idea to suggest this as the base example.

The compose file are meant to serve as minimalist examples, and users are supposed to read the doc and build from there.

PR #2659, opened 2025-09-11
Add container filtering feature

eorea:
This feature allows the filtering of containers with an specific label, those containers that have the filtering label defined would be the ones that would become available for processing.

The feature is fully backwards compatible, if no filtering label is provided via CONTAINER_FILTER_LABEL environment variable all containers in the docker host will be considered for processing.

4 test scenarios has been included, 2 to validate new behavior in both single o multiple instance setups, and 2 to validate the backwards compatible scenarios with the feature not enabled also in both single o multiple instance setups.

The scenarios identified where using this filtering is helpful are:
- Docker engine running huge number of containers not required to be proxied. Shortening the unfiltered container list from the very beginning helps to reduce processing time in further container list iterations.
- Multiple nginx-proxy instances. Assigning different filtering labels to different instances of nginx-proxy allows them to proxy independent services within the same docker engine. This also has the benefit to split responsibilities for scenarios requiring split external and internal facing proxies. In cases where a service is required to be proxied in multiple nginx-proxy instances it suffices only to add all the respective filtering labels to the service's container definition.

### DIFF
--- a/test/test_container_filtering/test_container_filtering-disabled-multiple-hosts.base.yml
+++ b/test/test_container_filtering/test_container_filtering-disabled-multiple-hosts.base.yml
@@ -1,0 +1,54 @@
+networks:
+  proxy:
+  private:
+    internal: true
+
+volumes:
+  nginx_conf:
+
+services:
+
+  nginx-proxy-nginx:
+    image: nginx
+    container_name: nginx
+    volumes:
+      - nginx_conf:/etc/nginx/conf.d:ro
+    ports:
+      - "80:80"
+      - "443:443"
+    networks:
+      - proxy
+    labels:
+      - "com.github.nginx-proxy.nginx-proxy.nginx"
+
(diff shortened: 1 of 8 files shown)


p12tic:
https://github.com/nginx-proxy/nginx-proxy/pull/2660 also implements support for a similar use case by using a different approach. The VIRTUAL_HOST and VIRTUAL_HOST_MULTIPORTS environment variables are made configurable for each of the different nginx-proxy instances.

I have no opinion on which approach is better.

eorea:
FYI, seems that docker_gen will be also implementing filtering using labels by end of month https://github.com/nginx-proxy/docker-gen/issues/276 this will achieve the exact same thing as this PR, and CONTAINER_FILTER_LABEL variable could be used to filter straight from docker_gen instead of doing it at the templating

p12tic:
Great, seems my PR is not needed then.

PR #2660, opened 2025-09-14
RFC: feat: Support multiple networks by running multiple instances of nginx-proxy

p12tic:
I'm not sure if it makes sense nginx-proxy to support a feature like this, but I already have a proof of concept and instead of opening an issue a PR was created straight away. Let me know if the feature itself makes sense and I will rewrite PR in any way needed.

----- 

This feature is useful in cases there are multiple networks to be managed with nginx-proxy on the same host. For example, a public network and a network that can only be reached through a VPN.

This commit makes it possible to use multiple instances of nginx-proxy in such use cases. Each instance has a separate "namespace" of VIRTUAL_HOST and VIRTUAL_HOST_MULTIPORTS variables used by containers, controlled by VIRTUAL_HOST_ENV_PREFIX environment variable set on the nginx-proxy. E.g. set VIRTUAL_HOST_ENV_PREFIX=VIRTUAL_HOST_MY_NETWORK, and only containers with VIRTUAL_HOST_MY_NETWORK=something will be managed by that nginx-proxy instance.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1326,6 +1326,15 @@ I'm 5b129ab83266
 
 ⬆️ [back to table of contents](#table-of-contents)
 
+## Managing multiple networks
+
+It is possible to manage multiple independent networks on single host by running multiple instances of nginx-proxy.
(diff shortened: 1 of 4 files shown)


p12tic:
Closing because https://github.com/nginx-proxy/docker-gen/issues/276 will implement the same.

PR #2697, opened 2026-01-29
Added bits to use LETSENCRYPT_CERTNAME in virtual host configuration

jekewa:
Added bits to use LETSENCRYPT_CERTNAME in virtual host configuration to use volume /etc/letsencrypt as maintained by external certbot.

I already have certbot configured to check and renew my domains. I've got post-renewal triggers that do other work. Adding an nginx-proxy sidecar to also maintain my certificates seemed unnecessary. Additionally, the nginx-proxy template uses different cert names and format than Let's Encrypt does.

I added a tiny bit to allow the environment variable LETSENCRYPT_CERTNAME to be added to a virtual host container, and for that certificate (if found) to be used in the nginx-proxy server block for the virtual host.

This could probably be improved to allow the dhparam and ssl stapling to remain, but as the blocks were dependent on the other cert names, I thought it less intrusive to introduce the Let's Encrypt cert files without impacting those conditions. The only changes in that block is white space, to try to keep the scripting aligned.

I have an example built from the Dockerfile.debian at https://hub.docker.com/repository/docker/jekewa/nginx-proxy, which includes this little bit that shows the change required to make it work:

### Run Proxy Container

Everything is the same as running a normal nginx-proxy container, except that the /etc/letsencrypt volume needs to be mounted.

```
docker run --detach \
    --name nginx-proxy \
    --publish 80:80 \
    --publish 443:443 \
    --volume /var/run/docker.sock:/tmp/docker.sock:ro \
    --volume /etc/letsencrypt:/etc/letsencrypt:ro \
    jekewa/nginx-proxy
```

The /etc/letsencrypt volume can be mounted by any comfortable Docker means. The example above assumes the host /etc/letsencrypt volume exists. Use a network bind or other mechanism as desired.

### Run Virtual Host

Add the LETSENCRYPT_CERTNAME variable to the virtual host container to name the certificate directory and trigger the necessary portion in the nginx-proxy template.

```
docker run --detach \
    --name your-proxied-app \
    --env VIRTUAL_HOST=foo.bar.com \
    --env LETSENCRYPT_CERTNAME=bar.com \
    nginx
```

In this example the assumption is that the Let's Encrypt certificate was created for the whole bar.com domain. This would lead to searching for /etc/letsencrypt/live/bar.com and using the expected files.


buchdag:
@jekewa that looks like a very specific solution to a somewhat specific use case.

I'd be more inclined to have a more generally usable way to configure the path of a given container's certificates.

jekewa:
I appreciate the feedback.

Maybe the names could be different, going more CERTBOT and less LETSENCRYPT, but it's a pretty standard configuration for a very popular use case, with claims of offering 60% of the SSL certificates used by sites in the data collected. 

https://eff-certbot.readthedocs.io/en/stable/using.html#where-are-my-certificates
https://w3techs.com/technologies/details/sc-letsencrypt

The current assumptions about how the SSL certificate files are structured and named is a bit particular and doesn't allow changing the name of the path or extension to allow any variation in configuration, requiring certs to be in the /etc/nginx/certs folder and be named with .crt extensions. Only the name of the file is allowed to be changed. This requires users of Let's Encrypt to reformat their certificate files, copy them to different locations, and make other administration choices to ensure certificate compliance with the nginx-proxy configuration requirements.

This simple solution offers an alignment with that default configuration, with minimal changes to the nginx.tmpl file, requiring only mounting a volume in the container, and providing the name of the doman/folder. It requires no changes to the Let's Encrypt certbot process, allowing users to externally maintain their certificates without additional sidecars or containers if so desired, using standard tooling for that certificate environment.

I agree there could be additional changes to make these changes more flexible to allow adherence to other certificate management practices, and am open to additional suggestions.

buchdag:
I think my answer lacked precision, the "somewhat specific use case" here is not using Let's Encrypt (I'm aware that they are providing most of today's SSL certificates on the web) or any other ACME-enabled CA, but rather using certbot on your Docker host and injecting the certificate obtained and managed by the host's certbot into your container, as opposed to using acme-companion.

However I do understand that if you're using certbot post renewal hooks, even if acme-companion does support post renewal hooks too it might not be capable enough, or that migrating your existing post renewal tasks to it might not be possible.

I'll give some thought to the implementation I'd prefer for this and get back to you to see if it suits you too.

jekewa:
I understand your thought, but this is not injecting anything anywhere, but simply exposes the already managed certificates to the containers. My simple example used a host-mounted volume for simplicity, but clearly Docker (et al) allows using other shares and mount points as well. It appears as a local folder in the container, but the external management is up to the user. I have one system that does all the certbot activities, and I share that via NFS mounts to the other containers. This way I can control the domains within a certificate, authentication mechanism and details, renewal scheduling and hooks, and all the other things in a single certbot instance. Then the dozens or hundreds of container instances in my infrastructure (or more if. you're bigger) can access the certbot standard configuration and results. Then renewal hooks simply restart containers when a certificate is renewed, meaning no copying or linking is required.

https://letsencrypt.org/docs/integration-guide/#storing-and-reusing-certificates-and-keys

buchdag:
@jekewa what would you think of the following proposal:

- a new env var (lets call it `CERTBOT_MANAGED_CERTIFICATE`) that take a boolean (`false` by default).
- when set to `true` either for a given app container or globally, the template will attempt to find the certbot managed certificate, key and chain inside `/etc/letsencrypt/live/{{VIRTUAL_HOST}}/` and use them if found.
- `CERT_NAME` will still be used to bypass `VIRTUAL_HOST` in the path if necessary.

I'm still not sure about what to do if `CERTBOT_MANAGED_CERTIFICATE` is `true` but expected files are not found. 
- nothing, ie do not enable SSL ?
- attempt to fall back to the matching certificate in `/etc/nginx/certs` then to the default certificate ?

This approach would integrate more seamlessly into the template's existing SSL certificate related logic rather than bypassing it like in the current implementation (which could have some unforeseen effects).

buchdag:
Ping @jekewa 

Would the above implementation proposal suits you ?

jekewa:
Apologies. I’ve got a bit started, but summer distractions.

I’ll get my updated script out next week.

buchdag:
Oh I wasn't suggesting you do it yourself, if what I suggested seems sound to you and would suit your need I can work on the implementation myself 🙂

jekewa:
Heh. Sure.

As long as the file domain name can be specified, because the file might
not be the needed domain in the SAN certs.

PR #2754, opened 2026-06-19
test: mark test_server-down tests as flaky

JamBalaya56562:
## What this does

Marks the three `test/test_server-down` tests with `@pytest.mark.flaky`.

These tests each issue an HTTP request to a proxied virtual host immediately after startup. Under CI load the vhost can occasionally fail to be (re)generated and reloaded before the request, producing a transient `404` — I hit this on the `Unit Tests (debian)` job (`test_load-balancing.py::test_web_has_no_server_down` → `assert 404 == 200`) while the `alpine` job passed on the same code.

`@pytest.mark.flaky` is consumed by `pytest --ignore-flaky`, which the CI workflow (`.github/workflows/test.yml`) already runs, turning a flaky failure into an `xfail` rather than failing the build. This matches the existing convention already used for the readiness-sensitive tests in `test/test_dockergen`.

This was split out of #2753 per review feedback to keep that PR focused on its feature.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/test_server-down/test_load-balancing.py
+++ b/test/test_server-down/test_load-balancing.py
@@ -1,3 +1,10 @@
+import pytest
+
+
+# Marked flaky: this test depends on the proxied vhost being generated and reloaded
+# before the request, which can race with container startup under CI load (observed
+# returning 404 before the upstream was ready). See pytest-ignore-flaky.
+@pytest.mark.flaky
 def test_web_has_no_server_down(docker_compose, nginxproxy):
     conf = nginxproxy.get_conf().decode('ASCII')
     r = nginxproxy.get("http://web.nginx-proxy.tld/port")
--- a/test/test_server-down/test_no-server-down.py
+++ b/test/test_server-down/test_no-server-down.py
@@ -1,3 +1,9 @@
+import pytest
+
+
+# Marked flaky: depends on the proxied vhost being generated and reloaded before the
+# request, which can race with container startup under CI load. See pytest-ignore-flaky.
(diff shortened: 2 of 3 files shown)


buchdag (commented):
I don't think `@pytest.mark.flaky` does anything without [pytest-rerunfailures](https://pytest-rerunfailures.readthedocs.io/latest/). 🤔

JamBalaya56562:
Good thing to double-check! You're right that the bare marker does nothing on its own — but in this repo it's handled by [pytest-ignore-flaky](https://pypi.org/project/pytest-ignore-flaky/) (pinned in `test/requirements/python-requirements.txt`), rather than pytest-rerunfailures.

That plugin turns a failure of a `@pytest.mark.flaky` test into an `xfail`, but only when pytest is run with the `--ignore-flaky` flag — which the CI workflow already does (`.github/workflows/test.yml` runs `pytest --ignore-flaky`). I confirmed locally that with that flag the otherwise-failing `test_web_has_no_server_down` is reported as `xfailed` and the run exits 0, whereas without it the marker is indeed inert. The same bare marker is already used in `test/test_dockergen/`, so this PR just follows that existing convention.

Happy to switch to a retry-based approach (pytest-rerunfailures) or anything else you'd prefer, though.

buchdag:
Right, I forgot we did use [pytest-ignore-flaky](https://pypi.org/project/pytest-ignore-flaky/), my bad 🤦

buchdag (approved):

PR #2764, opened 2026-07-12
feat: optional ocsp stapling (disabled by default)

buchdag:
This PR introduce optional OCSP stapling and disable it by default, rather than enabling it automatically and unconditionaly when the chain file is present like before.

See https://github.com/nginx-proxy/nginx-proxy/issues/2765 and https://letsencrypt.org/2024/12/05/ending-ocsp for the context (and rationale of disabling OCSP stapling by default).


TowyTowy:
Sorry — I completely missed your ping on this back on the 12th, and you ended up merging it without the review you'd asked for.

Read through it properly now. The global env plus per-vhost label split reads right, and gating on the existing `chain.pem` check means anyone who deliberately opts back in keeps the same behaviour as before. It also follows the same `groupByLabel … | keys | first | default …` shape as the http2 and debug-endpoint toggles, so there's nothing surprising in it for anyone reading the template later. Thanks for picking it up.

buchdag:
No worries and thanks for taking the time to review 🙏

PR #2774, opened 2026-08-26
Support containers sharing another container's network namespace

andrioli:
This PR adds support for containers that share the network namespace of another container. This is a common pattern for VPN sidecar containers: an application container is joined to the network namespace of a VPN container (Tailscale, WireGuard, etc.) so its traffic can flow through the VPN.

`docker-gen` already exposes `NetworkMode`, which for these containers holds `container:<id>` of the namespace owner. This change makes `container_ip` resolve the owning container and use its networks instead. When the owning container cannot be found, the behavior is unchanged.

Tests cover the default-port fallback and explicit `VIRTUAL_PORT`. 

Note: The Docker engine rejects create-time port exposure (`--expose` /  `expose:`) combined with a container type network mode (`conflicting options: port exposing and the container type network mode`), so a joining container's exposed ports can only come from its image. I manually tested a custom image with an `EXPOSE` port (other than 80) without the `VIRTUAL_PORT` env var.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -331,6 +331,12 @@ In this example, the `my-nginx-proxy` container will be connected to `my-network
 
 Proxyed containers running in host network mode **must** use the [`VIRTUAL_PORT`](#virtual-ports) environment variable, as this is the only way for `nginx-proxy` to get the correct port (or a port at all) for those containers.
 
+### Shared network namespaces
+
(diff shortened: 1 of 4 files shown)


andrioli:
Just wanted to leave some context on the motivation while it's fresh, in case it helps whenever someone gets here.

As I said before, the use case is the standard VPN-sidecar pattern. Tailscale's own Docker [guide](https://tailscale.com/blog/docker-tailscale-guide) recommends it  and Gluetun [wiki](https://github.com/qdm12/gluetun-wiki/blob/main/setup/connect-a-container-to-gluetun.md) for WireGuard/OpenVPN also do it.

Today that combination fails silently with nginx-proxy: the app container declares `VIRTUAL_HOST` but has no networks of its own, so the upstream gets no server and nginx answers 503.

The workaround we're running meanwhile is moving `VIRTUAL_HOST` (and `LETSENCRYPT_HOST`, and the `expose:`) onto the sidecar. It works, but the routing announcement ends up on a container that isn't the one serving the traffic.

Happy to do whatever helps this land, as this will also enhance our current deployment configuration!

PR #2775, opened 2026-08-28
fix: make entrypoint independent of WORKDIR

### LINKED ISSUE #1238: Setting WORKDIR in Dockerfile breaks ENTRYPOINT script.
Using this Dockerfile:

```Dockerfile
FROM jwilder/nginx-proxy

WORKDIR /etc/nginx
```

Will cause this error when starting a container based on it:

> ERROR: open Procfile: no such file or directory

Removing the WORKDIR statement resolves the issue. I'm guessing the ENTRYPOINT shouldn't rely on the WORKDIR being set to `/app`.

ump45nose:
## Summary

`forego start` looks for its `Procfile` in the current working directory (it defaults to `./Procfile`), so the image's `CMD ["forego", "start", "-r"]` breaks with `ERROR: open Procfile: no such file or directory` whenever a downstream image overrides `WORKDIR`, e.g.:

```Dockerfile
FROM nginxproxy/nginx-proxy
WORKDIR /etc/nginx
```

This makes the entrypoint/process independent of the current `WORKDIR` by passing the Procfile path explicitly to Forego with the `-f` flag:

```Dockerfile
CMD ["forego", "start", "-r", "-f", "/app/Procfile"]
```

`-f /app/Procfile` both loads the Procfile from a fixed location and (per Forego's `start.go`, which uses `filepath.Dir(flagProcfile)` as the child working directory) pins the child processes' working directory to `/app`, matching the existing `WORKDIR /app/` behavior.

The `docker-entrypoint.sh` default-CMD detection is updated to match the new arguments so the init logic (docker socket check, DNS resolver detection, dhparam setup) keeps running unchanged.

## Regression test

Added `test/test_workdir`, modeled on the existing `test_resolvers` module: a `.base.yml` runs the proxy with `working_dir: /etc/nginx` and asserts the container stays running and its logs do not contain `open Procfile: no such file or directory`.

## Testing

Docker is not available in the local environment, so the Docker-based pytest suite could not be run locally. Verified:
- `bash -n app/docker-entrypoint.sh` (syntax)
- the entrypoint's default-CMD detection logic matches the new argument string
- `python3 -m py_compile test/test_workdir/test_workdir.py`
- the new test YAML parses correctly

The CI matrix (`test.yml`) builds both `Dockerfile.alpine` and `Dockerfile.debian` and runs the full suite, which will exercise this fix.

Fixes #1238

## AI disclosure

This PR was prepared with AI assistance (Claude Opus 5, via the DeepSeek harness coding agent). The implementing commit carries the AI agent as co-author, and the changes were authored and verified by the account owner before submission.


ump45nose:
CI note: the alpine job failed only on 4 SSL tests in `test_ssl/test_virtual-path.py` (`assert 404 == 301` / HTTPS connection refused), while:

- **debian**: all green
- **alpine**: 470 passed, only those 4 SSL cases failed
- the new regression test `test_workdir/test_workdir.py::test_starts_with_overridden_workdir` **passed**

This change only adds `-f /app/Procfile` to the forego CMD (and the matching entrypoint detection string); it does not touch TLS/SSL/certs. The failures look like the known-flaky SSL suite. Could you rerun the failed jobs when convenient?

Fixes #1238

buchdag (approved):

ump45nose:
The only failed check is `Unit Tests (alpine)`: three existing HTTP/3 vhost `/headers` tests return 404 (`test_http3-vhost.py` enabled/disabled/default-disabled), while the new `test_workdir` coverage is not among the failures. The PR is otherwise approved. I cannot rerun an upstream workflow from this fork (GitHub requires repository admin rights); could a maintainer rerun this job or confirm the Alpine HTTP/3 environment issue?

buchdag:
@ump45nose is this PR AI-generated ? If yes, I'd prefer it to be disclosed both in the PR opening message and by adding the relevant AI agent as co-author of the commit.

ump45nose:
Hi @buchdag — thanks for the review and approval.

To answer your question: yes, this PR was prepared with AI assistance (Claude Opus 5, running in a coding-agent harness). In line with your request:

- The PR description now contains an explicit AI disclosure section.
- The commit has been amended to add the AI agent as co-author (`Co-authored-by: Claude Opus 5 <noreply@anthropic.com>`) and force-pushed to the PR branch.

Regarding the remaining Unit Tests (alpine) failures: they are the pre-existing HTTP/3 /headers and SSL vhost tests failing in the Alpine environment, unrelated to this change (the new test_workdir coverage passes). I can't rerun upstream workflows from the fork — could a maintainer rerun that job when convenient?

Thanks!

ump45nose:
Current CI refresh: the failed job has moved to `Unit Tests (debian)`, with only two existing ACME challenge tests failing (`expected 301/404, got 200`); 472 tests pass, and the PR-specific `test_workdir` regression remains outside the failures. Since the approved change only pins Forego\047s Procfile path and does not touch ACME routing, this again appears environment/flakiness-related. A maintainer rerun is still needed because fork authors cannot rerun the upstream job.

